Issue #17, opened 2017-05-03
getCurrentTrack returns null


jasongrishkoff:
If I fire this while a song is playing:
```
    TrackPlayer.getCurrentTrack(id,function(string) {
        console.log(string)
    })
```
I get string = null.  I'm hoping to use this feature so that if the app does close in the background while the player keeps going, when resuming the app I'm able to restore the state and show the currently playing song in the UI. From the docs, it seems like it's expected to return the track ID?  Would be cool if it returned the whole object, including artist name, song title, artwork, etc.

Guichaguri:
Are you using `load` or `add` to load the tracks? Which player are you using (ExoPlayer, native Android player, etc)?

For now I get rid of the whole object, keeping only the important information. I don't know if I should keep it, it can definitely be helpful for changing the player (from a local one to a Chromecast device). What do you think? The only reason to dispose the object is to reduce a bit of memory usage.

jasongrishkoff:
I'm using `load` to play each track; I'm using `add` to line up the future tracks.  This is with ExoPlayer.

1. Load track
2. Play track
3. Cycle through and add 50 other tracks

When you say remove the whole object, which are you referring to?

jasongrishkoff:
Was messing around with this today because occasionally my React instance gets misaligned with the player and shows the wrong currently playing song.  Would be rad if I could somehow sync them, but no matter what I do this keeps returning "null." I was able to add a bit of a console log and convert currentTrack to a string so that I could output its value, and I do indeed see it increment from 0 to 1 to 2, etc. Dunno if something's maybe wrong with "queue.get"?

jasongrishkoff:
Any tips on how I can get this to return the currently-playing song object?

dcvz:
At the moment `getCurrentTrack` returns the `id` of the track (which you're supposed to provide; mandatory according to the docs). It could be changed to provide the whole object but at the moment there's no way of doing this.

jasongrishkoff:
This is returning null on Android for me:
```
TrackPlayer.getCurrentTrack((track) => {
    console.log(track)
})
```

jasongrishkoff:
Okay, resolved this issue. It's because I was passing an integer instead of a string as the track id.

PR #24, opened 2017-07-01
Destroy players onDestroy

dcvz:
There was an issue that even when the application was closed by the user; the service would remain and continue playing.. This will fix that issue.

### DIFF
--- a/android/src/main/java/guichaguri/trackplayer/TrackModule.java
+++ b/android/src/main/java/guichaguri/trackplayer/TrackModule.java
@@ -56,6 +56,7 @@ public class TrackModule extends ReactContextBaseJavaModule implements ServiceCo
         super.onCatalystInstanceDestroy();
 
         // Unbinds the service
+        manager.destroy(-1);
         getReactApplicationContext().unbindService(this);
     }
 


jasongrishkoff:
Hey @dcvz, one of the important things to consider is that sometimes Android does memory sweeps and closes apps in the background. This is problematic if you're listening to a music app that has a multi-song playlist (such as the one I'm building), and the music randomly stops because Android sweeps up the app.  So, Gui separated the Notification Component from the React Native app.

Does this change differentiate between whether the app was manually closed or whether it was memory cleaned?

Guichaguri:
I've been considering this as a feature, not a bug. Apps like SoundCloud or Spotify keep playing even after you close them.

The main problem is that the music is being fed by the service itself on those apps, and in React Native, there's not really a difference between "background code" and "UI code", this is the problem I'm struggling with: running JS code from a service without messing stuff up.
I think I have a solution in mind, but it will take time to implement.

Also, I'm pretty sure @jasongrishkoff is right and the music would stop in memory sweeps

jasongrishkoff:
Yeah, right now what I do is I use the "queue" method of this library to line up 100 tracks -- and use JS code in my app to have those same 100 tracks in a playlist. So it's like I have two concurrent playlists running -- one for the UI/JS in the Native app, and the other which feeds this library and ensures that the songs keep playing even if the app gets memory swept.  Very occasionally my JS code gets unsynced from the library -- see issue #17

dcvz:
I understand. In that case it might be good to provide a setting for this? In my case playback is tied to the current state of the app; so playing music without the app being activated isn't desired.

Guichaguri:
@dcvz We can provide a `background` boolean parameter in `setupPlayer`. It should be enabled by default, but when disabled, the playback should stop and destroy itself as soon as the app closes.

Guichaguri:
Added in 8698097aff55d86b5d213813effbef0df639246a

PR #25, opened 2017-07-02
iOS Module

dcvz:
**!! NOT READY FOR MERGING !!**

This is the very first integration with iOS. It's currently operational via the `load/play` methods 🌻

**Missing Functionality:**
- [x] queueing
- [ ] integration with ControlCenter
- [x] support for local files

**Missing JS Methods:**
- [ ] setOptions
- [x] add
- [x] remove
- [x] skip
- [x] skipToNext
- [x] skipToPrevious
- [ ] getBufferedPosition


ios/RNTrackPlayer/Models/Player.m
```diff
@@ -0,0 +1,127 @@
...
+}
+
+- (NSInteger)position {
+    return [_player progress];
+}
+
+- (NSInteger)state {
+    // Mapping to values used by Android project
+    // TODO: - Unify into a shared enum
+    switch ([_player state]) {
```
dcvz: At the moment I'm mapping the native player state to the values used by the Android project. @Guichaguri it would make sense for us to create a shared enum.

Guichaguri: We don't need a shared enum, React Native (at least for Android) has a `getConstants` method, which you can provide your own values. The only thing you need is to have the same variable names that Android has.

Guichaguri:
Let me know if anything in iOS works differently from Android, so we can make them work together the same way.

For instance, I know headless tasks are not a thing in iOS, so we'll have to use a different method for events. Maybe just a few simple functions in the module itself? (on/off/once)

Also, I've been talking to Jason about the support for multiple players:
* Remove the support, leaving only the main player (and a function to turn the main player into a Chromecast player)
* Keep the support, but make other players (except the main one) pause when the app is not in foreground
* Keep the support the way it is, but have a small performance slowdown.

I don't see any reason to keep support for multiple players other than a mixing app or for music transitions (such as cross fading), so I'd like to hear your opinion.

ios/RNTrackPlayer/Models/Player.m
dcvz: That's what I meant by shared enum :) In theory we'd have the same values for both platforms.. which is probably best to be exposed as an enum instead of constants.. but either way yes we'll need the same values for both.

dcvz:
In iOS we don't need to run a background service for playing music so we could easily detect the controls a user sends through the control center (equivalent as from notification center in Android) and post events that our users can subscribe to.

re: multiple players. I also wondered the same thing. I don't think there's a need for multiple players; it should be enough to have one (with option to turn into chromecast). Even crossfading can be achieved without multiple players. Having 1 player will make it easier to do other cool things in the future (like preloading)

Guichaguri:
Is there a reason to keep this PR open, as all the progress on the iOS side is being done in the dev branch?

chrismang:
> In iOS we don't need to run a background service for playing music so we could easily detect the controls a user sends through the control center (equivalent as from notification center in Android) and post events that our users can subscribe to.
> 
> re: multiple players. I also wondered the same thing. I don't think there's a need for multiple players; it should be enough to have one (with option to turn into chromecast). Even crossfading can be achieved without multiple players. Having 1 player will make it easier to do other cool things in the future (like preloading)

Hi! I'm looking for support to implement in my app a crossfading feature in a queue of songs. There is a guideline or some tips to do it with this library? What should I know to do this myself if this is not done yet?
Thanks!

Guichaguri:
@chrismang  See #670

PR #50, opened 2017-09-20
[ios] Fix playing local files

anderslemke:


dcvz:
Good fix @anderslemke :D

PR #53, opened 2017-10-04
Unify events into one handler

dcvz:
you will now register one handler (like Android)

`TrackPlayer.registerEventHandler((event) => {})`

which will return all the different events defined in the wiki :)

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -71,10 +71,10 @@ function registerEventHandler(handler) {
         ]
 
         for (let i = 0; i < events.length; i++) {
-            eventEmitter.addListener(event, (payload) => { warpResponse(event, payload) });
+            eventEmitter.addListener(warpResponse.bind(null, events[i]));
         }
     } else {
-        AppRegistry.registerHeadlessTask('TrackPlayer', handler);
+        AppRegistry.registerHeadlessTask('TrackPlayer', () => handler);
     }
 }
 
--- a/lib/index.js
+++ b/lib/index.js
@@ -71,7 +71,7 @@ function registerEventHandler(handler) {
         ]
 
         for (let i = 0; i < events.length; i++) {
-            eventEmitter.addListener(warpResponse.bind(null, events[i]));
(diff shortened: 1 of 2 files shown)


lib/index.js
```diff
@@ -48,6 +48,36 @@ function remove(tracks) {
     return TrackPlayer.remove(tracks);
 }
 
+function registerEventHandler(handler) {
+    if (Platform.OS !== 'android') {
+        function warpResponse(event, payload) {
+            // transform into headlessTask format and return to handler
+            const additionalKeys = payload || {};
+            handler({ type: event, ...additionalKeys });
```
Guichaguri: "handler" is a function that returns the event handler function. It's looks unnecessary, but that's how react native works

```js
AppRegistry.registerHeadlessTask('TrackPlayer', () => {
    return async (data) {
        // Handle event
    }
});
```

I guess it's to only initialize the event handler code when it needs to.
https://facebook.github.io/react-native/docs/headless-js-android.html#the-js-api

lib/index.js
```diff
@@ -48,6 +48,36 @@ function remove(tracks) {
...
+
+            'remote-play',
+            'remote-next',
+            'remote-stop',
+            'remote-pause',
+            'remote-previous',
+        ]
+
+        for (let i = 0; i < events.length; i++) {
+            eventEmitter.addListener(event, (payload) => { warpResponse(event, payload) });
```
Guichaguri: `event` does not exist?

I think you can replace `(payload) => { warpResponse(event, payload) }` with `wrapResponse.bind(null, events[i])`

dcvz:
want to take up another look @Guichaguri

lib/index.js
```diff
@@ -71,10 +71,10 @@ function registerEventHandler(handler) {
         ]
 
         for (let i = 0; i < events.length; i++) {
-            eventEmitter.addListener(event, (payload) => { warpResponse(event, payload) });
+            eventEmitter.addListener(warpResponse.bind(null, events[i]));
```
Guichaguri: You still need the event name as the first parameter in `addListener`

PR #66, opened 2017-10-26
Fix/number exception

moduval:
Fix a bug with maxArtworkSize and ratingType throwing a double cast exception when setting them


Guichaguri:
Can you send the crash report?

Temp was not meant to be edited at all, it was created as a temporary placeholder that is going to be removed sometime soon, as newer versions of react-native now have those same methods.

Perhaps this can be fixed the part I retrieve the `maxArtworkSize`, `ratingType`. Or this PR should go into RN's code

moduval:
Here is the error i got before fixing the bug

```
10-27 23:02:33.679 3977-3977/ca.nexapp.radioplace W/Bundle: Key ratingType expected Integer but value was a java.lang.Double.  The default value 1 was returned.
10-27 23:02:33.679 3977-3977/ca.nexapp.radioplace W/Bundle: Attempt to cast generated internal exception:
                                                            java.lang.ClassCastException: java.lang.Double cannot be cast to java.lang.Integer
                                                                at android.os.BaseBundle.getInt(BaseBundle.java:873)
                                                                at guichaguri.trackplayer.metadata.Metadata.updateOptions(Metadata.java:78)
                                                                at guichaguri.trackplayer.logic.MediaManager.updateOptions(MediaManager.java:67)
                                                                at guichaguri.trackplayer.logic.components.MediaWrapper$2.run(MediaWrapper.java:45)
                                                                at android.os.Handler.handleCallback(Handler.java:751)
                                                                at android.os.Handler.dispatchMessage(Handler.java:95)
                                                                at android.os.Looper.loop(Looper.java:154)
                                                                at android.app.ActivityThread.main(ActivityThread.java:6119)
                                                                at java.lang.reflect.Method.invoke(Native Method)
                                                                at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:886)
                                                                at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:776)
```

moduval:
I updated my pull request using react-native method instead and fixing the cast exception bug in the Metadata file (react-native method has the same problem than your Temp file)

Guichaguri:
Can you keep the Temp class for now? The plan is to keep this version working with older react-native versions, and only update it in the next one.
Thanks

moduval:
I bring back the old Temp class. This pull request should be ready to merge now

Guichaguri:
Thank you :)

PR #69, opened 2017-10-29
Fix/compact capabilities

moduval:
When setting compact capabilities like this in code : 

```javascript
    TrackPlayer.updateOptions({
      ratingType: RATING_HEART,
      capabilities: [CAPABILITY_PLAY, CAPABILITY_PAUSE, CAPABILITY_STOP],
      compactCapabilities: [CAPABILITY_PLAY, CAPABILITY_PAUSE, CAPABILITY_STOP],
      stopWithApp: true
    });
```

They doesn't appear in player.

Also, when an action is not enabled in capabilities, they were not currently added to the compact capabilities either even if they are set in the compactCapabilities array.

This pull request fix these problems.

### DIFF
--- a/android/src/main/java/guichaguri/trackplayer/metadata/components/MediaNotification.java
+++ b/android/src/main/java/guichaguri/trackplayer/metadata/components/MediaNotification.java
@@ -181,7 +181,7 @@ public class MediaNotification {
         if(action == PlaybackStateCompat.ACTION_PAUSE && !playing) return instance;
 
         // Add it to the compact view if it's allowed to
-        if((compactCapabilities & action) == 0) {
+        if((compactCapabilities & action) == action) {
             compactView.add(instance);
         }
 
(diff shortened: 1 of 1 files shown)


android/src/main/java/guichaguri/trackplayer/metadata/components/MediaNotification.java
```diff
@@ -171,6 +171,11 @@ private int loadIcon(String iconName) {
 
     private Action addAction(Action instance, long mask, long action, String title, int icon,
                              List<Action> list, List<Action> compactView, boolean playing) {
+        // Add it to the compact view if it's allowed to
+        if((compactCapabilities & action) == action) {
+            compactView.add(instance);
+        }
+
```
Guichaguri: Why did you move it up here?

moduval: When an action is not enabled in capabilities, they were not currently added to the compact capabilities either even if they are set in the compactCapabilities array. This is because the updateActions method return null if the capabilities is not set.

moduval: Hum after retesting my code it seem to throw an exception when moving it here...

moduval: ha I think that the actions in compact view should be at least the same than them in capabilities array. Sorry for that i'm gonna put the code down where it was

Guichaguri:
Thank you :)

PR #70, opened 2017-10-30
Fix ios build related to constantsToExport method

### LINKED ISSUE #61: iOS: Build failed
I tried to us this but i cant build a app cause of this:

  override func constantsToExport() -> [String: Any] { !! Overriding method with selector 'constantsToExport' has incompatible type '() -> [String : Any]'

in

https://github.com/react-native-kit/react-native-track-player/blob/dev/ios/RNTrackPlayer/RNTrackPlayer.swift

using: xcode Version 9.0 (9A235)

moduval:
fix #61


olivierlesnicki:
it should be 
`override func constantsToExport() -> [AnyHashable: Any]! {`
instead of 
`override func constantsToExport() -> [AnyHashable: Any] {`

olivierlesnicki:
It should be `[AnyHashable: Any]!` instead of `[AnyHashable: Any]`

Issue #44, opened 2017-09-14
URL-encoding inconsistencies


anderslemke:
On iOS, I have to decode the URL for the player:

    url: "https://example.com/æ.mp3"

On Android, I have to encode it:

    url: "https://example.com/%C3%A6.mp3"

Guichaguri:
The right solution would be to support both of them on both platforms, right?

dcvz:
I think we should expect the user to use non-encode URL's. It should not be the client's job to encode.

Guichaguri:
We also should expect encoded URLs. The URL can be received encoded from a server or even hardcoded.

Guichaguri:
For now on, you'll have to encode the URL to use on both platforms (See #71)

Issue #40, opened 2017-09-11
[iOS] Having `%` in URLs break the player


anderslemke:
Having an audio URL like this:

```
https://d23q6m8guwx14t.cloudfront.net/e1173810-9496-11e7-afc9-25da05a60416/mindstel%C3%B8n_mixdown.mp3
```

will make the `AudioPlayer` fail.

anderslemke:
A solution for know, would be to `decodeURI("https://d23q6m8guwx14t.cloudfront.net/e1173810-9496-11e7-afc9-25da05a60416/mindstel%C3%B8n_mixdown.mp3
")` before adding the track to player...

PR #71, opened 2017-10-31
Fix URL encoding problems

olivierlesnicki:
Removed encoding of the URL for iOS.

### DIFF
--- a/ios/RNTrackPlayer/RNTrackPlayer.swift
+++ b/ios/RNTrackPlayer/RNTrackPlayer.swift
@@ -45,7 +45,7 @@ class RNTrackPlayer: RCTEventEmitter, MediaWrapperDelegate {
     // MARK: - Required Methods
     
     @objc(constantsToExport)
-    override func constantsToExport() -> [AnyHashable: Any] {
+    override func constantsToExport() -> [String: Any] {
         return [
             "STATE_NONE": AudioPlayerState.stopped,
             "STATE_PLAYING": AudioPlayerState.playing,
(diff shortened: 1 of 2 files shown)


Guichaguri:
Why is this needed?

olivierlesnicki:
I'm using Firebase storage, which produces audio URL that looks a bit like this:
`https://firebasestorage.googleapis.com/v0/b/theviemusic-c2f2f.appspot.com/o/27723940%2Faudio%2FB1YqFDL0--CHILL.mp3?alt=media&token=fd8026a3-18ba-4c95-9a43-b032cb0f71bd`
(note the presence of %2 in the URN)

The problem is when added to Track Player, these URL are encoded to:
`https://firebasestorage.googleapis.com/v0/b/theviemusic-c2f2f.appspot.com/o/27723940%252Faudio%252FB1YqFDL0--CHILL.mp3?alt=media&token=fd8026a3-18ba-4c95-9a43-b032cb0f71bd`

I think it's a much better rule not to encode URL and leave this decision to the client.

ios/RNTrackPlayer/RNTrackPlayer.swift
```diff
@@ -45,7 +45,7 @@ class RNTrackPlayer: RCTEventEmitter, MediaWrapperDelegate {
     // MARK: - Required Methods
     
     @objc(constantsToExport)
-    override func constantsToExport() -> [String: Any] {
+    override func constantsToExport() -> [AnyHashable: Any] {
```
Guichaguri: Please, make one modification per commit. We already have #70 for this one.

Guichaguri:
So this is a possible fix for #40 and #44, right?

olivierlesnicki:
It should fix both #40 and #44

dcvz:
I agree with statement that we should leave encoding to the client. Let's merge this 👍

AylanBoscarino:
The problem persists tho. 
I can't play a track on iOS with Firebase Storage's URL.

PR #80, opened 2017-11-17
Add track and position to playback-queue-ended

anderslemke:
As my use case requires a double check of the actual position when a track ends, I need a way to get the position for the last track playing on a queue. This is currently not possible (at least on iOS).

When the last track ends, only `playback-queue-ended` is fired, and not `playback-track-changed` which otherwise contains the needed information.

I would love to work on a solution for this and submit a PR, but would like to get confirmation on API before I implement it.

The easy fix, would be to add `track` and `position` to `playback-queue-ended`. I've started this work here (and it seems to work just fine).

Let me know if you want me to continue down this road.

- [x] iOS
- [x] Android
- [x] Windows
- [ ] Update wiki

### DIFF
--- a/windows/RNTrackPlayer/Players/Playback.cs
+++ b/windows/RNTrackPlayer/Players/Playback.cs
@@ -61,6 +61,13 @@ namespace TrackPlayer.Players {
             manager.OnTrackUpdate(previous, position, track, true);
         }
 
+        protected void HandleOnEnd() {
+            Track previous = GetCurrentTrack();
+            double position = GetPosition();
+
+            manager.OnEnd(previous, position);
+        }
+
         public Track GetCurrentTrack() {
             return currentTrack >= 0 && currentTrack < queue.Count ? queue[currentTrack] : null;
         }
--- a/android/src/main/java/guichaguri/trackplayer/logic/MediaManager.java
+++ b/android/src/main/java/guichaguri/trackplayer/logic/MediaManager.java
@@ -212,10 +212,13 @@ public class MediaManager {
         }
     }
 
-    public void onEnd() {
+    public void onEnd(Track previous, long prevPos) {
         Log.d(Utils.TAG, "onEnd");
 
-        Events.dispatchEvent(service, Events.PLAYBACK_QUEUE_ENDED, null);
+        Bundle bundle = new Bundle();
+        bundle.putString("track", previous != null ? previous.id : null);
+        bundle.putDouble("position", Utils.toSeconds(prevPos));
+        Events.dispatchEvent(service, Events.PLAYBACK_QUEUE_ENDED, bundle);
     }
 
     public void onStateChange(int state) {
--- a/android/src/main/java/guichaguri/trackplayer/player/Playback.java
+++ b/android/src/main/java/guichaguri/trackplayer/player/Playback.java
@@ -231,4 +231,11 @@ public abstract class Playback {
 
         manager.onTrackUpdate(previous, position, next, true);
     }
+
+    protected void onEnd() {
+      Track previous = getCurrentTrack();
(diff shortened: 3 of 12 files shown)


dcvz:
if the last track ends then you know it played the whole duration of the track. Do you explicitly need it to be returned?

anderslemke:
Hi @dcvz . Yes. I would prefer to get the actual position.
Our use case, makes it is crucial that we know the position, and I want to be able to verify that the position of the player is actually the end of the file. So I compare it to a previously computed value of the file length and determine if the file is _actually_ played in whole.

My experience tells me that I can't trust iOS to download and play the _whole_ file, which would lead to a false "completed" for a given track, which we can't tolerate.

I hope this makes sense, and I'm happy to elaborate.

anderslemke:
Another solution might be to not `stop` the player on iOS when queue ends.

Then I would be able to poll the player for current position and use that. This is actually possible on Android.

But on iOS, the position resets to 0 when the queue ends, so I'm not able to get the position there. I think it's because you call `stop()` and not just `pause()` on iOS.

In any case, I would prefer to _not_ call `stop` on iOS as it also removes the track from the lock screen and control center. But that's a different issue and just my opinion.

This is the line I'm talking about:
https://github.com/react-native-kit/react-native-track-player/blob/dev/ios/RNTrackPlayer/Logic/MediaWrapper.swift#L115

dcvz:
I've changed the functionality to only `pause()` to be more on track with Android. The other requested API changes I will talk about with @Guichaguri. In the meantime this should allow you to work around your problem right?

dcvz:
If this works for you let me know if we can close the PR :)

anderslemke:
@dcvz Where did you change the `stop` to a `pause`?
I can't find it...

dcvz:
Here: https://github.com/react-native-kit/react-native-track-player/commit/dec1bb6c3f7cd75d7e71dc3086d163290c81d643

sorry was hidden by another change :(

anderslemke:
OK. But what about this line: https://github.com/react-native-kit/react-native-track-player/blob/dev/ios/RNTrackPlayer/Logic/MediaWrapper.swift#L138

Won't that run on queue exhaustion, and then make the player stop?

anderslemke:
I've tested https://github.com/react-native-kit/react-native-track-player/commit/dec1bb6c3f7cd75d7e71dc3086d163290c81d643 and it doesn't solve my problem.

When I poll the player after receiving the ´playback-queue-ended` it returns position 0.

I'm quite confident it's because of this line: https://github.com/react-native-kit/react-native-track-player/blob/dev/ios/RNTrackPlayer/Logic/MediaWrapper.swift#L138

I've tried changing it to read `pause()` instead of `stop()`, and that makes it work.

I've created #83 to show you what I mean.

BUT;
I would actually rather have the event include the information explicitly, and not rely on some behavior that might change.

anderslemke:
I've added Android and Windows support.
Android is tested, but I haven't been able to test Windows support.

windows/RNTrackPlayer/Players/Playback.cs
```diff
@@ -61,6 +61,18 @@ public abstract class Playback {
...
+                UpdateCurrentTrack(currentTrack + 1, null);
+                Play();
+            } else {
+                Track previous = GetCurrentTrack();
+                double position = GetPosition();
+
+                manager.OnEnd(previous, position);
+            }
+        }
+
```
Guichaguri: Why did you moved this function here? It accepts a MediaPlayer as an argument because it is an event listener for the local playback.

anderslemke: OK. I'll give it another go.

PR #84, opened 2017-11-24
Add ability to set speed and pitch

anderslemke:


Guichaguri:
I've just pushed Android and UWP support. Unfortunately, the UWP API doesn't support changing the speed and the pitch individually, so it will change the playback rate instead.

egorkhmelev:
Hey guys, I have found several issues here.

1. Calling `getPlaybackParams` when AndroidPlayer has not been initialized leads to an IllegalStateException (which actually happens within `TrackPlayer.updateOptions` method call during app init). `getPlaybackParams` should only be called on initialized state which occurs only after `setDatasource` call has been made.

2. Rate doesn't really work. The right way to set playback rate is:
```player.setPlaybackParams(player.getPlaybackParams().setSpeed(rate));```

Hope this helps.

**Comment by egorkhmelev** [comment]
Also, one tricky moment of using setSpeed method:

> After the object is prepared, calling it with zero speed is equivalent to calling pause(). After the object is prepared, calling it with non-zero speed is equivalent to calling start().

That is not obvious for developers and leads to an issues when music suddenly starts playing without user press any button.

https://developer.android.com/reference/android/media/MediaPlayer.html#setPlaybackParams(android.media.PlaybackParams)

**Comment by Guichaguri** [comment]
@egorkhmelev Thanks, I'll fix those when I get back home :)

**Comment by Guichaguri** [comment]
@egorkhmelev Fixed both of them. Even though I'm keeping it working now, it's very likely we'll get rid of `AndroidPlayer` and use ExoPlayer as our main one in the future. It is way faster, less buggy and predictable, and also supports adaptive streaming.

**Comment by egorkhmelev** [comment]
@Guichaguri thank you! Your fix is exactly how I done it myself :)

Never heard of `ExoPlayer` before, I'll definitely give it a try!

**Comment by gregoiredumas** [comment]
Any update on the merge?

**Comment by Guichaguri** [comment]
@gregoiredumas I'm waiting for the iOS side to be done and then accepted by @dcvz

**Comment by anderslemke** [comment]
Oh. That might be my fault here. The iOS side is done. So this is ready for review.

File: ios/RNTrackPlayer/RNTrackPlayer.swift [hunk 1]
```diff
@@ -218,6 +218,18 @@ class RNTrackPlayer: RCTEventEmitter, MediaWrapperDelegate {
...
+    @objc(setRate:)
+    func setRate(rate: Float) {
+        /* TODO */
+        mediaWrapper.rate = rate
+    }
+    
+    @objc(getRate)
+    func getRate() {
+        print("Getting current rate")
+        /* TODO */
```
dcvz: It looks like there's still implementation missing here?

ios/RNTrackPlayer/RNTrackPlayer.swift
```diff
@@ -218,6 +218,18 @@ class RNTrackPlayer: RCTEventEmitter, MediaWrapperDelegate {
...
     }
     
+    @objc(setRate:)
+    func setRate(rate: Float) {
+        /* TODO */
+        mediaWrapper.rate = rate
+    }
+    
+    @objc(getRate)
+    func getRate() {
```
dcvz: This method is not being bridged in the `.m` file.

ios/RNTrackPlayer/Models/Track.swift
```diff
@@ -42,6 +43,7 @@ class Track: NSObject {
         self.album = dictionary["album"] as? String
         self.genre = dictionary["genre"] as? String
         self.desc = dictionary["description"] as? String
+        self.pitchAlgorithm = dictionary["pitchAlgorithm"] as? String
```
dcvz: For this you can probably export a constant that the JS side can use. Or you can add a helper on the JS side to make sure that only the allowed values are passed in

alangumer:
Hey guys, any update on this? Thanks!

ios/RNTrackPlayer/Logic/MediaWrapper.swift
```diff
@@ -19,6 +19,7 @@ protocol MediaWrapperDelegate: class {
 class MediaWrapper: AudioPlayerDelegate {
     private var queue: [Track]
     private var currentIndex: Int
+    private var _rate: Float
```
dcvz: Is there a need for this? can't we just `set` or `get` `player.rate`?

ios/RNTrackPlayer/RNTrackPlayer.swift
```diff
@@ -60,6 +60,10 @@ class RNTrackPlayer: RCTEventEmitter, MediaWrapperDelegate {
             "STATE_STOPPED": "STATE_STOPPED",
             "STATE_BUFFERING": "STATE_BUFFERING",
             
+            "PITCH_ALGORITHM_LINEAR": PitchAlgorithm.linear.rawValue,
```
dcvz: Amazing!

mahmoudfelfel:
Thanks guys for adding this 👍, I've been searching for a library that supports both streaming and speed control, any plans of merging this PR soon?

ios/RNTrackPlayer/Logic/MediaWrapper.swift
anderslemke: It does seem a bit silly. I'll remove 👍

ios/RNTrackPlayer/Models/Track.swift
anderslemke: I've added constants that are exported. 

(Just for the record)

dcvz:
Is this good on your side @Guichaguri? If so, I'll merge this now.

dcvz:
Thanks for this contribution @anderslemke sorry that it took so long! 🚀

PR #88, opened 2017-12-03
Update README.md: make function async

puckey:
A function in the example code in the README needed to be async.. (I also ran into the problem in this issue: https://github.com/react-native-kit/react-native-track-player/issues/55 – wondering if the solution in there should not be added to the example code?)


Guichaguri:
Somehow I missed that :P
Thanks

PR #98, opened 2017-12-19
fix empty constants for remote media player

chrissbendel:
This fixes the empty constants issue for the media controls in iOS. Tested and working on physical device


ios/RNTrackPlayer/RNTrackPlayer.swift
```diff
@@ -56,11 +56,11 @@ class RNTrackPlayer: RCTEventEmitter, MediaWrapperDelegate {
             "STATE_STOPPED": AudioPlayerState.stopped,
             "STATE_BUFFERING": AudioPlayerState.buffering,
```
Guichaguri: Shouldn't the same be applied to the states? AFAIK, they are also not working properly

dcvz:
Thank you for contribution! This was already added in: ade467305003890273a271881051ca2535bd39ec and 5e2ce7db043a16fe26af6f3237b7488c57541024

PR #103, opened 2017-12-28
[Android] Keep js context after unbind

egorkhmelev:
Give an ability for js context to restart and keep working (as headless task) after unbind. This is useful when project needs some logic like sleep timer or analytics. Also its useful to keep track on current progress to restore later. And it will also helps to keep notification responsive when use controls if app is closed.

Its worth noting that timers is needed (setTimeout/setInterval) in Headless task its possible to return promise on `playback-unbind` event from js and keep it (do not resolve) until app become active again.

Not really sure about constant naming as its not really related to playback.
Would love to hear some feedback.

### DIFF
--- a/android/src/main/java/guichaguri/trackplayer/logic/services/PlayerService.java
+++ b/android/src/main/java/guichaguri/trackplayer/logic/services/PlayerService.java
@@ -36,6 +36,14 @@ public class PlayerService extends MediaBrowserServiceCompat {
         }
     }
 
+    @Override
+    public boolean onUnbind(Intent intent) {
+        Log.d(Utils.TAG, "onUnbind");
+        super.onUnbind(intent);
+        manager.onServiceUnbounded();
+        return true;
+    }
+
     @Nullable
     @Override
     public BrowserRoot onGetRoot(@NonNull String clientPackageName, int clientUid, @Nullable Bundle rootHints) {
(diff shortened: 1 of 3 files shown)


lib/index.js
```diff
@@ -72,6 +72,7 @@ function registerEventHandler(handler) {
             'playback-error',
             'playback-queue-ended',
             'playback-track-changed',
+            'playback-unbind',
```
Guichaguri: This array is for iOS and Windows events, not Android ones.

android/src/main/java/guichaguri/trackplayer/logic/services/PlayerService.java
```diff
@@ -36,6 +36,14 @@ public IBinder onBind(Intent intent) {
         }
     }
 
+    @Override
+    public boolean onUnbind(Intent intent) {
+        Log.d(Utils.TAG, "onUnbind");
+        super.onUnbind(intent);
+        manager.onServiceUnbounded();
+        return true;
```
Guichaguri: Can you leave the result from `super.onUnbind` as the return value?

egorkhmelev: Docs are pretty much clear about that. Default implementation returns false which prevents from call this onUnbind method second time and so on. So we forced to return true here.

https://developer.android.com/reference/android/app/Service.html#onUnbind(android.content.Intent)

Guichaguri: Fair enough.

Issue #94, opened 2017-12-13
Link media buttons directly to the playback


Guichaguri:
At the moment, all media controls fire events. Users have to listen to all events and then manually manipulate the player, like so:
```js
if(event.type == 'remote-play') {
    TrackPlayer.play();
} else if(event.type == 'remove-pause') {
    TrackPlayer.pause();
} ....
```

When I decided to design the API to work like this, I thought that someone might implement a different behavior to the media controls, but I don't think anyone has implemented anything different, it ended up being an annoying boilerplate code that every music app needs to implement the exactly same way.

The biggest problem with that are the issues that it has been causing:
* Being slow to process the action, since it has to go from the native code, to the event emitter, to the event listener, and finally back to the native code again. That can not only cause a small delay, but can also slow down the whole app, since JS runs in a single thread.
* In Android, when the app is in background, it needs to create a new service, start the JS engine, send the event, wait for the JS module to connect to the native player and then finally trigger the action. That might slow down the whole device, mainly when the device is low on memory.

I don't see any reason to keep them. If you really think these events shouldn't be removed, please, let me know why, I would really appreciate some feedback.

Guichaguri:
@dcvz @anderslemke @jasongrishkoff @egorkhmelev Any thoughts? I'd love to see some feedback on this

dcvz:
I would much prefer the media to be manipulated directly. I also don't see a downside to it unless our users are performing other actions upon these presses (analytics, etc). However, we could leave the events for users who wish to do other actions based on pressed but leave the playback manipulation to the library -- where it will be much faster to process.

Guichaguri:
If there is really a reason to keep them, we should definitely go with your idea, but even though someone might want to perform background actions (such as analytics), they can still use the `playback` events.

I think we should provide what most audio apps need in the most efficient and bug free way. If someone really needs an unusual feature, they can fork the project and implement themselves. We can also move it to an addon (the same thing we'll be doing with adaptive streaming and chromecast), so it's completely optional

dcvz:
Let's remove them and let the lib manage playback changes 👍 If we start hearing from our users that they're missing something we can think about how we can help them.

egorkhmelev:
[...]

PR #106, opened 2017-12-31
Change media buttons to directly affect the player

Guichaguri:
As discussed in #94, this PR removes the media button events and link them directly to the player.


ferrannp:
@Guichaguri do you plan to merge this? 😃

Guichaguri:
No, we didn't decided what we will be doing yet. If you want, you can join the discussion in #94.

dcvz:
@Guichaguri multiple users have requested that we keep the capability to forward remote events to JS. However as previously stated: I think there will be a huge benefit, especially on Android, to handle this directly from the player. So let's handle it from the player and allow the option for users to still get notified of changes.

I've updated the iOS side to always handle natively. I've also added a new key `forwardRemoteEventsToJS` that users can provide in the `updateOptions`. If that key is `true` then I'll also forward the event to JS.

dcvz:
I believe this was solved with release v1.0.0 so I'll close this now.

PR #112, opened 2018-01-08
Fix #105 where onTaskRemoved was no longer being called

### LINKED ISSUE #105: Audio Playback not stopping even after app close
App plays song even if app closed from task manager on Android
set stopWithApp flag to true but no luck

any idea how to stop music or close the app completely when closed from task manager?

puckey:
See more on the difference between passing `0` and `BIND_AUTO_CREATE` to `bindService` : https://stackoverflow.com/questions/14746245/use-0-or-bind-auto-create-for-bindservices-flag

Seeing as `Context.BIND_AUTO_CREATE` starts the service for us automatically, we need to do it manually by calling `context.startService(intent);`

### DIFF
--- a/android/src/main/java/guichaguri/trackplayer/TrackModule.java
+++ b/android/src/main/java/guichaguri/trackplayer/TrackModule.java
@@ -83,8 +83,9 @@ public class TrackModule extends ReactContextBaseJavaModule implements ServiceCo
 
         // Binds the service to get a MediaWrapper instance
         Intent intent = new Intent(context, PlayerService.class);
+        context.startService(intent);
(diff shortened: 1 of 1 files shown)


Guichaguri:
Thank you :)

donghungx:
Sorry but not perfect fix, when you open the App again it will show The playback is not initialized. You need to kill app once again before player back to work normal.

I tried to change code like this

if(manager.shouldStopWithApp()) {
            stopSelf();
            manager.destroyPlayer();
            stopForeground(true);
            manager.onServiceDestroy();
            manager = null;
}

But it force the app when I kill the app :(, btw when open app again player work

PR #118, opened 2018-01-15
MediaManager.java: add null check

puckey:
Fix crashes as described in https://github.com/react-native-kit/react-native-track-player/issues/117


Guichaguri:
Thank you :)

PR #129, opened 2018-01-27
Fix incorrect capability type and implement requiresMainQueueSetup

peterlazar1993:
This should fix https://github.com/react-native-kit/react-native-track-player/issues/109 and https://github.com/react-native-kit/react-native-track-player/issues/126.

Fix for 126 is based on https://github.com/wuxudong/react-native-charts-wrapper/blob/master/ios/ReactNativeCharts/pie/RNPieChartManager.swift#L14-L16. I am not familiar with swift so not sure if this is correct. But the library compiles for me and I am able to play tracks.


markv:
This branch is the only way I can get the dev branch to build - @dcvz

PR #145, opened 2018-02-12
Add ear-piece support

xotahal:
This pull request will add ability to play a track via ear-piece instead of lound-speaker.

Closes https://github.com/react-native-kit/react-native-track-player/issues/99

Note that it's my first time with Swift and maybe second time with Java :)

#### Use case:
When we use `react-native-track-player` for playing "chat message".

### DIFF
--- a/android/src/main/AndroidManifest.xml
+++ b/android/src/main/AndroidManifest.xml
@@ -3,6 +3,7 @@
           package="guichaguri.trackplayer">
 
     <uses-permission android:name="android.permission.WAKE_LOCK" />
+    <uses-permission android:name="android.permission.MODIFY_AUDIO_SETTINGS" />
 
     <application>
 
--- a/android/src/main/AndroidManifest.xml
+++ b/android/src/main/AndroidManifest.xml
@@ -3,7 +3,6 @@
           package="guichaguri.trackplayer">
 
     <uses-permission android:name="android.permission.WAKE_LOCK" />
-    <uses-permission android:name="android.permission.MODIFY_AUDIO_SETTINGS" />
 
     <application>
 
--- a/android/src/main/java/guichaguri/trackplayer/player/Playback.java
+++ b/android/src/main/java/guichaguri/trackplayer/player/Playback.java
@@ -134,7 +134,7 @@ public abstract class Playback {
 
     public abstract void play();
 
-    public abstract void playWithEarPiece();
+    public abstract void play(boolean earpiece);
 
     public abstract void pause();
 
--- a/android/src/main/java/guichaguri/trackplayer/player/Playback.java
+++ b/android/src/main/java/guichaguri/trackplayer/player/Playback.java
@@ -134,6 +134,8 @@ public abstract class Playback {
 
     public abstract void play();
 
+    public abstract void playWithEarPiece();
+
     public abstract void pause();
 
     public abstract void stop();
--- a/android/src/main/java/guichaguri/trackplayer/player/players/AndroidPlayback.java
+++ b/android/src/main/java/guichaguri/trackplayer/player/players/AndroidPlayback.java
@@ -38,6 +38,7 @@ public class AndroidPlayback extends Playback implements OnInfoListener, OnCompl
     private boolean buffering = false;
     private boolean ended = false;
     private boolean started = false;
+    private boolean earpiece = false;
 
     private int startPos = 0;
     private float buffered = 0;
@@ -90,6 +91,13 @@ public class AndroidPlayback extends Playback implements OnInfoListener, OnCompl
             if(player.isPlaying()) player.stop();
             player.reset();
             player.setDataSource(context, url);
+
+            if (earpiece == true) {
+                player.setAudioStreamType(AudioManager.STREAM_VOICE_CALL);
+            } else {
+                player.setAudioStreamType(AudioManager.STREAM_MUSIC);
+            }
+
             player.prepareAsync();
         } catch(IOException ex) {
             loadCallback = null;
@@ -119,7 +127,37 @@ public class AndroidPlayback extends Playback implements OnInfoListener, OnCompl
     public void play() {
         started = true;
 
+        if (earpiece == true) {
+            earpiece = false;
+
+            Track track;
+            track = super.getCurrentTrack();
+
+            load(track, null);
+        }
+
         if(!loaded) return;
+
+        player.start();
+
+        buffering = false;
+        ended = false;
+        updateState();
+    }
+    @Override
+    public void playWithEarPiece() {
+        started = true;
+
+        if (earpiece == false) {
+            earpiece = true;
+
+            Track track;
+            track = super.getCurrentTrack();
+
+            load(track, null);
+        }
+
+        if (!loaded) return;
 
         player.start();
 
--- a/android/src/main/java/guichaguri/trackplayer/player/players/AndroidPlayback.java
+++ b/android/src/main/java/guichaguri/trackplayer/player/players/AndroidPlayback.java
@@ -125,31 +125,14 @@ public class AndroidPlayback extends Playback implements OnInfoListener, OnCompl
 
     @Override
     public void play() {
-        started = true;
-
-        if (earpiece == true) {
-            earpiece = false;
-
-            Track track;
-            track = super.getCurrentTrack();
-
-            load(track, null);
-        }
-
-        if(!loaded) return;
-
-        player.start();
-
-        buffering = false;
-        ended = false;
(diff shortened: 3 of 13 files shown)


android/src/main/java/guichaguri/trackplayer/player/players/AndroidPlayback.java
```diff
@@ -90,6 +91,13 @@ public void load(Track track, Promise callback) {
             if(player.isPlaying()) player.stop();
             player.reset();
             player.setDataSource(context, url);
+
+            if (earpiece == true) {
+                player.setAudioStreamType(AudioManager.STREAM_VOICE_CALL);
+            } else {
+                player.setAudioStreamType(AudioManager.STREAM_MUSIC);
+            }
```
Guichaguri: Does it have set the stream type before it loads the track?

android/src/main/java/guichaguri/trackplayer/player/players/ExoPlayback.java
```diff
@@ -105,6 +105,10 @@ public void reset() {
     public void play() {
         player.setPlayWhenReady(true);
     }
+    @Override
+    public void playWithEarPiece() {
+        player.setPlayWhenReady(true);
```
Guichaguri: No support for ExoPlayer?
The "Android Player" is going to be removed in a future version, as ExoPlayer is much more reliable

android/src/main/AndroidManifest.xml
```diff
@@ -3,6 +3,7 @@
           package="guichaguri.trackplayer">
 
     <uses-permission android:name="android.permission.WAKE_LOCK" />
+    <uses-permission android:name="android.permission.MODIFY_AUDIO_SETTINGS" />
```
Guichaguri: Is this really required?

android/src/main/java/guichaguri/trackplayer/player/Playback.java
```diff
@@ -134,6 +134,8 @@ public void reset() {
 
     public abstract void play();
 
+    public abstract void playWithEarPiece();
```
Guichaguri: I think passing it as an option instead of a new function would be better

android/src/main/java/guichaguri/trackplayer/player/players/ExoPlayback.java
xotahal: Will take a look on this. Thanks

android/src/main/java/guichaguri/trackplayer/player/players/AndroidPlayback.java
xotahal: Not sure what do you think.

android/src/main/java/guichaguri/trackplayer/player/Playback.java
xotahal: I was thinking about this. I'm not sure if there will be more options in a future. But yeah, I will do it like you said. I also like the "parameter solution" ...

android/src/main/java/guichaguri/trackplayer/player/players/AndroidPlayback.java
xotahal: Well, if you ask about some default value. There is a line 41: `private boolean earpiece = false`. So STREAM_MUSIC will be set up by default .. Like it was before ..

xotahal:
- Added support for ExoPlayer. I upgraded version to 2.5.0 because `setAudioStreamType` is deprecated now. No breaking changes.
- I've remove useless permission for android.
- Used parameter instead of different names of function. Anyway I had to export to React with different names, because it seems it doesn't support more declarations of function with same names and different parameters

xotahal:
Any progress with this?

dcvz:
@xotahal sorry for the late review: right now it seems you're adding this as an extra method to play with earpiece. Don't you think it might make more sense to make this part of the configuration and let the regular `play()` handle whether you're in earpiece mode or not?

xotahal:
Actually it works like that. Internally it is just `private earpiece = true | false` and `playWithEarpiece` just set up this to `true`.

So instead of:
```js
TrackPlayer.setMode('earpiece')
TrackPlayer.play()
```

You can do just:
```js
TrackPlayer.playWithEarpiece()
```

So it's probably more question about one's preferences than about code quality.

Guichaguri:
I shouldn't be asking this this late, but why do you need an ear-piece support?

xotahal:
I wrote it in the first comment :)

> When we use react-native-track-player for playing "chat message".

So for example when someone uses the track-player for application similar to `Whatsapp` (or any chat app). The user of these apps can receive a sound message and when he put the phone closer to ear, the sound goes through ear-piece.

Is there any problem to make this library useful for bigger audience? :)

Guichaguri:
> Is there any problem to make this library useful for bigger audience? :)

No, the problem is that this library is not meant to be used by chat apps, it's fully built around the concept of music/podcast apps, creating services and native music controls that shouldn't be used by regular apps.

For regular apps, there's always `react-native-sound` and other similar projects, which are lightweight, simple audio players for apps that don't need background mode, music controls, smartwatch integration, etc

xotahal:
Why it shouldn't be used by regular apps? We need a background mode, music controls, smartwatch, etc ...

dcvz:
@xotahal I know internally it works that way, but I think it should also work that way externally. We should provide it as an option/configuration parameter and not part of the main API as this is not a standard use case.

moonstruck:
Basic use case for ear-piece support is play/pause while listening to music over bluetooth headphone.

bourgois:
+1 Support of ear-piece is great for our app too !

dcvz:
would you mind rebasing this @xotahal ? I think this would be a great a addition, I can help you finalise the last details ☀️

xotahal:
@dcvz Well, it is not in my roadmap right now. It works for us in our fork. Maybe later ...

dcvz:
Sure @xotahal 🙂

You may have to do this anyway if you update to v1.0.0 which we just released. If not, I can take a look at it in the future.

xotahal:
Understand ...

Guichaguri:
Since this has not been requested anymore and it would probably be easier to create a new PR instead of rebasing this one, I'll be closing this.

bourgois:
HI @Guichaguri 
We are using this in production for 2 years now with thousands of users. As you didn't want to merge the PR we are maintaining our own fork. Would love to offer a new PR if you would consider it!

Guichaguri:
@bourgois We would, but it will probably require a few API changes as the module changed quite a bit since this PR was first created.

bourgois:
@Guichaguri Great! We will work on that at some point during the summer. Our fork isn't based on this PR anymore anyway.

jordibruin:
@bourgois have you worked on this over the summer? Would be super useful for a project I'm working on right now, thanks!

MatthieuPeyrot:
@jordibruin 
We have work on it, you can check our new PR here 
https://github.com/react-native-kit/react-native-track-player/pull/759

regalstreak:
Any updates?

bourgois:
We didn't hear back on our PR. We are still using it in production...

andrekovac:
@bourgois Which branch are you using?

Can one set the default output to the loud speaker on that branch? Or can one only choose the earpeace?

I'm asking because I face the following issue: https://github.com/react-native-kit/react-native-track-player/issues/953

MatthieuPeyrot:
@andrekovac we are using this branch https://github.com/Voxist/react-native-track-player#fix-submodule , but i didn't work on it since July  so it's no up to date there is 97 commits behind. so i need to update my code. i will when i have time to do it. maybe fork again and rewrite it

andrekovac:
@MatthieuPeyrot Thanks! I took a look at your branch. I saw that I don't need such fine level control over the earpeace support. 

Btw, you can also use `expo-av` and its `playThroughEarpieceAndroid` config to set the audio output on Android.

Buthrakaur:
@MatthieuPeyrot thanks for providing the branch (fixed branch link: https://github.com/Voxist/react-native-track-player/commits/fix-post-install btw). Unfortunately the branch is based on 1.2.0 and track-player is now 2.x already. Anyway the audio device output selection is missing even for us - we're building a travel guide app and sometimes users (especially indoors or in a bus) just don't have headphones and are asking for possibility to switch to the earpiece. So this is yet another upvote @dcvz @bradleyflood

bradleyflood:
Thanks for the feedback. It's. or something we have capacity for But we are open to a new PR that is based on v2. 

Thanks all :)

PR #160, opened 2018-03-12
Enable Exoplayer 302 redirects involving HTTP/HTTPS

callmetwan:
By default Exoplayer rejects requests that redirect from HTTP to HTTPS or vice-versa. Exoplayer's DefaultHttpDataSourceFactory takes an argument that enables redirection behavior.

Also added .idea/ to .gitignore for anyone using JetBrains' products.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ build/
 DerivedData/
 package-lock.json
 yarn.lock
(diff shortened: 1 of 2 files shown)


callmetwan:
Indeed! Thank you for being so responsive!!

PR #173, opened 2018-04-04
Add Example Project and First Example

dcvz:
This adds a basic project setup with an example of using the queue setup and tracking progress via the `ProgressComponent` 💃 

This sets a place for users to submit more examples and I'll be adding more over time. It'll also be a good resource for reproducing and fixing reported bugs.


dcvz:
@Guichaguri can you verify this runs correctly on Android? I want to merge this ASAP so I can start working on more examples + bug fixes.

example/react/screens/PlaylistScreen.js
```diff
@@ -0,0 +1,87 @@
...
+import TrackStore, { playbackStates } from '../stores/Track';
+
+@observer
+export default class LandingScreen extends Component {
+  static navigationOptions = {
+    title: 'Playlist Example',
+  };
+
+  componentDidMount() {
+    TrackPlayer.setupPlayer();
```
Guichaguri: You should retrieve the player state or enable the [stopWithApp](https://github.com/react-native-kit/react-native-track-player/wiki/Documentation#updateoptionsdata) option

example/index.js
```diff
@@ -9,12 +9,13 @@ AppRegistry.registerComponent('example', () => App);
...
-    TrackStore.artist = track.artist;
-    TrackStore.artwork = track.artwork;
+    if (data.nextTrack) {
+      const track = await TrackPlayer.getTrack(data.nextTrack);
+      TrackStore.title = track.title;
+      TrackStore.artist = track.artist;
+      TrackStore.artwork = track.artwork;
+    }
   } else if (data.type === 'playback-state') {
-    PlayerStore.playbackState = data.state;
```
dcvz: Was this causing an issue on Android @Guichaguri ? this is what updates the state label in the playlist view.

Guichaguri: No, it doesn't make sense to set the playbackState here and then set it again right after it. (Lines 19 and 21)

PR #176, opened 2018-04-07
Fix Issues With Skipping & Playback

### LINKED ISSUE #146: iOS MediaWrapper.swift - play() invoked by playNext()/playPrevious()
Hi - this is an improvement suggestion.

On iOS, if I pause the player and then call TrackPlayer.skipToNext(), it starts playing the current track.

I checked MediaWrapper.swift and noticed an issue: Both playNext() and playPrevious() invokes play(). play() goes into this block:
```
if player.state == .paused {
    player.resume()
    return
}
```

So it never calls

```
let track = queue[currentIndex]
player.play(track: track)
```
which is the actual skipping part. For now, I created a method, playAfterSkip() in which I just entirely removed the first if block. Let me know if you agree to add this change or something based on this.

dcvz:
Would fix #146

### DIFF
--- a/android/src/main/java/guichaguri/trackplayer/logic/components/MediaWrapper.java
+++ b/android/src/main/java/guichaguri/trackplayer/logic/components/MediaWrapper.java
@@ -151,7 +151,7 @@ public class MediaWrapper extends Binder {
         Track track = pb.getCurrentTrack();
 
         if(track == null) {
-            Utils.rejectCallback(callback, "track", "No track playing");
+            Utils.resolveCallback(callback, null);
         } else {
             Utils.resolveCallback(callback, track.id);
         }
--- a/ios/RNTrackPlayer/Logic/MediaWrapper.swift
+++ b/ios/RNTrackPlayer/Logic/MediaWrapper.swift
@@ -42,7 +42,7 @@ class MediaWrapper: AudioPlayerDelegate {
     }
     
     var currentTrack: Track? {
-        return queue.indices.contains(0) ? queue[currentIndex] : nil
+        return player.currentItem
     }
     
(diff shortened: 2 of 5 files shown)


dcvz:
@Guichaguri including you because I changed the API for `getCurrentTrack`. I think it doesn't really make sense to reject the promise when nothing is playing. The user wishes to see what it is, I think returning `null` instead of rejecting is more aligned with what the user would expect.

osiloke:
Unfortunately, the same issue as #146 still occurs with this commit

willcosgrove:
I'm not a Swift developer, so I could be off base, but looking at this block:

https://github.com/react-native-kit/react-native-track-player/blob/7db804f3a128966b8f74c663e4766cd004c5fbb3/ios/RNTrackPlayer/Logic/MediaWrapper.swift#L172-L176

I see that we check that `currentTrack?.id == queue[currentIndex].id`.  But `currentTrack` is defined as the same thing:
https://github.com/react-native-kit/react-native-track-player/blob/7db804f3a128966b8f74c663e4766cd004c5fbb3/ios/RNTrackPlayer/Logic/MediaWrapper.swift#L48-L50

So that should always be true, which means it isn't adding any behavior to this if statement.

I don't know enough Swift to be able to comment on what a good solution would be.  Obviously, the `play` function needs to know if the track has been skipped, but there's a lot of different ways it could get that information.  At the moment, I'm working around this bug from the JS side, not issuing a command to skip if we're paused.  Ideally the player would handle this case properly, but because it is able to be worked around, I'm not desperate enough to learn Swift to submit a fix 😬 

But to any future devs looking to fix it, I think this comment basically outlines the issue.

PR #185, opened 2018-04-20
remote media control fixes

malangaveuday:
IOS on lock screen media control buttons are enabled now


Guichaguri:
This don't really fix the problem. You're just commenting the logic and hardcoding it.

PR #224, opened 2018-06-26
getduration and state.duration return 0 in android 7.1.2 working fine in android 6.0

krishna1012:
i tried progress component too both return 0 in android 7.1.2 on usb debugging and release apk testing.


Guichaguri:
Is this a bug report? It doesn't seem like a proper PR to me.

PR #235, opened 2018-07-25
IOS - Allow local files from static server.

tbatch92:
Local files might be located on a static server (eg using https://www.npmjs.com/package/react-native-static-server). In these cases the file will not be prefixed with file:// but are still playable with a network connection.

### DIFF
--- a/ios/RNTrackPlayer/Models/MediaURL.swift
+++ b/ios/RNTrackPlayer/Models/MediaURL.swift
@@ -12,11 +12,11 @@ struct MediaURL {
     let value: URL
     let isLocal: Bool
     private let originalObject: Any
-    
+
     init?(object: Any?) {
         guard let object = object else { return nil }
         originalObject = object
-        
+
         if let localObject = object as? [String: Any] {
             isLocal = true
             let uri = localObject["uri"] as! String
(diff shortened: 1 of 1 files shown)


Guichaguri:
Does iOS have a restriction to localhost URLs? I thought passing the URL directly would work as expected

tbatch92:
A use case of a recent project was playing downloaded HLS tracks. 

The problem we ran into was if completely offline (no wifi or mobile data) the track would not be played at all if you were "adding". However if the track was added whilst online, it would still play after the connection was turned off. 

Seems to be attributed to [line 95 here](https://github.com/react-native-kit/react-native-track-player/blob/dev/ios/RNTrackPlayer/Vendor/AudioPlayer/player/AudioPlayer.swift), which checks if you're online or if the URL is flagged as 'isLocal'.

ios/RNTrackPlayer/Models/MediaURL.swift
```diff
@@ -12,19 +12,19 @@ struct MediaURL {
...
+
         if let localObject = object as? [String: Any] {
             isLocal = true
             let uri = localObject["uri"] as! String
             let encodedURI = uri.addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed)!
             value = URL(string: encodedURI)!
         } else {
             let url = object as! String
-            isLocal = url.lowercased().hasPrefix("file://")
+            isLocal = url.lowercased().hasPrefix("file://") || url.lowercased().hasPrefix("http://localhost")
```
dcvz: What if someone was to use ‘https://localhost’?

Guichaguri: And how about `http://192.168.0.1` or `https://192.168.0.1`?

bradleyflood:
Closing this as outdated. Thanks all.

PR #251, opened 2018-08-16
Update license in package.json

lihop:
In 05b6bfe the license was changed from GPL-3.0 to Apache-2.0. The license field in package.json should be updated to reflect this.


Guichaguri:
I totally forgot about that, thanks :)

PR #266, opened 2018-09-04
Update compile SDK and target SDK to 26 and build tools to version 26.0.3

theeomm:


dcvz:
We will update to API 27 with https://github.com/react-native-kit/react-native-track-player/pull/165

PR #277, opened 2018-09-10
Allow setting rate while buffering in ios

biomancer:
Currently trying to set rate right after calling play is a noop if track is being buffered and has not started playing yet, this pr fixes that, so code like that would work correctly:
```
await TrackPlayer.play()
TrackPlayer.setRate(someRate)
```


dcvz:
Thanks for this @biomancer :) I agree, that should be the expected functionality.

PR #298, opened 2018-10-12
Fix android 5 and android 8 related issues

seb0zz:
This PR fixes / or finds a workaround for different cases in which Android 5 and Android 8 users are facing errors.

### DIFF
--- a/android/src/main/AndroidManifest.xml
+++ b/android/src/main/AndroidManifest.xml
@@ -7,14 +7,14 @@
     <application>
 
         <!-- The main service, handles playback, playlists, media buttons and chromecast -->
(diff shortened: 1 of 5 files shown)


Buthrakaur:
@seb0zz this should fix https://github.com/react-native-kit/react-native-track-player/issues/226 , right?

seb0zz:
@Buthrakaur yes it should :v: at least for us it does fix it.

dcvz:
Is this still necessary following the release of 1.0.0?

Guichaguri:
I don't think it is, I'll be closing it for inactivity. Feel free to reopen it if there are still anything that can be reused from this PR.

PR #318, opened 2018-10-27
[android] adding Ongoing to updateOptions

Drazail:
adding the option to make the notification onGoing, this should introduce a workaround for https://github.com/react-native-kit/react-native-track-player/issues/317 and https://github.com/react-native-kit/react-native-track-player/issues/307

```javascript
    TrackPlayer.updateOptions({
      onGoing: true,
    });
```

### DIFF
--- a/android/src/main/java/com/guichaguri/trackplayer/service/metadata/MetadataManager.java
+++ b/android/src/main/java/com/guichaguri/trackplayer/service/metadata/MetadataManager.java
@@ -131,6 +131,11 @@ public class MetadataManager {
                 for(int cap : compact) compactActions |= cap;
             }
         }
+        
+        // make the notification ongoing
+        if(options.getBoolean("onGoing") == true){
+            builder.setOngoing(true);
+        }
 
         // Update the color
         builder.setColor(options.getInt("color", NotificationCompat.COLOR_DEFAULT));


Guichaguri:
Preventing the user from dismissing the notification is not the desired workaround for this problem. We should be discussing other ways to fix the problem instead of changing the behavior.

Drazail:
of course this is not the desired solution, its not intended to be one, the workaround is a lucky side effects. however I think having the option to make the notification onGoing would be a nice addition.

if you feel the same let me know so i can add a false case to the onGoing option as well. ( to dynamically change the behavior back to default)

Guichaguri:
Google recommends the opposite, it's annoying to not be able to swipe a notification out that's not relevant anymore. We shouldn't be giving options to go against Android design guidelines.

> Ongoing notifications keep users informed about an ongoing process in the background. For example, music players announce the currently playing track in the notification system and continue to do so **until the user stops the playback**.
>
> Android Docs

I still don't like the `stopWithApp` option, it's against the expected media app behavior on Android, but I still keep it because it's easier for an react-native app, and a few users are already used to the same behavior on iOS. This one only works around one issue that can be solved in another manner.

Guichaguri:
Those issues should have been fixed by https://github.com/react-native-kit/react-native-track-player/commit/80b97301bf340fa29ab9d971b9781c81d578e243.

PR #321, opened 2018-10-29
Update Player & Refactor `

dcvz:
Updates player to the one written by @jorgenhenrichsen which I've started contributing too. It's almost everything which I'd already incorporated here but better.

**Breaking Changes**
- Can't remove currently playing track


dcvz:
@Guichaguri could you implement the change of event `playback-track-done`?

ios/RNTrackPlayer/RNTrackPlayer.swift
```diff
@@ -185,159 +187,184 @@ public class RNTrackPlayer: RCTEventEmitter, MediaWrapperDelegate {
...
     
     @objc(remove:resolver:rejecter:)
     public func remove(tracks ids: [String], resolve: RCTPromiseResolveBlock, reject: RCTPromiseRejectBlock) {
         print("Removing tracks:", ids)
-        mediaWrapper.removeTracks(ids: ids)
+        var indexesToRemove: [Int] = []
+        
+        for id in ids {
+            if let index = player.queueManager.items.firstIndex(where: { ($0 as! Track).id == id }) {
+                if index == player.queueManager.currentIndex { return }
```
Guichaguri: Is it not rejecting the promise? Which code should we use here?

anderslemke:
My immediate reaction is that I really like the `playback-track-done` event. ❤️
This will cover our use case just fine.

ios/RNTrackPlayer/RNTrackPlayer.swift
dcvz: I think we should not reject. We should just ignore `ids` that we don't find and if the `id` corresponds to the current track

Guichaguri:
Are we getting rid of the `nextTrack` property? I think it's useful for updating the UI accordingly.

Guichaguri:
I haven't tested yet, but here is what can happen if anything goes wrong:
* The notification may not update properly
* The event may not trigger in certain scenarios

I've also made the event backwards compatible in 27aac7b0170ca8ce8c835747c9d694cb4cebb626. Let me know if you agree with that.

dcvz:
@Guichaguri could you rebase the android files? Let's go ahead and add a `nextTrack` property that can be `null` when there's no next track.. also be careful about when the reason is `jumpedToIndex` -- as the nextTrack you'll have to fetch.

Guichaguri:
While I was testing it, I noticed a few reasons where `playback-track-changed` is better than `playback-track-done`:

* You don't receive the event when the first track starts.
* You can work around it by updating the state every time you call `play()`. Weirdly, the same technique can be used to know the reason why the track has changed.
* You can also use the `playback-state` event instead, which seems pretty out of place.
* I was planning it as the "metadata update" event, it would also be triggered when a cast device updates the current track.

Both solutions would result in a lot of unefficient boilerplate code. Here's an example (the code might be triggered and call these native functions multiple times during the same track):
```js
TrackPlayer.addEventListener('playback-state', async (data) => {
    PlayerState.state = data.state;
    const lastTrackId = PlayerState.track.id;
    const trackId = await TrackPlayer.getCurrentTrack();
    PlayerState.track = await TrackPlayer.getTrack(trackId);
    PlayerState.duration = await TrackPlayer.getDuration();
    // ...

    // maybe even a bit of logic
    if (data.state === TrackPlayer.STATE_PLAYING && lastTrackId != trackId) {
        // send data to the analytics server
    }
});
```

I think we should keep the `playback-track-changed` the way it is and add the new `reason` property. You'll still be able to know when and how a track has ended by checking the reason.

dcvz:
@Guichaguri I've kept events as are, so I rolled back your changes except for not allowing the currently playing track to be removed.

Issue #194, opened 2018-05-07
Updating Track Data


JDBlackman:
Is it possible  to be able to update the track data without playing the next track. I am using this to play a radio stream and since metadata is currently not supported, I would like to manually update the notification.

Guichaguri:
Duplicate of #93 -- Or even #192

Issue #192, opened 2018-05-04
Add support to icy streams


mydesweb:
It would be nice to have a way to display the song name and artist currently playing from a stream. thank you

I saw this feature in [https://github.com/tlenclos/react-native-audio-streaming](https://github.com/tlenclos/react-native-audio-streaming)

Issue #93, opened 2017-12-08
metadata QUESTION


holospeed:
Hi!

I am wondering, is it possible to change (update) the artwork, artist and title during the track is playing ? 
I love to play internet radio stream on my app and change the artist data (title, artist, artwork) when a new song starts.

Guichaguri:
Not yet, but I'll consider it as a suggestion :)

puckey:
For mp3 streams, we could use something like this – but it does involve changes to the internals of the ExoPlayer library.: https://github.com/google/ExoPlayer/issues/466#issuecomment-290358993

adidoes:
Any updates on this?

questionablequestion:
Bump. Any update?

cgomezmendez:
No need for internal modifications of exoplayer anymore it seems...

Grandmac:
It would be a great feature ! any update ?

PR #331, opened 2018-11-02
Add the option to update trackMetadata

HybridFox:
Works for android & iOS. Handy for icecast / shoutcast / etc... audio streams.

#194 #93 #192


leonhh:
Any chance this can get merged soon?

HybridFox:
Is there a possibility for an update on this?

Guichaguri:
I'm waiting for a review on the iOS part from @dcvz, once he gives a green light, we can merge it.

ios/RNTrackPlayer/Logic/MediaWrapper.swift
```diff
@@ -112,6 +112,14 @@ class MediaWrapper: AudioPlayerDelegate {
         }
     }
     
+    func updateMetadata(_ item: Track) {
+        MPNowPlayingInfoCenter.default().ap_update(
+            with: item,
+            duration: 100,
```
dcvz: we should have track duration from the item, right? perhaps even progression can be inferred here.

Guichaguri:
You're dynamically sending metadata updates, don't you think it would be better if you updated the current track object instead?

I was thinking about something like `updateTrack(id, obj)`, which allows you to update only metadata-related properties. That way, it not only allows you to update metadata from other tracks in the queue, but also keeps the metadata changes through any playback updates.

For instance, if you have used `updateMetadata` and the track is stopped and started again, you'll have to use `updateMetadata` again. The same can happen with any other playback update.

PR #358, opened 2018-12-03
Cache queue to allow resume after service is killed

antontanderup:
Caches queue when service is killed.

Works for our setup, but we only ever have one track in the queue, so it might need some work for others.

I had trouble understanding the notifications, so i simply replaced the notification after it is removed. 
This causes a brief moment where the notification is gone and replaced, when resuming the playback and when the service is killed.


benjipott:
Really good feature , @Guichaguri can take a look about this feature.

davidsalib:
thanks for this - curious how you dealt with the issue on iOS? 
I've actually not had this issue on Android, but I have on iOS.

antontanderup:
I'm closing this pull request. It was only reliable for our use case but not everyones, and track player went another direction than this anyway.

Issue #359, opened 2018-12-04
duration is always zero in iOS


alikazemkhanloo:
I'm streaming audio from direct links. for the progress bar, I use ProgressComponent. In iOS I get duration = 0, while it's working fine in android.
RNTP version 1.0.0
RN version 0.57.4

gazedash:
For me it is reverse: it is 0 on android, but works on iOS.

I have provided the duration for the track, I have awaited for player setup and add, and only then play.

dcvz:
@alikazemkhanloo could you try with this commit? 4c0483332779995e98e5f9164ad3c7d548f46615

ishigamii:
@alikazemkhanloo  can you try using the new version and close this issue if everything is good now ?

alikazemkhanloo:
Ok I will check it out ASAP.  But it may take couple of days.

ishigamii:
No problem ;) should work now take your time and close the issue if done ;)

ishigamii:
@alikazemkhanloo did you have time to check it out ? otherwise can you close it and open a new issue when you get to check it with latest ? 👍

alikazemkhanloo:
OK, I'll close it.
Sorry, I'm too busy to update my old project.

PR #360, opened 2018-12-05
get duration from track

alikazemkhanloo:
duration when playing from URL is always zero in iOS, as it happened for me here #359, if you could replicate this issue, then you could update duration from track info.
this PR checks if the duration is zero or not, if so, gets the duration from track's config.


Guichaguri:
This doesn't solve the problem at all, it just works around a bug that needs to be fixed properly. A fix like this should be implemented directly into the native code only when there's nothing that can be done.

You can use this code in your own app as a quick fix, but it won't be merged. Feel free to share it in the issue.

PR #382, opened 2019-01-05
ICY/Icecast/Shoutcast metadata [Android]

xgustavoh:
First of all I apologize for my English. I'm Brazilian, and my English is not very good at the moment, so I'll be using Google Translate.

I was doing the application on my radio and found problems with metadata.

I used the code created by @saschpe ([android-exoplayer2-ext-icy](https://github.com/saschpe/android-exoplayer2-ext-icy) or [ExoPlayer](https://github.com/saschpe/ExoPlayer/tree/icy)), and I integrated it with the code created by @HybridFox (#331).

I created two new events:

- **icy-header-update**: This event is issued when the icecast Header is received.
- **icy-metadata-update**: This event is issued when meta-data (stream) is updated.

> I updated the demo file to show how it works.

Currently I do not have an IOS, so I could not do the development, just implement the code created by @HybridFox.

I also did not make the best implementation possible because I was "tired" with the problem I was having. So the solution I was able to create was this.

Note:
- The duration of the song is not something important when it is streaming, because the music is an endless transmission. And some radio does not have the information of the duration of the music in its meta-dates.
- If you find any errors, I will be very happy to help you correct the same.
- If you need help or hint at the subject of audio streaming, I would consider making test streaming available to you.


android/src/main/java/com/guichaguri/trackplayer/service/player/IcyStreamUpdate.java
```diff
@@ -0,0 +1,52 @@
...
+            if(IcyStreamUpdate.service != null) {
+                
+                Bundle bundle = new Bundle();
+                bundle.putString("streamTitle", icyMetadata.getStreamTitle());
+                bundle.putString("streamUrl", icyMetadata.getStreamUrl());
+                IcyStreamUpdate.service.emit(MusicEvents.ICY_MATADATA_UPDATE, bundle);      
+            }    
+        }
+    };
+}
```
Guichaguri: This class seems totally unnecessary, it seems to much code to simply handle two events.
I also don't like the fact that it is static, as you can easily leak memory.

android/src/main/java/com/guichaguri/trackplayer/service/MusicBinder.java
```diff
@@ -16,6 +18,8 @@
     public MusicBinder(MusicService service, MusicManager manager) {
         this.service = service;
         this.manager = manager;
+        
+        IcyStreamUpdate.setService(service);
```
Guichaguri: Here's why you shouldn't make it a static class: You're leaking memory. You're storing the service in a field, but not removing it later.
That's why all the classes used by the service are instances, so they can be removed by the garbage collector when the service stops.

example/App.js
```diff
@@ -32,11 +32,64 @@ export default class App extends Component {
...
+      Track.title = music;
+      Track.artist = singer;
+      TrackPlayer.updateMetadata(Track);
+      
+      TrackStore.title = Track.title;
+      TrackStore.artist = Track.artist;
+
+      const params = this.findGetParameter(streamUrl);
+      console.log("MetadataUpdate, Url params:", params, "Stream Title:", streamTitle);
+    }
```
Guichaguri: I don't think this type of code fits with the example app. It's meant to be a simple app showing the basic usage of the module.

android/src/main/java/com/guichaguri/trackplayer/module/MusicEvents.java
```diff
@@ -36,6 +36,11 @@
     public static final String PLAYBACK_ERROR = "playback-error";
     public static final String PLAYBACK_UNBIND = "playback-unbind";
 
+
+    // ICY METADATA Events
+    public static final String ICY_HEADER_UPDATE = "icy-header-update";
+    public static final String ICY_MATADATA_UPDATE = "icy-metadata-update";
```
Guichaguri: Are those two events necessary? I mean, they could be implemented as one, such as `icy-update`.

android/src/main/java/com/guichaguri/trackplayer/service/models/Track.java
```diff
@@ -88,6 +97,7 @@ public Track(Context context, Bundle bundle, int ratingType) {
 
         userAgent = bundle.getString("userAgent");
         artwork = Utils.getUri(context, bundle, "artwork");
+        icymetadata = bundle.getBoolean("icymetadata");
```
Guichaguri: Why not add one new track `type` value?

android/src/main/java/com/guichaguri/trackplayer/service/models/Track.java
```diff
@@ -166,20 +176,27 @@ public DataSource createDataSource() {
             // Creates a local source factory
             ds = new DefaultDataSourceFactory(ctx, userAgent);
 
+        } else if ( icymetadata ) {
```
Guichaguri: Please, follow the same code style as the rest of the project.

android/src/main/java/com/guichaguri/trackplayer/service/models/Track.java
```diff
@@ -166,20 +176,27 @@ public DataSource createDataSource() {
...
             // Creates a default http source factory, enabling cross protocol redirects
             ds = new DefaultHttpDataSourceFactory(
                     userAgent, null,
                     DefaultHttpDataSource.DEFAULT_CONNECT_TIMEOUT_MILLIS,
                     DefaultHttpDataSource.DEFAULT_READ_TIMEOUT_MILLIS,
                     true
             );
-
             ds = playback.enableCaching(ds);
-
```
Guichaguri: Don't remove those empty lines, they are meant to help reading the code.

android/src/main/java/com/guichaguri/trackplayer/service/models/Track.java
```diff
@@ -195,4 +212,5 @@ public DataSource createDataSource() {
                         .createMediaSource(uri);
         }
     }
-}
+
```
Guichaguri: And also, don't add new unnecessary lines

android/src/main/java/com/guichaguri/trackplayer/service/models/TrackType.java
```diff
@@ -25,7 +25,6 @@
      */
     SMOOTH_STREAMING("smoothstreaming");
 
-
```
Guichaguri: Same here.

android/src/main/java/com/guichaguri/trackplayer/service/MusicManager.java
```diff
@@ -197,6 +197,10 @@ public void onTrackUpdate(Track previous, long prevPos, Track next) {
         service.emit(MusicEvents.PLAYBACK_TRACK_CHANGED, bundle);
     }
 
+    public void updateMetadata(Track track) {
+        metadata.updateMetadata(track);
+    }
+
```
Guichaguri: Please, don't merge another PR that has not been approved yet into yours. There are a few problems with its code that should be fixed first. And also, you're removing @HybridFox's authored commits.

xgustavoh: The comits created by @HybridFox, I can not get access to merge.

android/src/main/java/com/guichaguri/trackplayer/service/models/Track.java
xgustavoh: Sorry, custom to remove the end lines and keep in the beginning.

android/src/main/java/com/guichaguri/trackplayer/service/models/Track.java
xgustavoh: In HLS it is possible to obtain a metadata. It is then not influenced by the `type` of transmission.

android/src/main/java/com/guichaguri/trackplayer/service/player/IcyStreamUpdate.java
xgustavoh: I would really like to not use static variables, due to the problem you have mentioned yourself, but since I did not find a way to get the Service through a Track, I ended up doing this way, I can review the code and find other wiggles, in my GIT in a Brach I did not Push in which I do not use static variable, but I do not yet attribute the value of `service`.

android/src/main/java/com/guichaguri/trackplayer/module/MusicEvents.java
xgustavoh: It could be implemented through `icy-update`, but as` icy-metadata-update` is received several times, and `icy-header-update` only in connection, to avoid a further check, I decided to separate in this way.

android/src/main/java/com/guichaguri/trackplayer/service/player/IcyStreamUpdate.java
Guichaguri: Look into `MusicManager`, it stores instances for a lot of classes, and this one should also belong there.

android/src/main/java/com/guichaguri/trackplayer/module/MusicEvents.java
Guichaguri: You can add a property to tell which type of data you're receiving, but I don't think having two different events accomplish anything.

android/src/main/java/com/guichaguri/trackplayer/service/models/Track.java
Guichaguri: I mean, a HLS stream can't be an icy stream and an icy stream can't be a HLS stream. Both protocols are not supported together, so it can only be one or another.

android/src/main/java/com/guichaguri/trackplayer/service/MusicManager.java
Guichaguri: You can fork his repository and then merge into yours, that would keep his name in the commits, but I don't think you should do that, as his PR still has problems that must be fixed first.

xgustavoh: Perfect! I will make the corrections you commented on, in fact some are already done. If you feel better, I will "Cancel" this "Pull Request" and I will send you with the corrections and without the use of static variable.

Issue #418, opened 2019-01-29
seekTo android restarts the track


gazedash:
Possibly related to https://github.com/react-native-kit/react-native-track-player/issues/348

The bug is reproduced on Nexus 5X API 28 x86 android virtual device emulator.
Possibly reproducible on real device(?)

Works on iOS fine though.

Guichaguri:
What track are you playing? Are you receiving `playback-error` events?

gazedash:
@Guichaguri 
I receive two playback-state events:
{state: 6}
{state: 3}

I am not an android expert but is this related to PlaybackStateCompat https://developer.android.com/reference/android/support/v4/media/session/PlaybackStateCompat ?
As I can see, there is multiple events tied to same int values. E.g. 3 is REPEAT_MODE_GROUP, STATE_PLAYING and ERROR_CODE_AUTHENTICATION_EXPIRED

Guichaguri:
It is, but you should be checking using our constants (e.g. `event.state === TrackPlayer.STATE_PLAYING`) instead, as the values will change depending on the version and the platform.

gazedash:
@Guichaguri I am able to reproduce this error on real device Android 4.4.2

gazedash:
Maybe it is related to a bug where duration is 0?

gazedash:
It does reproduce even with example tracks:
https://github.com/react-native-kit/react-native-track-player/blob/dev/example/react/data/playlist.json

Guichaguri:
That seems to be it. I'll take a look later today

PR #425, opened 2019-02-01
Feature: enable seeking on mp3's with variable bitrate

gazedash:
#418 

https://google.github.io/ExoPlayer/faqs.html#why-are-some-media-files-not-seekable


Fariha-Khalid:
seekto restarts the track on android ..i am not able to solve it

t-tornado:
@Guichaguri, any solution for the seeking problem on android?

PR #427, opened 2019-02-02
Fix unresolved remove promise

retyui:
When trying to remove current played track returned promise never resolved


android/src/main/java/com/guichaguri/trackplayer/service/player/LocalPlayback.java
```diff
@@ -106,10 +107,15 @@ public void remove(List<Integer> indexes, Promise promise) {
...
                 source.removeMediaSource(index, Utils.toRunnable(promise));
+                promiseWasCalled = true;
             } else {
                 source.removeMediaSource(index, null);
             }
         }
+        
+        if(!promiseWasCalled){
+            promise.resolve(null);
+        }
```
Guichaguri: It seems unnecessary adding a flag to check whether the promise was already resolved

PR #441, opened 2019-02-12
[Android] Use root project's sdk versions if specified

timmywil:
This is a popular strategy for react native plugins and fixes several issues. I recently upgraded react native and the default sdk for react native 0.58.4 is 28. Rather than updating it manually in RNTP, I suggest using whatever sdk is being used in the parent project. For my purposes, RNTP seems to work fine with 28. Projects still on 27 are not affected by this change.

Also, each sdk has a default build tools version associated with it. Sometimes the buildToolsVersion setting is completely ignored if it doesn't match the sdk. There's no need to set it.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -22,13 +22,16 @@ if(file.exists()) {
     smoothstreaming = json.smoothstreaming ?: smoothstreaming
 }
 
+def safeExtGet(prop, fallback) {
+    rootProject.ext.has(prop) ? rootProject.ext.get(prop) : fallback
+}
+
 android {
-    compileSdkVersion 27
-    buildToolsVersion "27.0.3"
+    compileSdkVersion safeExtGet("compileSdkVersion", 27)
 
     defaultConfig {
-        minSdkVersion 16 // RN's minimum version
-        targetSdkVersion 27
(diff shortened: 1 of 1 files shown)


android/build.gradle
```diff
@@ -22,13 +22,16 @@ if(file.exists()) {
...
 android {
-    compileSdkVersion 27
-    buildToolsVersion "27.0.3"
+    compileSdkVersion safeExtGet("compileSdkVersion", 27)
 
     defaultConfig {
-        minSdkVersion 16 // RN's minimum version
-        targetSdkVersion 27
+        minSdkVersion safeExtGet('minSdkVersion', 16) // RN's minimum version
+        targetSdkVersion safeExtGet("compileSdkVersion", 27)
```
Guichaguri: How about the support library? It needs to match the compleSdkVersion, right?

timmywil: Honestly, I'm not sure. I assumed the support library could be top-level only.

timmywil: I'll go ahead and add it

timmywil: Added. I didn't mean to set `compileSdkVersion` for `targetSdkVersion`. Changed that to `targetSdkVersion` and used `supportLibVersion` for support libraries.

Guichaguri:
LGTM. Thank you.

PR #456, opened 2019-02-19
Check for playWhenReady

### LINKED ISSUE #452: [iOS] Player doesn't load songs when simulating slow internet connection
Hey,
I'm currently having an issue when I use the Network Link Conditioner with "Edge" setting. If I enable the slow internet connection, the player doesn't play any songs, even if they are saved locally. I've tried it with several local saved songs, and I'm only able to play one songs, after loading the second, the player stops updating any information and stops playing music.

dnish:
We should also check for <code>playWhenReady = true</code> within the add function, otherwise the player will always load the first item. This could be a problem when you have a slow internet connection. Fixes #452


dnish:
Just noticed that this will causes problems if you want to play the first item, because it wasn't loaded. I think we should add the abilitiy to addTracks, which item id we want to load/play after adding.

bradleyflood:
Thanks all. Closing this as outdated. Cheers.

Issue #469, opened 2019-03-01
[iOS] add Category option to allow custom AVAudioSession category


timmywil:
Category is set here: https://github.com/react-native-kit/react-native-track-player/blob/686e7a4d9fe6607d58347c573ba83ed6010d7ee5/ios/RNTrackPlayer/RNTrackPlayer.swift#L134

I'm using another plugin that uses the microphone and because AVAudioSession is shared, it needs to set `playAndRecord` as the category. Obviously, it doesn't seem appropriate to just set that here. It would be ideal if this was configurable somehow in react-native-track-player.

Here's a good explanation of the issue: https://github.com/wenkesj/react-native-voice/issues/50#issuecomment-347066628

react-native-sound `setCategory` method: https://github.com/zmxv/react-native-sound/blob/master/RNSound/RNSound.m#L110

I envision an option passed to `setupPlayer` rather than a method like the above. Maybe not `updateOptions` as it seems to be more work to change the category after the audio session becomes active.

curiousdustin:
This was addressed in #484.

PR #470, opened 2019-03-01
[iOS] allow configuring the category for the shared AVAudioSession instance

timmywil:
I'm new to swift so I could have overlooked something or gotten the syntax wrong, but trying to make this change as painless as possible. It shouldn't affect any existing code, and seems to work for my purposes. I considered adding options for `mode` and `options` too, but thought I'd keep it simple. I'm using it like this:

```js
TrackPlayer.setupPlayer({ iosCategory: 'playAndRecord' }).then(...
```

I'd also be willing to add this option to the wiki if this is approved.

**Edit:** I've only added the categories that include playback (i.e. ones that should work with RNTP).

- I also noticed while testing that `resolve` was getting called even after `reject` was called, which was throwing its own error.

Fixes gh-469

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -69,6 +69,7 @@ declare namespace RNTrackPlayer {
     maxBuffer?: number;
     playBuffer?: number;
     maxCacheSize?: number;
+    iosCategory?: 'playback' | 'playAndRecord' | 'multiRoute';
   }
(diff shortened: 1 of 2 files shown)


timmywil:
Closing in favor of #484

PR #484, opened 2019-03-10
[iOS] Move audio session category setting to `play()`

thessler27:
When the audio session is created on the `setupPlayer()` method, entering the application causes current background music to end. This will set the category on `play()` and end the background audio only when necessary

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -69,9 +69,9 @@ declare namespace RNTrackPlayer {
     maxBuffer?: number;
     playBuffer?: number;
     maxCacheSize?: number;
-    iosCategory?: 'playback'|'playAndRecord'|'multiRoute'|'ambient'|'soloAmbient'|'record';
-    iosCategoryOptions?: 'mixWithOthers'|'duckOthers'|'interruptSpokenAudioAndMixWithOthers'|'allowBluetooth'|'allowBluetoothA2DP'|'allowAirPlay'|'defaultToSpeaker';
-    iosCategoryMode?: 'default'|'gameChat'|'measurement'|'moviePlayback'|'spokenAudio'|'videoChat'|'videoRecording'|'voiceChat'|'voicePrompt';
+    iosCategory?: 'playback' | 'playAndRecord' | 'multiRoute' | 'ambient' | 'soloAmbient' | 'record';
+    iosCategoryOptions?: 'mixWithOthers' | 'duckOthers' | 'interruptSpokenAudioAndMixWithOthers' | 'allowBluetooth' | 'allowBluetoothA2DP' | 'allowAirPlay' | 'defaultToSpeaker';
+    iosCategoryMode?: 'default' | 'gameChat' | 'measurement' | 'moviePlayback' | 'spokenAudio' | 'videoChat' | 'videoRecording' | 'voiceChat' | 'voicePrompt';
   }
 
   export interface MetadataOptions {
--- a/index.d.ts
+++ b/index.d.ts
@@ -69,6 +69,7 @@ declare namespace RNTrackPlayer {
     maxBuffer?: number;
     playBuffer?: number;
     maxCacheSize?: number;
+    iosCategory?: 'playback' | 'playAndRecord' | 'multiRoute';
   }
 
   export interface MetadataOptions {
--- a/index.d.ts
+++ b/index.d.ts
@@ -69,7 +69,9 @@ declare namespace RNTrackPlayer {
     maxBuffer?: number;
     playBuffer?: number;
     maxCacheSize?: number;
-    iosCategory?: 'playback' | 'playAndRecord' | 'multiRoute';
+    iosCategory?: 'playback'|'playAndRecord'|'multiRoute'|'ambient'|'soloAmbient'|'record';
+    iosCategoryOptions?: 'mixWithOthers'|'duckOthers'|'interruptSpokenAudioAndMixWithOthers'|'allowBluetooth'|'allowBluetoothA2DP'|'allowAirPlay'|'defaultToSpeaker';
+    iosCategoryMode?: 'default'|'gameChat'|'measurement'|'moviePlayback'|'spokenAudio'|'videoChat'|'videoRecording'|'voiceChat'|'voicePrompt';
   }
 
   export interface MetadataOptions {
--- a/ios/RNTrackPlayer/Models/SessionCategories.swift
+++ b/ios/RNTrackPlayer/Models/SessionCategories.swift
@@ -1,0 +1,89 @@
+//
+//  SessionCategory.swift
+//  RNTrackPlayer
+//
+//  Created by Thomas Hessler on 3/12/19.
+//  Copyright © 2019 David Chavez. All rights reserved.
+//
+
+import Foundation
+import MediaPlayer
+import AVFoundation
+
+enum SessionCategory : String {
+    
+    case playAndRecord, multiRoute, playback, ambient, soloAmbient
+    
+    func mapConfigToAVAudioSessionCategory() -> AVAudioSession.Category {
+        switch self {
+        case .playAndRecord:
+            return .playAndRecord
+        case .multiRoute:
+            return .multiRoute
+        case .playback:
+            return .playback
+        case .ambient:
+            return .ambient
+        case .soloAmbient:
+            return .soloAmbient
+        }
+    }
+}
+
+enum SessionCategoryOptions : String {
+    
+    case mixWithOthers, duckOthers, interruptSpokenAudioAndMixWithOthers, allowBluetooth, allowBluetoothA2DP, allowAirPlay, defaultToSpeaker
+
+    func mapConfigToAVAudioSessionCategoryOptions() -> AVAudioSession.CategoryOptions? {
+        switch self {
+        case .mixWithOthers:
+            return .mixWithOthers
+        case .duckOthers:
+            return .duckOthers
+        case .interruptSpokenAudioAndMixWithOthers:
+            return .interruptSpokenAudioAndMixWithOthers
+        case .allowBluetooth:
+            return .allowBluetooth
+        case .allowBluetoothA2DP:
+            return .allowBluetoothA2DP
+        case .allowAirPlay:
+            return .allowAirPlay
+        case .defaultToSpeaker:
+            return .defaultToSpeaker
+        }
+    }
+}
+
+enum SessionCategoryMode : String {
+    
+    case `default`, gameChat, measurement, moviePlayback, spokenAudio, videoChat, videoRecording, voiceChat, voicePrompt
+    
+    func mapConfigToAVAudioSessionCategoryMode() -> AVAudioSession.Mode {
+        switch self {
+        case .default:
+            return .default
+        case .gameChat:
+            return .gameChat
+        case .measurement:
+            return .measurement
+        case .moviePlayback:
+            return .moviePlayback
+        case .spokenAudio:
+            return .spokenAudio
+        case .videoChat:
+            return .videoChat
+        case .videoRecording:
+            return .videoRecording
+        case .voiceChat:
+            return .voiceChat
+        case .voicePrompt:
+            if #available(iOS 12.0, *) {
+                return .voicePrompt
+            } else {
+                // Do Nothing
+                return .default
+            }
+        }
+    }
+}
+
--- a/ios/RNTrackPlayer/RNTrackPlayer.swift
+++ b/ios/RNTrackPlayer/RNTrackPlayer.swift
@@ -133,73 +133,23 @@ public class RNTrackPlayer: RCTEventEmitter, AudioPlayerDelegate {
     
     @objc(setupPlayer:resolver:rejecter:)
     public func setupPlayer(config: [String: Any], resolve: RCTPromiseResolveBlock, reject: RCTPromiseRejectBlock) {
-        
+
+        //configure base category -- defaults to .playback
         let sessionCategory: String = config["iosCategory"] as? String ?? "playback"
-        switch sessionCategory {
-            case "playAndRecord":
-                self.sessionCategory = .playAndRecord
-            case "multiRoute":
-                self.sessionCategory = .multiRoute
-            case "playback":
-                self.sessionCategory = .playback
-            case "ambient":
-                self.sessionCategory = .ambient
-            case "soloAmbient":
-                self.sessionCategory = .soloAmbient
-            default:
-                self.sessionCategory = .playback
-        }
+        self.sessionCategory = SessionCategory(rawValue: sessionCategory)!.mapConfigToAVAudioSessionCategory()
         
-        let sessionCategoryOpts: String = config["iosCategoryOptions"] as? String ?? ""
-        
-        switch sessionCategoryOpts {
-        case "mixWithOthers":
-            self.sessionCategoryOptions = .mixWithOthers
-        case "duckOthers":
-            self.sessionCategoryOptions = .duckOthers
-        case "interruptSpokenAudioAndMixWithOthers":
-            self.sessionCategoryOptions = .interruptSpokenAudioAndMixWithOthers
-        case "allowBluetooth":
-            self.sessionCategoryOptions = .allowBluetooth
-        case "allowBluetoothA2DP":
-            self.sessionCategoryOptions = .allowBluetoothA2DP
-        case "allowAirPlay":
-            self.sessionCategoryOptions = .allowAirPlay
-        case "defaultToSpeaker":
-            self.sessionCategoryOptions = .defaultToSpeaker
-        default:
-            //do nothing
+        //configure extra opts -- defaults to nil
+        let sessionCategoryOpts: String? = config["iosCategoryOptions"] as? String ?? nil
+        //this is because this doesnt default as anything
+        if(sessionCategoryOpts != nil) {
+            self.sessionCategoryOptions = SessionCategoryOptions(rawValue: sessionCategoryOpts!)!.mapConfigToAVAudioSessionCategoryOptions()
+        } else {
             self.sessionCategoryOptions = nil
         }
-        
+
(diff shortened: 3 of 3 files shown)


timmywil:
This would conflict with my proposal for a category option in iOS: https://github.com/react-native-kit/react-native-track-player/pull/470. I took the approach of react-native-sound, providing a way to set categories that all work with react-native-track-player. Unfortunately, neither of our PRs handles both of our cases.
However, I've become more familiar with how all this works and RNTP *could* take an approach more like react-native-video, where category could be set to something like "inherit" and not override the category at all. This PR is more in that vein and I think the "inherit" option would handle both of our cases.

thessler27:
@timmywil Agreed. With moving category setting to `play` we could also include your category choosing configuration so we could offer that functionality as well. Shall I add your updates to my PR?

timmywil:
@thessler27 That should work. 👍  My only concern is AVAudioSession doesn't like changing the category when active so if changing the category during 'play', it might look more like this...

```js
try AVAudioSession.sharedInstance().setActive(false)
try AVAudioSession.sharedInstance().setCategory(category, mode: .default)
try AVAudioSession.sharedInstance().setActive(true)
```

dcvz:
I think many people in their applications setup the player at the start of their application, just out of how people architect their apps.

I find it very annoying to open and app and have it immediately stop my current music. I want that to be an explicit action.

_free styling ideas:_
What we could do is have you set up the category type on setup, but only apply & make active -- when we're ready to play.

timmywil:
@dcvz I'm fine with that; it just wouldn't support changing the category in `play`.

thessler27:
I've added it. You can set the options in `setupPlayer` as written in #470 and then it'll utilize those options in `play()`

timmywil:
I plugged these changes into my app and they look good. 👍

thessler27:
Hi @dcvz  and @timmywil. I've updated the configuration to support your recommendations. I've also included 2 more options for more dynamic option choosing (audio ducking, for example). These are `iosCategoryOptions` and `iosCategoryMode`. This will give us much more control over the session. I've tested this locally and it works. Let me know if you run into issues or want different changes. Thanks

index.d.ts
```diff
@@ -69,6 +69,9 @@ declare namespace RNTrackPlayer {
     maxBuffer?: number;
     playBuffer?: number;
     maxCacheSize?: number;
+    iosCategory?: 'playback'|'playAndRecord'|'multiRoute'|'ambient'|'soloAmbient'|'record';
```
dcvz: nitpick: let's keep the space between `|`.

ios/RNTrackPlayer/RNTrackPlayer.swift
```diff
@@ -130,11 +133,72 @@ public class RNTrackPlayer: RCTEventEmitter, AudioPlayerDelegate {
     
     @objc(setupPlayer:resolver:rejecter:)
     public func setupPlayer(config: [String: Any], resolve: RCTPromiseResolveBlock, reject: RCTPromiseRejectBlock) {
-        do {
-            try AVAudioSession.sharedInstance().setCategory(.playback, mode: .default)
-            try AVAudioSession.sharedInstance().setActive(true)
-        } catch {
-            reject("setup_audio_session_failed", "Failed to setup audio session", error)
+        
+        let sessionCategory: String = config["iosCategory"] as? String ?? "playback"
```
dcvz: another comment: this has all gotten quite big. Let's take the approach we've taken for  `Capabilities` & `PitchAlgorithms` on these.

You can see the examples here: https://github.com/react-native-kit/react-native-track-player/tree/dev/ios/RNTrackPlayer/Models

thessler27:
@dcvz let me know if this is what you were thinking. I tested it locally and it worked for me. Thanks

ios/RNTrackPlayer/RNTrackPlayer.swift
```diff
@@ -130,12 +133,23 @@ public class RNTrackPlayer: RCTEventEmitter, AudioPlayerDelegate {
...
     public func setupPlayer(config: [String: Any], resolve: RCTPromiseResolveBlock, reject: RCTPromiseRejectBlock) {
-        do {
-            try AVAudioSession.sharedInstance().setCategory(.playback, mode: .default)
-            try AVAudioSession.sharedInstance().setActive(true)
-        } catch {
-            reject("setup_audio_session_failed", "Failed to setup audio session", error)
+
+        //configure base category -- defaults to .playback
+        let sessionCategory: String = config["iosCategory"] as? String ?? "playback"
+        self.sessionCategory = SessionCategory(rawValue: sessionCategory)!.mapConfigToAVAudioSessionCategory()
```
dcvz: This has the bug that if you give it a `iosCategory` but it's not a valid one, it will crash because of the force unwrap (!).

It'd be safer to write this as:
```swift
if let sessionCategory = config["iosCategory"] as? String {
  let mappedCategory = SessionCategory(rawValue: sessionCategory)
  self.sessionCategory = (mappedCategory ?? .playback).mapConfigToAVAudioSessionCategory()
}
```

ios/RNTrackPlayer/RNTrackPlayer.swift
```diff
@@ -130,12 +133,23 @@ public class RNTrackPlayer: RCTEventEmitter, AudioPlayerDelegate {
...
+        let sessionCategoryOpts: String? = config["iosCategoryOptions"] as? String ?? nil
+        //this is because this doesnt default as anything
+        if(sessionCategoryOpts != nil) {
+            self.sessionCategoryOptions = SessionCategoryOptions(rawValue: sessionCategoryOpts!)!.mapConfigToAVAudioSessionCategoryOptions()
+        } else {
+            self.sessionCategoryOptions = nil
         }
+
+        //configure mode -- defaults to .default
+        let sessionCategoryMode: String = config["iosCategoryMode"] as? String ?? "default"
```
dcvz: I'd use a similar approach as the one I've mentioned for categoryOptions.

ios/RNTrackPlayer/RNTrackPlayer.swift
```diff
@@ -346,6 +360,14 @@ public class RNTrackPlayer: RCTEventEmitter, AudioPlayerDelegate {
     @objc(play:rejecter:)
     public func play(resolve: RCTPromiseResolveBlock, reject: RCTPromiseRejectBlock) {
         print("Starting/Resuming playback")
+        //do this here so we can have bg audio until we play
+        try? AVAudioSession.sharedInstance().setActive(false)
+        if(self.sessionCategoryOptions != nil) {
+            try? AVAudioSession.sharedInstance().setCategory(self.sessionCategory, mode: self.sessionCategoryMode, options: self.sessionCategoryOptions!)
```
dcvz: Once the default is set to `[]` you should only have to call:
```swift
try? AVAudioSession.sharedInstance().setActive(false)
try? AVAudioSession.sharedInstance().setCategory(self.sessionCategory, mode: self.sessionCategoryMode, options: self.sessionCategoryOptions)
try? AVAudioSession.sharedInstance().setActive(true)
```

ios/RNTrackPlayer/Models/SessionCategories.swift
```diff
@@ -0,0 +1,89 @@
...
+        case .videoChat:
+            return .videoChat
+        case .videoRecording:
+            return .videoRecording
+        case .voiceChat:
+            return .voiceChat
+        case .voicePrompt:
+            if #available(iOS 12.0, *) {
+                return .voicePrompt
+            } else {
```
dcvz: Is this else needed then? 😉 Let's just omit it and treat it as invalid options for `< iOS 12` the logic in the setup method can handle this just fine.

ios/RNTrackPlayer/RNTrackPlayer.swift
```diff
@@ -21,6 +21,9 @@ public class RNTrackPlayer: RCTEventEmitter, AudioPlayerDelegate {
         return player
     }()
     
+    private var sessionCategory: AVAudioSession.Category = .playback
+    private var sessionCategoryOptions: AVAudioSession.CategoryOptions? = nil
```
dcvz: Let's change this to `[]` as per my other comment.

ios/RNTrackPlayer/RNTrackPlayer.swift
```diff
@@ -130,12 +133,23 @@ public class RNTrackPlayer: RCTEventEmitter, AudioPlayerDelegate {
...
-            reject("setup_audio_session_failed", "Failed to setup audio session", error)
+
+        //configure base category -- defaults to .playback
+        let sessionCategory: String = config["iosCategory"] as? String ?? "playback"
+        self.sessionCategory = SessionCategory(rawValue: sessionCategory)!.mapConfigToAVAudioSessionCategory()
+        
+        //configure extra opts -- defaults to nil
+        let sessionCategoryOpts: String? = config["iosCategoryOptions"] as? String ?? nil
+        //this is because this doesnt default as anything
+        if(sessionCategoryOpts != nil) {
```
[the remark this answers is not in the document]
dcvz: Also, it actually does default to something.. it defaults to `[]`. Let's use this instead.

Screenshot 2019-03-14 at 01 17 07

ios/RNTrackPlayer/Models/SessionCategories.swift
thessler27: my XCode was throwing an error here because it's expecting a return. that's why I was defaulting to `.default`. Is there a way to catch for it differently?

thessler27:
Not a problem at all! Happy to help. I've added most of the fixes you asked for, except for removing the `else` on the iOS 12 issue. My XCode was throwing an error, I wrote a response on your comment. Let me know. Thanks

ios/RNTrackPlayer/Models/SessionCategories.swift
dcvz: ah, interesting.. there probably is, but I can't think of it at the moment, so let's leave it as-is 👍

ios/RNTrackPlayer/RNTrackPlayer.swift
```diff
@@ -21,6 +21,9 @@ public class RNTrackPlayer: RCTEventEmitter, AudioPlayerDelegate {
         return player
     }()
     
+    private var sessionCategory: AVAudioSession.Category = .playback
+    private var sessionCategoryOptions: AVAudioSession.CategoryOptions? = []
```
dcvz: Why does this need to be an optional? As far as I can tell, there's no need for it to be.

thessler27: I had taken out the optional but I wasn't sure if it would affect the different type of `if let` we had written. just realized that we're initializing it so it'll likely never be `nil` haha. my bad. i'll change it.

thessler27:
@dcvz let me know if those changes work

dcvz:
maybe you'd also be kind enough to reflect these changes in the docs @thessler27 ?

thessler27:
@dcvz All set! #498 is up

timmywil:
I should have looked closer at this sooner, but category options is an OptionSet, which means it can set multiple at once, and I think I need to set 5 of them. I think category options needs to be an array of strings. For instance, for my app, I'm going to want to set `[.duckOthers, .allowBluetooth, .allowBluetoothA2DP, .allowAirPlay, .defaultToSpeaker]`.

rreiser-auti:
Sorry to bother you guys, but after i changed the version to dev, my app stopped to build in the ios. Then i had tried to investigated what was happening and i´ve got this error 
Captura de Tela 2019-03-18 às 17 51 59
But my swift knowledge is 0, so i just revert this function to the old version.
```
    do {
            try AVAudioSession.sharedInstance().setCategory(.playback, mode: .default)
            try AVAudioSession.sharedInstance().setActive(true)
        } catch {
            reject("setup_audio_session_failed", "Failed to setup audio session", error)
```

Any clue of why this is happening? i dont want to hardcode the package. Im using RN 0.59 by the way.

timmywil:
@rreiser-auti I noticed that too, but cleaning and rebuilding fixed it for me.

timmywil:
@rreiser-auti @thessler27 Sorry, I hadn't tried without pods. The model file wasn't added to the xcode project file. I've added a fix for this in my PR to fix category options (#500).

rreiser-auti:
Ty Timmy

NedyUdombat:
Hi I am having a weird issue that I have not been able to fix.

The audio is playing in the background of the IOS simulator on my laptop, but on an actual IOS device it stops playing in the background after about a minute and stop the app too.

This is the code from my Audio Player component

```
import React from 'react';
import { View, Image, TouchableOpacity } from 'react-native';
import TrackPlayer from 'react-native-track-player';

import ProgressBar from './ProgressBar';

class AudioPlayer extends React.Component {
  constructor() {
    super();
    this.state = {
      isPlaying: false,
      audioLoaded: false,
    };
  }

  componentDidMount = async () => {
    await TrackPlayer.setupPlayer({
      iosCategory: 'playback',
      iosCategoryMode: 'default',
      iosCategoryOptions: [
        'allowBluetooth',
        'allowBluetoothA2DP',
        'duckOthers',
        'allowAirPlay',
        'defaultToSpeaker',
      ],
    });
      stopWithApp: false,
      capabilities: [
        TrackPlayer.CAPABILITY_PLAY,
        TrackPlayer.CAPABILITY_PAUSE,
        TrackPlayer.CAPABILITY_SEEK_TO,
        TrackPlayer.CAPABILITY_JUMP_FORWARD,
        TrackPlayer.CAPABILITY_JUMP_BACKWARD,
      ],
      compactCapabilities: [
        TrackPlayer.CAPABILITY_PLAY,
        TrackPlayer.CAPABILITY_PAUSE,
        TrackPlayer.CAPABILITY_SEEK_TO,
        TrackPlayer.CAPABILITY_JUMP_FORWARD,
        TrackPlayer.CAPABILITY_JUMP_BACKWARD,
      ],
    });
    await TrackPlayer.add([
      {
        id: this.props.audioId,
        url: this.props.audioUri,
        title: this.props.audioTitle,
        artist: '',
        pitchAlgorithm: TrackPlayer.PITCH_ALGORITHM_MUSIC,
      },
    ]);
    await this.setState(previousState => ({
      ...previousState,
      audioLoaded: true,
    }));
    this.timeout = setTimeout(() => {
      if (this.props.autoPlay) {
        this.handlePlayPause();
      }
    }, 300);
  };

  componentWillUnmount = async () => {
    await TrackPlayer.stop();
  };

  componentDidUpdate = async () => {
    if (this.props.stopAudioPlayback && this.state.audioLoaded) {
      await TrackPlayer.stop();
    }
  };

  handlePlayPause = async () => {
    if (this.state.isPlaying) {
      await TrackPlayer.pause();
      this.setState(previousState => ({
        ...previousState,
        isPlaying: false,
      }));
    } else {
      await TrackPlayer.play();
      this.setState(previousState => ({
        ...previousState,
        isPlaying: true,
      }));
    }
  };

  setPlayingStatus = async playStatus => {
    if (playStatus === 'play') {
      this.setState(previousState => ({
        ...previousState,
        isPlaying: true,
      }));
    } else {
      this.setState(previousState => ({
        ...previousState,
        isPlaying: false,
      }));
    }
  };

  handleSeekTo = async seconds => {
    TrackPlayer.getPosition().then(position => {
      TrackPlayer.seekTo(seconds + position);
    });
  };

  render() {
    const { isPlaying } = this.state;
    return (
      <View
        style={{
          height: 90,
          backgroundColor: '#303030',
          borderRadius: 4,
          paddingVertical: 12,
          paddingHorizontal: 18,
          shadowColor: 'rgba(0, 0, 0, 0.25)',
          shadowOffset: { width: 3, height: 3 },
          shadowOpacity: 1,
          shadowRadius: 2,
          elevation: 3,
          position: 'relative',
        }}
      >
        <ProgressBar
          audioDuration={this.props.audioDuration}
          setPlayingStatus={this.setPlayingStatus}
        />
        <View
          style={{
            display: 'flex',
            flexDirection: 'row',
            justifyContent: 'center',
            alignItems: 'center',
            height: 35,
            marginBottom: 12,
          }}
        >
          <TouchableOpacity
            testID="jumpPrev15Seconds"
            style={{
              marginHorizontal: 22,
            }}
            onPress={() => this.handleSeekTo(-15)}
          >
            <Image
              source={require('../../assets/images/icons/rewind.png')}
              style={{
                height: 18,
                width: 18,
              }}
            />
          </TouchableOpacity>

          {isPlaying ? (
            <TouchableOpacity
              style={{
                marginHorizontal: 22,
              }}
              onPress={() => this.handlePlayPause()}
              testID="pauseMeditation"
            >
              <Image
                source={require('../../assets/images/icons/pause.png')}
                style={{
                  height: 18,
                  width: 13,
                }}
              />
            </TouchableOpacity>
          ) : (
            <TouchableOpacity
              style={{
                marginHorizontal: 22,
              }}
              onPress={() => this.handlePlayPause()}
              testID="playMeditation"
            >
              <Image
                source={require('../../assets/images/icons/play_white.png')}
                style={{
                  height: 18,
                  width: 13,
                }}
              />
            </TouchableOpacity>
          )}
          <TouchableOpacity
            testID="jumpNext15Seconds"
            style={{
              marginHorizontal: 22,
            }}
            onPress={() => this.handleSeekTo(15)}
          >
            <Image
              source={require('../../assets/images/icons/skip.png')}
              style={{
                height: 18,
                width: 18,
              }}
            />
          </TouchableOpacity>
        </View>
      </View>
    );
  }
}

export default AudioPlayer;

```

Below is a GIF of this happening on an actual IOS device.

[ScreenGrabbed Video](https://res.cloudinary.com/enle/video/upload/v1601480122/RPReplay_Final1601475355_vxykfy.mp4)

Please I need some assistance to fix this issue as it is long overdue, I feel like I am missing something. If anyone could just point me in the right direction. I will really appreciate



cc @thessler27 @timmywil @dcvz

Issue #478, opened 2019-03-08
[iOS] Stop or Reset player when going to background?


curiousdustin:
Most of the time I want my app to play music in the background.

But, there is a specific track that I want to only play while in the foreground on a certain screen.

I am stopping (or resetting) the player when `AppState` reports that the app is going to the background.

This works fine, except that the lock screen player controls persist. Not only are they there, but the have incorrect duration, and the play button doesn't work. (even though I don't actually want it to be there at all)

What steps do I need to take to make sure the iOS lock screen controls do NOT show up?

curiousdustin:
`AppState.addEventListener('change', this.onAppStateChange);`

Guichaguri:
I'm pretty sure you can `destroy()` it, but you'll have to implement logic to set it up again.

curiousdustin:
I still need to do more testing to narrow this down further but, here is some more info. I intend to do some testing in the example app, but this info is from my app. 

The trouble I’m having is specifically when leaving the app via device sleep button. 

```
Playing audio 
Device home button
Stop audio (triggered by AppState change)
Device sleep button
Device sleep button (wake to lock screen)
```
This scenario works as expected. Audio is stopped. Lock screen shows no controls. 

```
Playing audio
Device sleep button 
Audio stops (triggered by AppState change)
Device sleep button (wake to lock screen)
```
This scenario behaves differently. The audio still stops as intended. However, now the lock screen shows the controls when it should not because nothing is playing. 

I did try `destroy()`. Interestingly, it didn’t even stop the audio. I tried stop AND destroy, and it behaves just like stop. 

I didn’t dig as deep as I would like to yet. But, I couldn’t easily find a place in the native code that toggles the lock screen controls on/off. Perhaps this is mostly controlling automatically by iOS, and we just need to do something specific to “trick” it into hiding the controls.

Guichaguri:
Does #485 fixed it for you?

curiousdustin:
Yes, this is definitely an improvement.

However, this only fixes the case of use `reset()`. This works for my specific case, but I feel it could still be improved.

If I do `stop()` when the app goes to the background via the device sleep button, The audio stops, but the info center lock screen still shows the info, but the buttons don't work at all.

Perhaps a similar fix should be applied to the `stop()` method? Along with `beginReceivingRemoteControlEvents` whenever a track is started?

roberto-buratti:
[...]

PR #485, opened 2019-03-10
[iOS] Clear nowplayinginfocenter on `reset()`

thessler27:
When you call `reset()` the track does not get removed from the `MPNowPlayingInfoCenter` delegate. If we call the `UIApplication.shared`'s method `beginReceivingRemoteControlEvents` we can instantiate it when someone adds a track. This is delayed by 500ms so that the UI thread is not blocked when attempting to do the rest of the functionality of adding tracks. By setting it to being able to receive them, we can then execute `endReceivingRemoteControlEvents` on `reset()` and clear the `MPNowPlayingInfoCenter` properly

### DIFF
--- a/ios/RNTrackPlayer/RNTrackPlayer.swift
+++ b/ios/RNTrackPlayer/RNTrackPlayer.swift
@@ -350,8 +350,8 @@ public class RNTrackPlayer: RCTEventEmitter, AudioPlayerDelegate {
         resolve(NSNull())
         //tell app to stop receiving remote control events
         //so the MPNowPlayingInfoCenter clears properly
-        DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) {
-            UIApplication.shared.beginReceivingRemoteControlEvents();
+        DispatchQueue.main.async {
+            UIApplication.shared.endReceivingRemoteControlEvents();
         }
     }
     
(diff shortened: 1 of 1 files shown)


curiousdustin:
Would this address #478 ?

thessler27:
@curiousdustin yes it does fix that

ios/RNTrackPlayer/RNTrackPlayer.swift
```diff
@@ -224,6 +224,13 @@ public class RNTrackPlayer: RCTEventEmitter, AudioPlayerDelegate {
     
     @objc(add:before:resolver:rejecter:)
     public func add(trackDicts: [[String: Any]], before trackId: String?, resolve: RCTPromiseResolveBlock, reject: RCTPromiseRejectBlock) {
+        
+        //tell app to receive remote control events on add
+        //so we can remove it on remove
+        DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) {
```
dcvz: What would be the downside of doing this on `play`?

thessler27: This is just giving the app the ability to send events to MPNowPlayingInfoCenter. When you add tracks, you are basically _intending_ on playing, imo. I didn't have this on play,  because this functionality isn't necessarily doing anything except for giving the app the ability to update the `MPNowPlayingInfoCenter`. Play will handle the updates to the info center. On `reset`, it'll clear everything, and take away access so the InfoCenter clears. Then, you'll need to add more tracks anyway, so it'll re-initialize and prepare for utilizing it again

thessler27: It also requires the main thread, so I'm trying to limit the amount of calls as little as necessary

thessler27:
@dcvz i left a comment, let me know thanks

PR #487, opened 2019-03-11
[iOS] set automaticallyWaitsToMinimizeStalling to true for more resilient buffering

timmywil:
We've been having issues with buffering where it would often start playing before it had enough to play. Simply setting this to `true` fixes all our problems. I *think* it's safe to set this to `true` whether it's a local file or not.


holospeed:
Set this value to true solved a tone of problems to me! Really thank you for your comment!

dcvz:
We actually had this change because many users were reporting that they wanted immediate playback and it was taking too long to start.

I think the way forward will be to make this configurable via `setupPlayer`

timmywil:
I think https://github.com/react-native-kit/react-native-track-player/pull/490 will cover this.

PR #489, opened 2019-03-12
fixed NullPointerException when empty url is given

AbelTesfaye:
Whole app crashes on android when the javascript's track.url is ""(empty string), null or undefined. A simple if statement can fix that.

### DIFF
--- a/android/src/main/java/com/guichaguri/trackplayer/service/models/Track.java
+++ b/android/src/main/java/com/guichaguri/trackplayer/service/models/Track.java
@@ -113,7 +113,7 @@ public class Track {
         builder.putString(METADATA_KEY_ALBUM, album);
         builder.putString(METADATA_KEY_DATE, date);
         builder.putString(METADATA_KEY_GENRE, genre);
-        builder.putString(METADATA_KEY_MEDIA_URI, uri.toString());
+        builder.putString(METADATA_KEY_MEDIA_URI, (uri != null)?uri.toString():"");
         builder.putString(METADATA_KEY_MEDIA_ID, id);
 
         builder.putLong(METADATA_KEY_DURATION, duration);


android/src/main/java/com/guichaguri/trackplayer/service/models/Track.java
```diff
@@ -113,7 +113,7 @@ public Track(Context context, Bundle bundle, int ratingType) {
         builder.putString(METADATA_KEY_ALBUM, album);
         builder.putString(METADATA_KEY_DATE, date);
         builder.putString(METADATA_KEY_GENRE, genre);
-        builder.putString(METADATA_KEY_MEDIA_URI, uri.toString());
+        builder.putString(METADATA_KEY_MEDIA_URI, (uri != null)?uri.toString():"");
```
Guichaguri: Why is the track url empty?

AbelTesfaye: The track url can be empty whenever there is a need to load the track queue first then get the url asynchronously.

A typical example would be when using a url that will expire if nothing connects to it within a few seconds after it is created. This type of url needs to establish a connection only when `playback-track-changed` event is fired.

Another example is when you want to populate the track queue's network streamable urls only when you need to.
Instead of making un-necessary API requests to get the url of tracks that have a low likelyhood of being listened to, an alternative for this would be to request only urls of tracks that are near the current track in track queue.

Both examples outline the fact that the track player should not wait for the track url to load before it can insert tracks to the queue.

Guichaguri: Wouldn't it be easier to manage the queue by yourself and then only add the track when you actually have something to play?

AbelTesfaye: Nope, that would require users to wait for all the urls to fetch before they can play pre-fetched urls. This means playback won't start until every single url is loaded.

Guichaguri: Sorry, but I still don't get what the problem is. Can you clarify a little bit more?

Here's what I mean by managing the queue yourself:
```js
let queue = [...]; // Your own queue
let currentTrack = 0; // The current track in your queue

async playNext() {
    currentTrack++;
    let track = queue[currentTrack];

    if (!track.url) {
        // Fetch the URL first
        track.url = await fetchUrl(track.id);
    }
    await TrackPlayer.add(track);
    TrackPlayer.play();
}

TrackPlayer.addEventListener('playback-queue-ended', () => playNext());
```

You can also build logic to add the track before the current one finishes, so the user doesn't wait for it to be retrieved.

AbelTesfaye: Here's my experience with custom queue management:

Imagine a senario where a user is playing a bunch of tracks using my app. To be able to play any track from my custom queue, i'll need to use the following code:

    let customQueue = [...];// Important part
    let currentTrack = 0; 
    
    async playAtIndex(index) {
        let track = customQueue[index];
        if(track){

            if (!track.url) {
                track.url = await fetchUrl(track.id);
            }
        TrackPlayer.reset();// Reset so it doesn't play tracks from the previous TrackPlayer queue
        await TrackPlayer.add(track);
        TrackPlayer.play();
        }
        else{//Track doesn't exist in queue
        }
    
    }

Now imagine that the user closes my app and re-opens it. customQueue variable will be emptied (didn't persist app restart). 

**How can I make customQueue persistent?** 
Using AsyncStorage or SQLite everytime customQueue is modified comes to mind. 
But what if there was another way with less overhead? What if there is a way to magically store track queues and get them at any convenient time with the least dependencies and with the best performance? I looked around but I couldn't find anything better than `TrackPlayer.getQueue()`. 

Using `TrackPlayer.getQueue()` I can be sure that I will be able to access the track queue even if the app restarts(since queue is stored in an ExoPlayer object in RAM, and this object won't be deleted until the JVM calls its `finalize()` method ).

If I used my own custom queue management method then I'll have to make sure the queue will persist app restart and that would require me to use additional libraries, not to mention the performance decline of getting files from storage(compared to RAM). This is why I think TrackPlayer's queue is more prefereable to a custom queue.


**What is preventing me from using the TrackPlayer queue?**
The one and only obstacle here is that TrackPlayer doesn't allow track.url to be empty. My app crashes whenever I try to add a track without a url.

**The Fix**
Make TrackPlayer support empty urls...Even ExoPlayer doesn't crash when using empty string urls so it won't cause problems with ExoPlayer.


**TL;DR** 
Custom queue management requires me to use a slower and bloated method to persist queue after app restart...its best to fix TrackPlayer and prevent it from causing my app to crash track.url is empty

Guichaguri: Still, you'd be creating unnecessary heap and processing for ExoPlayer to pre-load those tracks. This isn't a fix, it's a work-around for your problem, doesn't seem right to put this kind of code into the project.

Why do you specifically need them to be empty? Can't you use a "fake" URI such as `my-app://track-to-replace`?

AbelTesfaye: Its always better to create a new heap than create, read and update files for this type of task.

I'm echoing my previous comment here. I need queues to be empty so I can be able to persist the queue after app restart, and since files are not the best option here i'd rather persist the queue using ExoPlayer.

I can use a fake URI to replace with the real URI, but I personally prefer to add less overhead to my code.

PR #490, opened 2019-03-12
[iOS] Add config to wait for buffer

rtmalone:
Adds a config option to set `automaticallyWaitsToMinimizeStalling` as `true`

Provides a solution to #487

### DIFF
--- a/ios/RNTrackPlayer/RNTrackPlayer.swift
+++ b/ios/RNTrackPlayer/RNTrackPlayer.swift
@@ -134,7 +134,7 @@ public class RNTrackPlayer: RCTEventEmitter, AudioPlayerDelegate {
     public func setupPlayer(config: [String: Any], resolve: RCTPromiseResolveBlock, reject: RCTPromiseRejectBlock) {
         // configure if player waits to play
         let autoWait: Bool = config["waitForBuffer"] as? Bool ?? false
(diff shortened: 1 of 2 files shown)


ios/RNTrackPlayer/RNTrackPlayer.swift
```diff
@@ -131,6 +131,10 @@ public class RNTrackPlayer: RCTEventEmitter, AudioPlayerDelegate {
     @objc(setupPlayer:resolver:rejecter:)
     public func setupPlayer(config: [String: Any], resolve: RCTPromiseResolveBlock, reject: RCTPromiseRejectBlock) {
         do {
+            let autoWait: Bool = config["waitForBuffer"] as? Bool ?? false
+            if autoWait {
+                player.automaticallyWaitsToMinimizeStalling = true
+            }
```
timmywil: I think you could either drop `?? false` or remove the `if` statement and set `autoWait` every time.

rtmalone: iirc, swift needs the default value because the option being unwrapped could be nil. But yeah, if it is always evaluating to a boolean I could set the option here every time.

rtmalone:
updated PR and rebased with `dev`

PR #498, opened 2019-03-18
[docs] Add docs for setting category, options, and mode on play

thessler27:


timmywil:
See https://github.com/react-native-kit/react-native-track-player/pull/500

thessler27:
Saw #500 @timmywil, looks good! Updated my docs in support 👍

PR #499, opened 2019-03-18
Buffering state before song is ready to play

ithustle:
When a song is tapped to play in IOS, the state changes from stop to buffering and then changes to play. It doesn't happened on Android.


Guichaguri:
Thank you :)

PR #500, opened 2019-03-18
[iOS] Allow multiple category options

timmywil:
`categoryOptions` is an OptionSet, which means it can set multiple at once. I think category options needs to be an array of strings. For instance, for my app, I'm going to want to set [.duckOthers, .allowBluetooth, .allowBluetoothA2DP, .allowAirPlay, .defaultToSpeaker].

I'm still new to Swift. There might be a better way to convert an array of strings to an OptionSet, but I'm currently using this PR in my app like so...

```js
  TrackPlayer.setupPlayer({
    iosCategory: 'playAndRecord',
    iosCategoryMode: 'default',
    iosCategoryOptions: [
      'duckOthers',
      'allowBluetooth',
      'allowBluetoothA2DP',
      'allowAirPlay',
      'defaultToSpeaker'
    ]
  })
```

Edit: my first pass wasn't actually setting the options. I didn't need to reduce, I just needed to pass the mapped options to the constructor.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -70,8 +70,8 @@ declare namespace RNTrackPlayer {
     playBuffer?: number;
     maxCacheSize?: number;
     iosCategory?: 'playback' | 'playAndRecord' | 'multiRoute' | 'ambient' | 'soloAmbient' | 'record';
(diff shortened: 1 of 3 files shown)


timmywil:
@dcvz @thessler27 For your eyes

thessler27:
Looks good to me! 👍

timmywil:
Added a separate commit to fix the xcode build, related to @thessler27's PR preceding this one. It seems the model file wasn't yet added to the pbxproj.

PR #504, opened 2019-03-20
Improve docs

puckey:
Some edits to the documentation clearing up a few things that threw me off.

It is the start of what would become a broader overhaul of the docs.

The first commit:
- added 'id' here and there to be more specific about talking about the id of the track and not the track object
- renamed 'Track Structure' to 'Track Object' to be more clear about the fact that these are plain javascript objects
- renamed skip(track) to skip(id) to avoid confusion with the track object
- renamed setOptions 'data' parameter to 'options'
- added missing full stops
- cleaned up whitespace of some tables

### DIFF
--- a/docs/Documentation.md
+++ b/docs/Documentation.md
@@ -20,7 +20,7 @@ permalink: /documentation/
 * [Components](#components)
   * [ProgressComponent](#progresscomponent)
 * [Objects](#objects)
-  * [Track Structure](#track-structure)
+  * [Track Object](#track-object)
   * [Resource Object](#resource-object)
 
 ## Constants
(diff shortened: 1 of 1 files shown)


puckey:
For the API docs I have a few questions:

### API.md -> Starting off

> You also need to register a playback service right after you have registered the main component of your app

Is this required? Why does it need to be done after registering the main component?

In this and other examples `module.exports` is used. Since the `import  x from 'x'` syntax is being used, this could become `export default`

### API.md -> Player information

> After a media file has been loaded, you can get information about it (such as retrieving the state)

How can you tell if a track was loaded?

### API.md -> Receiving Events -> React components

> Works as long as the UI is mounted
> Should only be used to update the UI state

What is meant above? What works? Why should it only be used to update UI state?

### Playback Service

The first question I would have here is: what  is a playback service?

It is unclear to me what the specifics are of this.. Are only events inside a function passed to registerPlaybackService called when the app is in the background? Others are ignored?

> ```javascript
> TrackPlayer.registerPlaybackService(() => require('./service.js'));
> ```

The example above uses `require` inside a function – is this necessary? It leads me to believe it would not be able to read anything outside of its own scope – is this the case?

It 'works as long as the player runs' – what does 'works' mean here? What happens when it no longer works?

### Notes

> Due to the notification not being swipable, we highly recommend you to have a stop button in the notification. The button should destroy() the player.

Is this android only?

puckey:
Oh darn – i should have been more clear about wanting to keep this pull request open for longer until I am finished  : ) should I open another one?

Guichaguri:
Yes, open a new one, please.

#### API.md -> Starting off
Yes, this is required. I expect someone to be trying it out while they follow the tutorial, and if you try to setup the player without registering the service, Android crashes with `No task registered for key TrackPlayer`.

Feel free to update `module.exports` to `export default`, I don't think there's anything against that.

#### API.md -> Player information
The track usually loads dynamically in background and currently there's no way to know that, but when the track starts playing, it definitely loaded.

We should rephrase that, you can retrieve the state even without adding any track to the queue as long as the player is already set up.

#### API.md -> Receiving Events -> React components
You should only use events inside React components to update the UI, purely visual updates.
Logic related to the behavior of the player (e.g. reacting to `remote-` events) should be added to the service, mostly because the UI might unmount while the playback is in background, but the service stays up as long as the playback runs. I hope I was clear enough and that made sense.

#### Playback Service
The playback service runs as long as the player runs, which means you can bind any logic related to the player directly into the player. (e.g. analytics)
It will start running when the player is set up, but it will only stop after the player is destroyed, which means it will keep running when the player also goes into background.
As said above, there are a few scenarios where you should use the playback service to handle logic instead of adding event handlers to React components.

Yes, the `require` is necessary. You can see it is wrapped inside a function, which means it will only `require` (and load the service code into memory) when it actually needs to load. It follows the same lifecycle as [Headless JS](https://facebook.github.io/react-native/docs/headless-js-android).

Basically, when the service is destroyed, the JS engine is usually also stopped (there are still a few cases where RN keeps it running, I haven't figured out them yet), which means that any code running will stop and terminate immediately (just like killing a process)

#### Notes
Yes, mostly it is. This is to solve a problem where Android 8+ doesn't allow non-foreground services to be running while the app is in background. When the app is in background and the user stops, we recommend you to destroy the service altogether to prevent any crashes.
I'm not really sure, but in iOS, this also seems like a good practice due to how media buttons work.

PR #522, opened 2019-04-03
fix issue with url + fix bug

PierreBresson:
Fix crash if url contains accents.
Improve url encoding.


rkbhochalya:
This one actually breaks firebase audio file urls. Please check issue #617 and #543 for more info.

PR #530, opened 2019-04-11
[iOS] Send remote events when interruption handling

anderslemke:
To allow the application doing whatever needs to be done when pausing.

### DIFF
--- a/ios/RNTrackPlayer/RNTrackPlayer.swift
+++ b/ios/RNTrackPlayer/RNTrackPlayer.swift
@@ -78,6 +78,7 @@ public class RNTrackPlayer: RCTEventEmitter {
             "remote-stop",
             "remote-pause",
             "remote-play",
+            "remote-duck",
             "remote-next",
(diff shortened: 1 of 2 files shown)


Guichaguri:
Correct if I'm wrong, but shouldn't it trigger the [`remote-duck`](https://react-native-kit.github.io/react-native-track-player/documentation/#remote-duck) event?

anderslemke:
I think you're right, @Guichaguri .

I'm using `remote-duck` now.

anderslemke:
This branch is running in production on our app now.

We might need to add some documentation, as your app needs to keep track of it's state when `remote-duck` is sent, to know if it should start playing or not when the ducking stops.

iOS will send event with `. shouldResume` regardless of if your app was playing or not when the interruption started.

vdesdoigts:
Any update for this? I need it :)

dcvz:
This looks great @anderslemke ! thank you for this 🙏

PR #531, opened 2019-04-11
[iOS] Add toggle play pause

anderslemke:
This solves a few issues with lined EarPods and AirPods.

Especially that both EarPods and AirPods now can toggle play/pause. (Which they couldn't before)


ios/RNTrackPlayer/RNTrackPlayer.swift
```diff
@@ -191,6 +193,11 @@ public class RNTrackPlayer: RCTEventEmitter, AudioPlayerDelegate {
             self?.sendEvent(withName: "remote-play", body: nil)
             return MPRemoteCommandHandlerStatus.success
         }
+
+        player.remoteCommandController.handleTogglePlayPauseCommand = { [weak self] _ in
```
dcvz: @anderslemke search for: `player.remoteCommandController.handleTogglePlayPauseCommand` -> this method already exists.. does it not work as expected?

android/src/main/java/com/guichaguri/trackplayer/module/MusicModule.java
```diff
@@ -113,6 +113,7 @@ private void waitForConnection(Runnable r) {
 
         // Capabilities
         constants.put("CAPABILITY_PLAY", PlaybackStateCompat.ACTION_PLAY);
+        constants.put("CAPABILITY_TOGGLE_PLAY_PAUSE", PlaybackStateCompat.ACTION_PLAY);
```
Guichaguri: There is also an [ACTION_PLAY_PAUSE](https://developer.android.com/reference/android/support/v4/media/session/PlaybackStateCompat.html#ACTION_PLAY_PAUSE) in Android

ios/RNTrackPlayer/RNTrackPlayer.swift
appinteractive: Yeah it seems that it exists but wasn't used until now, corret?

appinteractive:
Any progress @anderslemke?

Guichaguri:
I think we should automatically set this capability based on whether `CAPABILITY_PLAY` and `CAPABILITY_PAUSE` are both set, we can also trigger `remote-play` or `remote-pause` based on the player state instead of adding a new event.

ios/RNTrackPlayer/RNTrackPlayer.swift
anderslemke: Yes. That already exists. I'll remove this.

I don't think it was used before though, as @appinteractive points out.

android/src/main/java/com/guichaguri/trackplayer/module/MusicModule.java
anderslemke: OK. I'll update.

anderslemke:
We are using this in production at Zetland now.

To answer your questions @Guichaguri :
1) I can't really tell if automagically adding the toggle capability based on play/pause is a good idea. I can't think if a situation where you wouldn't want that, but on the other hand, I don't understand why Apple would make it a separate capability in the first place.

2) The code that emitted a new event in now removed, so we use `remote-pause` and `remote-play` now.

appinteractive:
Thanks @anderslemke for your work! 👌
@Guichaguri mergeable?

Guichaguri:
@anderslemke I'm still not sure if we should expose those kind of API decisions to our API, but I'll leave that to @dcvz to decide.

dcvz:
what are we doing on Android @Guichaguri ? My intuition would be to just enable `playpause` when `play` and `pause` are enabled. That's what I tried to do the first time around.

Guichaguri:
Android does have a PLAY_PAUSE capability, but it doesn't have anything different over enabling PLAY and PAUSE separately. I think we should go with automatically enabling PLAY_PAUSE when PLAY and PAUSE are both enabled, and then splitting the event into `remote-play` and `remote-pause` depending on the state.

My guess is that the device uses the PLAY_PAUSE when it doesn't know (or is not sure about) the state of the player, like a smartwatch that doesn't have a stable connection to the phone due to battery saving reasons. And that's probably the same thing that happens with AirPods/EarPods, they may not know the current player state, so PLAY_PAUSE is triggered instead.

mitchdowney:
This feature would be greatly appreciated :) the ability to toggle play/pause with the headphone control is very convenient.

t-moster:
> This feature would be greatly appreciated :) the ability to toggle play/pause with the headphone control is very convenient.

Same here, we encountered the exact same problem and would love to have the feature.

rcerrejon:
We need that feature too, would be certainly great if someone make a the review and merge the changes. I wouldn't  like to Fork this module just for that feature.

curiousdustin:
This would be a great addition :) 

@Guichaguri , If I understand correctly the only thing left here is to add the logic to automatically enable the CAPABILITY_TOGGLE_PLAY_PAUSE  capability if both CAPABILITY_PLAY and CAPABILITY_PAUSE are provided?

Guichaguri:
@curiousdustin Yes, remove the `CAPABILITY_TOGGLE_PLAY_PAUSE`, automatically set it when both `CAPABILITY_PLAY` and `CAPABILITY_PAUSE` are set, add logic to trigger the `remote-play` or `remote-pause` events when the play-pause button is pressed. No need for breaking changes or new API

puckey:
I made a PR here that automatically deduces togglePlayPause from play and pause capabilities: https://github.com/react-native-kit/react-native-track-player/pull/957

puckey:
Merged https://github.com/react-native-kit/react-native-track-player/pull/957

biancamagalhaes:
@anderslemke i had the same problem with ios and you helped me a lot. Thanks!

PR #549, opened 2019-04-18
Added warning that Expo is not supported.

dglinyanov:
I've just started my experience with React Native and was really confused - Quickstart page for RN suggests using Expo as recommended way to create new projects:

https://facebook.github.io/react-native/docs/getting-started

I've tried both managed and ejected Expo project and bare project, though neither of them seems to be compatible with `react-native-track-player`. Finally with plain `react-native init` we made it, but till that moment it was very frustrating.

Considering that Expo is suggested in official docs and this package currently has no support for it I propose to declare it in Installation page of docs so people could save time searching project issues on random build exceptions on both platforms.

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -3,9 +3,6 @@ title: Installation
 permalink: /install/
 ---
 
-## Warning
-**Currently react-native-track-player does not support projects with Expo or Expokit.**
-
 ## Installing the packages
 **1. Install the module from npm or yarn**
 ```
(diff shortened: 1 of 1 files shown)


docs/Installation.md
```diff
@@ -3,6 +3,9 @@ title: Installation
 permalink: /install/
 ---
 
+## Warning
+**Currently react-native-track-player does not support projects with Expo or Expokit.**
+
```
Guichaguri: I think that shouldn't be a warning, but a note. Perhaps it should belong to the troubleshooting section instead.

Guichaguri:
Thank you 💃

PR #550, opened 2019-04-19
[ios] Handle interruption & support remote-duck

minhtc:


Guichaguri:
Why not use the same parameters sent by Android? It seems like it can be mapped in the following way:

| iOS State | TrackPlayer parameter |
| ---------- | ------------------------ |
| `began` | `{paused: true, permanent: false, ducking: false}` |
| `shouldResume` | `{paused: false, permanent: false, ducking: false}` |
| `shouldNotResume` | `{paused: false, permanent: true, ducking: false}` |

An implementation like this one would work on both platforms:
```js
(event) => {
    if (event.permanent) {
        TrackPlayer.stop();
    } else if (event.paused) {
        TrackPlayer.pause();
    } else if (event.ducking) {
        TrackPlayer.setVolume(0.5);
    } else {
        TrackPlayer.setVolume(1);
        TrackPlayer.play();
    }
}
```

minhtc:
> Why not use the same parameters sent by Android? It seems like it can be mapped in the following way:
> 
> iOS State	TrackPlayer parameter
> `began`	`{paused: true, permanent: false, ducking: false}`
> `shouldResume`	`{paused: false, permanent: false, ducking: false}`
> `shouldNotResume`	`{paused: false, permanent: true, ducking: false}`
> An implementation like this one would work on both platforms:
> 
> ```js
> (event) => {
>     if (event.permanent) {
>         TrackPlayer.stop();
>     } else if (event.paused) {
>         TrackPlayer.pause();
>     } else if (event.ducking) {
>         TrackPlayer.setVolume(0.5);
>     } else {
>         TrackPlayer.setVolume(1);
>         TrackPlayer.play();
>     }
> }
> ```

I was think about mapping same parameters, but I decided to keep separate from android, because it's very different handler on each platform. In iOS we can't play, pause, or setVolume during call. Bellow is my js code to handle audio interruptions:

```js
      if (Platform.OS === 'ios' && data === 'shouldResume') {
        Player.play();
      } else if (Platform.OS === 'android') {
        if (data.paused) {
          const currentState = await TrackPlayer.getState();
          wasPlaying = currentState === TrackPlayer.STATE_PLAYING;
          Player.pause();
        } else if (data.ducking) {
          Player.pause();
        } else if (wasPlaying) {
          Player.play();
          wasPlaying = false;
        }
      }
```

anderslemke:
Duplicate of #530?

dcvz:
This has been merged as part of #530

PR #563, opened 2019-04-23
[iOS] updateMetadataForTrack fix

holospeed:
The fix change "withProperties" method name to "properties" method name, which is the right instance.


dcvz:
Thanks for this fix @holospeed <3

PR #571, opened 2019-04-26
[ios] Make setupPlayer idempotent

anderslemke:


anderslemke:
OK. This doesn't seem to work.

So your best option is probably to only call `setupPlayer` once from your JS.

Either way, if you call more than once, you get multiple event listeners.

dcvz:
Why not just add a flag that makes sure it isn't run again?

curiousdustin:
@anderslemke , sounds like this should be closed?

puckey:
@anderslemke I am going to close this pull request, since you stated it does not work. If you want to have it reopened, please let us know.

PR #573, opened 2019-04-29
feat: React hook 🎣 for TrackPlayer events

matthamil:
## Description

React v16.8 introduced [hooks](https://reactjs.org/docs/hooks-intro.html), which are now available in React Native as of [v0.59](https://facebook.github.io/react-native/blog/2019/03/12/releasing-react-native-059).

This PR makes two changes: 

#### 1. Makes all event types available via a named export

Saves someone the time from declaring their own constants

```jsx
import { TrackPlayerEventTypes } from 'react-native-track-player';

console.log(TrackPlayerEventTypes.REMOTE_PLAY); // 'remote-play'
```

#### 2. Exports a hook that allows users to register an event listener from a component

Example with React hook:

```jsx
import React from 'react';
import { Text, View } from 'react-native';
import { useTrackPlayerEvent, TrackPlayerEvents } from 'react-native-track-player';

const MyComponent = () => {
  const [playerState, setState] = useState(null)

  useTrackPlayerEvent(TrackPlayerEvents.PLAYBACK_STATE, (playbackState) => {
    setState(playbackState)
  });

  const isPlaying = playerState === 'STATE_PLAYING';

  return (
    <View>
      <Text>The TrackPlayer is {isPlaying ? 'playing' : 'not playing'}</Text>
    </View>
  ); 
};
```

The hook isn't anything crazy, but I figured that PRing it for the library might save a few people a couple of minutes so they don't have to write their own hook :)

If this seems 👌 , I can also go back and update the TS definitions to include these.


matthamil:
I noticed there aren't any tests for any of the JS code (there's not much going on in the JS side of things anyway), but I'd be happy to include test coverage for the hook if you have a preferred test lib like Jest.

Guichaguri:
Pretty awesome! I'm sold on the idea, but I think the implementation can be improved a bit. @dcvz what do you think?

lib/hooks.js
```diff
@@ -0,0 +1,28 @@
...
+
+function isValidTrackPlayerEvent(evt) {
+    return TrackPlayerEvents[evt] !== undefined;
+}
+
+/**
+ * @description
+ *   Attaches handler to TrackPlayer event and performs cleanup on unmount
+ */
+module.exports.useTrackPlayerEvent = (event, handler) => {
```
dcvz: would it make sense for this to be able to take an array of event types? so that you could use this callback, to subscribe to multiple? instead of having a few callbacks?

Guichaguri: Also, why not move this function to `index.js`?

matthamil: I can make this accept an array of event types :+1:. Should we also log a dev-only warning when the library user has a typo or invalid event type? (i.e., `TrackPlayerEventTypes[event] !== undefined`). 

I opted to move this into its own file because it seemed like `index.js` was already sizable, but I can move it into `index.js` if you want. 

I am also not sure if users under RN v0.59 will see errors if this is part of `index.js` when they do `import * as TrackPlayer from 'react-native-track-player'` in their code since `useEffect` will not exist in React for them (yet). It might only error when they attempt to use the function.

I also noticed there's not really a JS build step in the repo. I'd be happy to set one up for the project as well (in a separate PR). This would fix any potential import problems that a library consumer might have if they aren't compiling the ES6+ plus the added benefit of a minified lib to be used (plus allows anyone to use esnext features in this repo).

matthamil:
Made changes to the PR. I also added a `useTrackPlayerProgress` hook in case other library users didn't want to inherit from a custom component to track progress. I haven't actually tested it yet, but figured I'd add that hook to this PR since it was another hook I was going to use in my project.

LMK what you think or if I need to make any further changes.

lib/hooks.js
```diff
@@ -0,0 +1,110 @@
...
+
+    async function getProgress() {
+        try {
+            const [position, bufferedPosition, duration] = await Promise.all([
+                TrackPlayer.getPosition(),
+                TrackPlayer.getBufferedPosition(),
+                TrackPlayer.getDuration()
+            ])
+
+            useDebugValue({ position, bufferedPosition, duration });
```
dcvz: Is this still necessary?

docs/Documentation.md
```diff
@@ -474,3 +480,73 @@ Only the `id`, `url`, `title` and `artist` properties are required for basic pla
...
+  useTrackPlayerEvents(events, (event) => {
+    if (event.type === TrackPlayerEvents.PLAYBACK_ERROR) {
+      console.warn('An error occured while playing the current track.');
+    }
+    if (event.type === TrackPlayerEvents.PLAYBACK_STATE) {
+      setState(playbackState)
+    }
+  });
+
+  const isPlaying = playerState === 'STATE_PLAYING';
```
Guichaguri: Please, use the `TrackPlayer.STATE_PLAYING` constant instead

lib/hooks.js
matthamil: It's not a necessary addition, but I thought it could be valuable. From what I understand about `useDebugValue`, it adds a friendly debug message to the React DevTools ([docs](https://reactjs.org/docs/hooks-reference.html#usedebugvalue). The docs say it is "most valuable for custom Hooks that are part of shared libraries." Thinking about this some more--Since this logs every second by default, its probably going to be a lot of noise in a user's React DevTools. I'll go back and remove it 👍

matthamil: I'll also update the TS definitions. I've been working on a TS project using `react-native-track-player`, so I am motivated to have types for these changes :).

docs/Documentation.md
matthamil: Good catch! I'll go back and change this.

lib/hooks.js
matthamil: tbh since there really isn't much JS in this project, I'd be happy to rewrite the current JS code into TS in another PR if you think that would be a welcome change.

Guichaguri:
Thank you!

matthamil:
thanks for the thorough review to both of you :) @Guichaguri @dcvz I've been a bit slow to add the TS types, but I will follow up with another PR either tonight or tomorrow to get those added as well.

lib/hooks.js
```diff
@@ -0,0 +1,108 @@
...
+ */
+module.exports.useTrackPlayerProgress = (interval = 1000) => {
+    const initialState = {
+        position: 0,
+        bufferedPosition: 0,
+        duration: 0 
+    };
+
+    const [state, setState] = useState(initialState);
+    const [interval, setInterval] = useState(interval);
```
xgustavoh: `interval` has been declared in:
```JS
module.exports.useTrackPlayerProgress = (interval = 1000) => {
```

matthamil: good catch! the dangers of contributions w/o tests I suppose :).

PR #579, opened 2019-05-06
[android] Fix: Bad notification for startForeground: java.lang.Runtim…

minhtc:
Fix bug:
Fatal Exception: android.app.RemoteServiceException
Bad notification for startForeground: java.lang.RuntimeException: invalid channel for service notification: null

The default  "Miscellaneous" notification channel (NotificationChannel.DEFAULT_CHANNEL_ID) doesn't exist on some devices

### DIFF
--- a/android/src/main/java/com/guichaguri/trackplayer/service/MusicService.java
+++ b/android/src/main/java/com/guichaguri/trackplayer/service/MusicService.java
@@ -1,8 +1,8 @@
 package com.guichaguri.trackplayer.service;
 
-import android.app.Activity;
-import android.app.Notification;
 import android.app.NotificationChannel;
+import android.app.NotificationManager;
+import android.content.Context;
 import android.content.Intent;
 import android.os.Build;
 import android.os.Bundle;
@@ -12,7 +12,6 @@ import android.support.v4.app.NotificationCompat;
 import android.support.v4.content.LocalBroadcastManager;
 import android.support.v4.media.session.MediaButtonReceiver;
 
-import android.support.v4.media.session.MediaSessionCompat;
 import com.facebook.react.HeadlessJsTaskService;
 import com.facebook.react.ReactInstanceManager;
 import com.facebook.react.bridge.Arguments;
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/guichaguri/trackplayer/service/MusicService.java
```diff
@@ -25,9 +24,24 @@
  */
 public class MusicService extends HeadlessJsTaskService {
 
+    private static final String MUSIC_SERVICE_NOTIFICATION_CHANNEL_ID = "MusicService";
```
Guichaguri: Is there a reason for creating a new channel? Can't you use `Utils.NOTIFICATION_CHANNEL` instead, unifying the registering code?

minhtc: Can we make sure that the `Utils.Notificaiton_CHANNEL` existed before init Music Service?

bemnet4u: NOTIFICATION_CHANNEL is static so should be created on class load.

Guichaguri: I know what you mean. You should probably move the registering code to a static method (probably in Utils), then call it whenever you need to use the channel.

minhtc: The  ExoPlayer Controls Notification is using Utils.Notificaiton_CHANNEL, I'm not if it's ok to re-create channel during Playback notification displaying?

Guichaguri: Yes, you can register the channel multiple times, the OS will just overwrite it when you register it again. In fact, Google recommends registering it whenever you want to display a notification, so you can be sure that the channel exists.

puckey: Do you mean something like the following to use wherever we need a notification channel?

`Utils.getNotificationChannel`:
```java
    public static String getNotificationChannel(Context context) {
        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
            NotificationChannel channel = new NotificationChannel(
                Utils.NOTIFICATION_CHANNEL,
                "MusicService",
                NotificationManager.IMPORTANCE_DEFAULT
            );
            channel.setShowBadge(false);
            channel.setSound(null, null);
            ((NotificationManager) context.getSystemService(Context.NOTIFICATION_SERVICE)).createNotificationChannel(channel);
        }
        return Utils.NOTIFICATION_CHANNEL;
    }
```

To use in:

https://github.com/react-native-kit/react-native-track-player/blob/6471be68dfc772bc3f94ad74d74455902651a7c8/android/src/main/java/com/guichaguri/trackplayer/service/MusicService.java#L81

https://github.com/react-native-kit/react-native-track-player/blob/6471be68dfc772bc3f94ad74d74455902651a7c8/android/src/main/java/com/guichaguri/trackplayer/service/metadata/MetadataManager.java#L65

mkuczera: Is there any update on this issue?

curiousdustin: What is needed to move forward here? Is @puckey's solution appropriate, and complete?

minhtc: > Yes, you can register the channel multiple times, the OS will just overwrite it when you register it again. In fact, Google recommends registering it whenever you want to display a notification, so you can be sure that the channel exists.

I created an other PR as your suggestion here https://github.com/react-native-kit/react-native-track-player/pull/769 
Please help to close this PR. Thank you!

Guichaguri:
Superseded by #723

PR #586, opened 2019-05-10
[iOS] Fix crash when setting rate

anderslemke:


dcvz:
This is already in place :) thanks @anderslemke

PR #595, opened 2019-05-17
add fix in example for unnecessary setupPlayer of the player

nisheetsun:
Prevented unnecessary setupPlayer of the player in PlaylistScreen screen.


dcvz:
the example has been redone in the main (v2) branch.. thanks for this @nisheetsun but I will close this now.

PR #602, opened 2019-05-21
[android] Fix LocalBroadcastManager import

puckey:


c1ngular:
i still can not compile example project with this fix , still says "MusicModule.java:9: error: package androidx.localbroadcastmanager.content does not exist", is there any additional steps involved ? thanks

puckey:
I would suggest using an earlier release if you want to use the example app. But if you want to see a work in progress version, I am busy reworking the example app here: https://github.com/react-native-kit/react-native-track-player/pull/600

(Note that I have yet to test the iOS version)

puckey:
> For the current androix version the module is using, the LocalBroadcastManager hasn't been refactored into the new package yet. I'll be updating it before publishing a new release.

https://github.com/react-native-kit/react-native-track-player/commit/0f411baf0c8b21184d4585c9515d805798b0a244

PR #603, opened 2019-05-21
[react] fix hooks

puckey:
I fixed a few things in the hooks while working on the example app:
- Add missing import
- Fix evt -> event variable name
- Fix warning of invalid types and avoid filtering in production, since subscribing to a non-existing event is not going to throw any errors. Add information about which event was invalid.
- Fix double const in useTrackPlayerProgress and just return the state, since we can control interval by what is passed into it
- only set progress interval when the player is either playing or buffering
- remove catching of player errors, since these functions should not be throwing errors (and if they do, we should fix the underlying cause)

@matthamil perhaps you would like to take a look?

### DIFF
--- a/lib/hooks.js
+++ b/lib/hooks.js
@@ -1,10 +1,6 @@
-import { useEffect, useState, useDebugValue } from 'react';
+import { useEffect, useState, useDebugValue, useRef } from 'react';
 import * as TrackPlayer from './index';
 import TrackPlayerEvents from './eventTypes';
-
-function isValidTrackPlayerEvent(evt) {
-    return TrackPlayerEvents[evt] !== undefined;
-}
 
 /**
  * @description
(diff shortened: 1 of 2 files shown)


lib/hooks.js
```diff
@@ -83,22 +109,19 @@ module.exports.useTrackPlayerProgress = (interval = 1000) => {
...
-                TrackPlayer.getBufferedPosition(),
-                TrackPlayer.getDuration()
-            ])
-
-            setState({ position, bufferedPosition, duration });
-        } catch (err) {
-            // The player is probably not ready yet.
-            // todo: better error handling
-        }
+    const getProgress = async () => {
```
matthamil: The try/catch block was taken from the progress component (https://github.com/react-native-kit/react-native-track-player/blob/0f411baf0c8b21184d4585c9515d805798b0a244/lib/ProgressComponent.js#L33-L46), I am not sure if we want to remove that error handling.

puckey: This will only ever be called when playback state is one of `TrackPlayer.STATE_PLAYING` ir `TrackPlayer.STATE_BUFFERING`, which can only occur when the player is initialised

matthamil:
@puckey hey thanks for fixing my bad code :)

I'll take a look at this tonight or tomorrow. Please do not hesitate to ping me again if I don't remember to do so!

lib/hooks.js
matthamil: Awesome, thanks for the clarification!

matthamil:
On an unrelated note, I see that you helped create [tonite.dance](https://tonite.dance). I remember seeing that when it made its way around the Twitterverse. Love it, super cool project 😄

puckey:
> On an unrelated note, I see that you helped create [tonite.dance](https://tonite.dance). I remember seeing that when it made its way around the Twitterverse. Love it, super cool project 😄

Thanks - it was fun to work on!

PR #615, opened 2019-06-03
Submodule for Vendored AudioPlayer & Updated example RN 0.59 with Hooks

dcvz:
Update the vendored audio player to be a submodule -- furthermore a fork of the original submodule to make it easier to be able to make our own modifications while keeping track of our difference with the original master.

Also updates our example app to not rely on mobx and use the latest RN (0.59.8) with support for hooks.

Also fixes some issues with hooks introduced in @puckey's branch :)


Guichaguri:
Since the fork is directly linked to this module, shouldn't it go into the react-native-kit org?

PR #619, opened 2019-06-07
[Android] Fix notification reinitialization after service is stopped

biomancer:
On Android when playback is stopped (not paused) and playback service is not in foreground mode, it's killed after some time if application is background. On my Huawei mate 10 pro with Android 9 that happens exactly after 1 minute after you minimize the app.
After that if app is reopened and you start playback, waitForConnection starts service once again, but **you get notification without any controls / action buttons**. That happens because when service is stopped, MetadataManager is destroyed, and new instance is created when we start playback later, and that means that options that were set via updateOptions, including all these notification capabilities, are lost. 

This fix keeps specified options in MusicModule that is not destroyed on service stop, and reapplies options when service is rerun.

### DIFF
--- a/android/src/main/java/com/guichaguri/trackplayer/module/MusicModule.java
+++ b/android/src/main/java/com/guichaguri/trackplayer/module/MusicModule.java
@@ -34,6 +34,7 @@ public class MusicModule extends ReactContextBaseJavaModule implements ServiceCo
     private MusicEvents eventHandler;
     private ArrayDeque<Runnable> initCallbacks = new ArrayDeque<>();
     private boolean connecting = false;
+    private Bundle options;
 
     public MusicModule(ReactApplicationContext reactContext) {
         super(reactContext);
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/guichaguri/trackplayer/module/MusicModule.java
```diff
@@ -70,6 +71,11 @@ public void onServiceConnected(ComponentName name, IBinder service) {
         binder = (MusicBinder)service;
         connecting = false;
 
+        // Reapply options that user set before with updateOptions
+        if (this.options != null) {
+          binder.updateOptions(this.options);
```
Guichaguri: Can you indent this line with 4 spaces instead of 2?

biomancer: Sure, fixed

Guichaguri:
Thank you!

Issue #543, opened 2019-04-18
[iOS] url decoding issue


kevex91:
Issue: When I load urls containing '%2F'. This gets converted to '/' when the track is loaded.

This is an example url: https://firebasestorage.googleapis.com${ourEndpoint}o/tracks%2F6bd9a8b7-4998-49c6-a86f-f747dd8fad9.mp3?alt=media&token=78a05743-862-4ddb-bce-d9797eb91bb

Firebase storage uses '%2F' 🤷‍♂️ to define a path into a folder. Seems related to #40 and #44 .

This is what the track player attempts to play:
https://firebasestorage.googleapis.com${ourEndpoint}o/tracks/6bd9a8b7-4998-49c6-a86f-f747dd8fad9.mp3?alt=media&token=78a05743-862-4ddb-bce-d9797eb91bb

This results in: load failed with error Error Domain=NSURLErrorDomain Code=-999 "cancelled"

AylanBoscarino:
I'm having the same issue. Is there a solution?

rkbhochalya:
I'm having this issue as well.

kevex91:
temporary fix, replace code in this file:

<img width="1512" alt="Screenshot 2019-06-04 at 9 30 48 AM" src="https://user-images.githubusercontent.com/20353382/59817836-140b0500-933f-11e9-9b6a-ed5788d47023.png">


```
//
//  MediaURL.swift
//  RNTrackPlayer
//
//  Created by David Chavez on 12.08.17.
//  Copyright © 2017 David Chavez. All rights reserved.
//

import Foundation

struct MediaURL {
    let value: URL
    let isLocal: Bool
    private let originalObject: Any
    
    init?(object: Any?) {
        guard let object = object else { return nil }
        originalObject = object
        
        if let localObject = object as? [String: Any] {
            let uri = localObject["uri"] as! String
            isLocal = uri.contains("http") ? false : true
            let encodedURI = uri.addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed)!
            value = URL(string: encodedURI.replacingOccurrences(of: "file://", with: ""))!
        } else {
            let url = object as! String
            isLocal = url.lowercased().hasPrefix("file://")
            value = URL(string: url.replacingOccurrences(of: "file://", with: ""))!
        }
    }
}

```

kevex91:
I see that this "fix" was done a few years ago and then reverted.

[...]

PR #631, opened 2019-06-20
[iOS] add optional urlDecoding

kevex91:
stems from this issue: #543 

- allow setting a `iosUrlDecoding` boolean in the track player config
- store it in the TrackPlayer class
- disable decoding of URLs if it is set to false
- enable decoding by default

I'm not a swift developer, tell me if there is anything that I should change. 

I feel that having an option will satisfy everyone

@Guichaguri  @dcvz


jspizziri:
It looks like this was solved via https://github.com/doublesymmetry/react-native-track-player/issues/543

Closing this issue.

PR #634, opened 2019-06-24
Split Metadata Functionality

dcvz:
@Guichaguri I propose that we stop being opinionated about how to update the information on the control centre / notification.

Instead, let's allow the user to update the metadata for an upcoming track (similar to how we already do), that we let them update the data that we're currently showing (doesn't have to be tied to the tracks) and we allow them to clear it.

This way we give users flexibility and control. Therefore satisfying a lot more needs.


ios/RNTrackPlayer/RNTrackPlayer.swift
```diff
@@ -454,30 +454,27 @@ public class RNTrackPlayer: RCTEventEmitter {
...
-                    })
-                    self?.player.nowPlayingInfoController.set(keyValue: MediaItemProperty.artwork(artwork))
-                }
-            }
+            Metadata.update(for: player, with: metadata)
         }
     }
+    
+    @objc(clearNowPlayingMetadata:rejecter:)
+    public func clearNowPlayingMetadata(resolve: RCTPromiseResolveBlock, reject: RCTPromiseRejectBlock) {
```
dcvz: @Guichaguri this and the following are the new additions

Guichaguri:
I'd love to see use-cases where `updateMetadataForTrack` isn't enough. Splitting the metadata functionality isn't a hard thing to do, but can cause many unwanted results (e.g. the player automatically updating the metadata and your custom metadata being overridden)

dcvz:
I think users who are making less conventional players might need things like this. I for example, have built some of these things custom-ly into the player. Clearing the metadata is something I really use often; because I don't want to have to stop the player for that information to clear.

I think however, it's all part of a bigger change that we can decide on.

More recently I also needed to continuously update the icons in the notification / control centre based on actions the user has taken. For example on iOS you can like and dislike / bookmark from the control centre.. however to update the state of that icon you must again set the icons with new state.

At the moment since all of this happens with the `setup` you can only do this once.

I think it's perhaps time that we treat both as different entities. We can even have an option to the play whether we should "automatically set the metadata for you" or not. that way you can expect your data to be possibly overwritten or not.

dcvz:
I wanted to update this PR further with a separation of setting notification actions.

dcvz:
I think at time where you might be listening to live audio -- while it's true you can still make it work by updating the current's track metadata.. that feels like a "hacky" solution.

In that case I wouldn't expect that one track should have multiple metadata.

In my particular case, I actually have 2 players that I "switch-between" (custom use case); however, it'd be nice to also be able to control what metadata is shown -- without having to choose with player's track is the chosen one.

I feel like there might be others as well. What I can do, is add one more config variable that let's you choose whether the player should automatically set the metadata, or not.

ios/RNTrackPlayer/RNTrackPlayer.swift
```diff
@@ -149,15 +149,19 @@ public class RNTrackPlayer: RCTEventEmitter {
         let autoWait: Bool = config["waitForBuffer"] as? Bool ?? false
         player.automaticallyWaitsToMinimizeStalling = autoWait
         
+        // configure if control center metdata should auto update
+        let autoUpdateMetadata: Bool = config["automaticallyUpdateNowPlayingInfo"] as? Bool ?? true
```
dcvz: @Guichaguri I've added this setting now :)

src/interfaces.ts
```diff
@@ -40,7 +40,16 @@ export interface PlayerOptions {
...
+  /**
+   * Indicates whether the player should automatically delay playback in order to minimize stalling.
+   * Defaults to `false`.
+   */
   waitForBuffer?: boolean
+  /**
+   * Indicates whether the player should automatically update now playing metadata data in control center / notification.
+   * Defaults to `true`.
+   */
+  automaticallyUpdateNowPlayingInfo?: boolean
```
dcvz: @Guichaguri I've added this config setting now - allowing you to choose whether you want the player to automatically update the notification / control centre metadata based on the playback items.

dcvz:
I think is pretty much ready @Guichaguri - we just have to align on how we feel about these changes.

ios/RNTrackPlayer/RNTrackPlayer.swift
Guichaguri: This parameter looks too long, how about something like `autoMetadataUpdate`?

ios/RNTrackPlayer/Utils/Metadata.swift
```diff
@@ -0,0 +1,49 @@
...
+        
+        if let artworkURL = MediaURL(object: metadata["artwork"]) {
+            URLSession.shared.dataTask(with: artworkURL.value, completionHandler: { [weak player] (data, _, error) in
+                if let data = data, let image = UIImage(data: data), error == nil {
+                    let artwork = MPMediaItemArtwork(boundsSize: image.size, requestHandler: { (size) -> UIImage in
+                        return image
+                    })
+                    player?.nowPlayingInfoController.set(keyValue: MediaItemProperty.artwork(artwork))
+                }
+            }).resume()
```
Guichaguri: What if I spam `updateNowPlayingMetadata` with different artworks? Wouldn't the artwork get messed up? Canceling the previous request might be a good idea here.

src/interfaces.ts
```diff
@@ -161,6 +170,15 @@ export interface TrackMetadata {
   artwork?: string | ResourceObject
 }
 
+export interface NowPlayingMetadata {
```
Guichaguri: I feel like this interface should extend `TrackMetadata`, as everything is pretty much the same, except the extra `elapsedTime`.

ios/RNTrackPlayer/RNTrackPlayer.swift
dcvz: Changed to `autoUpdateMetadata`

src/interfaces.ts
dcvz: Yes and no.. because the type checker will believe you can pass other items that are not going to be applied. But I have fixed this by abstracting a Base.

dcvz:
Address the issues @Guichaguri :) I will go ahead and merge this unless you wish to make Android updates also in this branch?

Guichaguri:
Don't merge it yet, I'll be able to work on this in the next weekend.

src/interfaces.ts
Guichaguri: Shouldn't this be `autoUpdateMetadata` instead?

src/interfaces.ts
```diff
@@ -149,16 +158,23 @@ export enum State {
...
+  artwork?: string | ResourceObject
+}
+
+export interface TrackMetadata extends TrackMetadataBase {
   description?: string
   genre?: string
   date?: string
   rating?: number | boolean
-  artwork?: string | ResourceObject
+}
```
Guichaguri: Why do we have to split out `description`, `genre`, `date` and `rating` from the now playing metadata?

src/interfaces.ts
```diff
@@ -149,16 +158,23 @@ export enum State {
...
+
+export interface NowPlayingMetadata extends TrackMetadataBase {
+  elapsedTime?: number
```
Guichaguri: Does the state / bufferedTime / playbackRate come directly from the player?

dcvz:
@Guichaguri is this read to merge? :)

Guichaguri:
@dcvz I've left a few reviews, can you address them so we can merge it? 😃

Guichaguri:
I've left a few reviews, I think they're the only thing left to address, then we can merge.

PR #650, opened 2019-07-03
Update hooks.js

alangumer:
usePlaybackState hook is always returning STATE_NONE which is fine if you are only using it in one part of your app, but if you have multiple views you need the actual player state. This change fixes it.


Guichaguri:
Thank you!

PR #654, opened 2019-07-05
Modernize Library Bundling

dcvz:
Rewrites library in typescript with proper `lib` and `types` output also adds `eslint` support :)

**WARNING**
With we deprecate the `ProgressComponent` which I think is ok as most people are on hooks now and we have an implementation for progress with hooks.

**BREAKING**
Now only direct methods on the player are exported as default on `TrackPlayer`. `State`, `Capability`, `PitchAlgorithm`, `TrackType` and `Event` are now exported as enums.

To use them you'll have to `import { State } from 'react-native-track-player'`

Same applies to hooks `import { useTrackPlayerProgress }  from 'react-native-track-player'`

I've updated the example accordingly.


dcvz:
I'll run some tests on existing project and then merge this @Guichaguri

dcvz:
With this we'd deprecate the `ProgressComponent` which I think is ok as most people are on hooks now. We can make this part of the `2.0.0` release.

dcvz:
BREAKING: `State`, `Capability`, `PitchAlgorithm`, `TrackType` and `Event` are now exported as enums and not part of the `default export`.

To use them you'll have to `import { State } from 'react-native-track-player`

I've updated the example accordingly.

Guichaguri:
Nice! That should be going into 2.0.0, right?

dcvz:
Yup @Guichaguri I'll merge it in now, but should be part of the 2.0.0 release

dcvz:
@Guichaguri Yes! I've merged this now into a `v2` branch where we can collect work

PR #685, opened 2019-08-01
Add support for InitialTiming interface to Track

zzorba:
We are trying to workaround the issue where seekTo doesn't work on iOS initially, I added this new parameter to the track `initialTime` which is already consumed by the underlying iOS library.  Several github issues referenced this problem, including https://github.com/react-native-kit/react-native-track-player/issues/548

No android support is included, as far as I know it does not exhibit this problem.

### DIFF
--- a/ios/RNTrackPlayer/Models/Track.swift
+++ b/ios/RNTrackPlayer/Models/Track.swift
@@ -10,7 +10,7 @@ import Foundation
 import MediaPlayer
 import AVFoundation
 
-class Track: NSObject, AudioItem, TimePitching, Authorizing {
+class Track: NSObject, AudioItem, InitialTiming, TimePitching, Authorizing {
     let id: String
     let url: MediaURL
     
@@ -22,6 +22,7 @@ class Track: NSObject, AudioItem, TimePitching, Authorizing {
     var genre: String?
     var duration: Double?
     var skipped: Bool = false
+    var initialTime: TimeInterval = 0.0;
     var artworkURL: MediaURL?
     let headers: [String: Any]?
     let pitchAlgorithm: String?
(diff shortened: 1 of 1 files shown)


Guichaguri:
We are trying to remove Android and iOS inconsistencies where possible. Either we add `initialTime` to Android too or change `seekTo` on iOS. @dcvz thoughts?

curiousdustin:
It would be really nice if we could use `initialTiming` or similar solution when adding the track. Attempting to seek immediately after play is unintuitive, when what you are really trying to do is start playing from x time. Added to that, it seems seeking doesn't always work as expected without added logic about state, so it is even more cumbersome to achieve the desired action of starting playback at a certain time.

zzorba:
Yeah, I've reviewed the ExoPlayer documentation and am unclear how this would translate over to Android.  I've basically run into the issue where seeks on iOS before play don't seem to work, and because the underlying library offered the feature it seemed like exposing it unblocks this use case.

But I understand that library differences are undesirable.  

Would be happy to attempt an Android implementation with some guidance on the best way to proceed.

thorbenprimke:
@zzorba - Are you working on the Android implementation for this? If not, I'll take it on as we need this feature for both platforms.

curiousdustin:
@thorbenprimke , that would be amazing, thanks!

curiousdustin:
Got something working on both iOS and Android on my fork. Probably not a final solution for everyone, but it seems to work.

https://github.com/react-native-kit/react-native-track-player/pull/713#issuecomment-564782799

Guichaguri:
Superseded by #713

kartavyaparekh96:
hey @curiousdustin @Guichaguri initial time change getposition but track start from 0 in iOS. Please guide me

curiousdustin:
As mentioned above and other places this behavior is not supported out of the box in react-native-track-player. For my project I currently use a fork with modifications to support my exact use case, which likely would not work for everyone.

PR #688, opened 2019-08-03
Queue improvements

### LINKED ISSUE #122: Playlist controls (ie loop/shuffle/repeat)
Does this library come with an API to control how to music plays? I've read through the docs and since I wasn't able to find it, I would appreciate some suggestions as to how to implement this behavior. I am using redux and redux-saga to handle state/effects. I've thought about implementing this logic within the listeners whenever events are triggered. But a lot of the time the app will be in the background or the screen will be locked. Does the JS still execute in those cases? Still a bit new, Thanks.

Guichaguri:
This PR adds a few queue improvements. Closes #122.

* Replaced track IDs to track queue indexes
    * Functions that used to reject with `track_not_in_queue` or `invalid_id` will now reject with `index_out_of_bounds` when the index is invalid.
* Added `move(index, newIndex)` - Moves a track from an index to another.
    * Returns a promise, rejecting with `index_out_of_bounds` when any of the indexes are invalid.
* Added `shuffle()` - Shuffles the queue using the Fisher-Yates algorithm.
    * Returns a promise, resolving when the shuffle is finished.
* Added `setRepeatMode(mode)` - Updates the repeat mode:
    * `REPEAT_OFF` - doesn't repeat.
    * `REPEAT_TRACK` - loops the current track.
    * `REPEAT_QUEUE` - repeats the whole queue.
* Added `getRepeatMode()`

### DIFF
--- a/android/src/main/java/com/guichaguri/trackplayer/service/player/LocalPlayback.java
+++ b/android/src/main/java/com/guichaguri/trackplayer/service/player/LocalPlayback.java
@@ -144,13 +144,20 @@ public class LocalPlayback extends ExoPlayback<SimpleExoPlayer> {
     @Override
     public void shuffle(final Promise promise) {
         Random rand = new Random();
+        int startIndex = player.getCurrentWindowIndex();
         int length = queue.size();
 
-        // Fisher-Yates shuffle
-        for (int i = 0; i < length; i++) {
-            int swapIndex = rand.nextInt(i + 1);
+        if (startIndex == C.INDEX_UNSET) {
+            startIndex = 0;
+        } else {
+            startIndex++;
+        }
 
-            queue.add(swapIndex, queue.remove(i));
+        // Fisher-Yates shuffle
+        for (int i = startIndex; i < length; i++) {
+            int swapIndex = rand.nextInt(i + 1 - startIndex) + startIndex;
+
+            Collections.swap(queue, i, swapIndex);
 
             if (length - 1 == i) {
                 // Resolve the promise after the last move command
--- a/android/src/main/java/com/guichaguri/trackplayer/service/player/LocalPlayback.java
+++ b/android/src/main/java/com/guichaguri/trackplayer/service/player/LocalPlayback.java
@@ -126,7 +126,7 @@ public class LocalPlayback extends ExoPlayback<SimpleExoPlayer> {
 
     @Override
     public void move(int index, int newIndex, Promise promise) {
-        queue.add(newIndex, queue.remove(index));
+        Collections.swap(queue, index, newIndex);
         source.moveMediaSource(index, newIndex, manager.getHandler(), Utils.toRunnable(promise));
     }
 
--- a/android/src/main/java/com/guichaguri/trackplayer/service/player/LocalPlayback.java
+++ b/android/src/main/java/com/guichaguri/trackplayer/service/player/LocalPlayback.java
@@ -73,7 +73,7 @@ public class LocalPlayback extends ExoPlayback<SimpleExoPlayer> {
     public void add(Track track, int index, Promise promise) {
         queue.add(index, track);
         MediaSource trackSource = track.toMediaSource(context, this);
-        source.addMediaSource(index, trackSource, manager.getHandler(), Utils.toRunnable(promise));
+        source.addMediaSource(index, trackSource, manager.getHandler(), () -> promise.resolve(index));
 
         prepare();
     }
@@ -87,7 +87,7 @@ public class LocalPlayback extends ExoPlayback<SimpleExoPlayer> {
         }
 
         queue.addAll(index, tracks);
-        source.addMediaSources(index, trackList, manager.getHandler(), Utils.toRunnable(promise));
+        source.addMediaSources(index, trackList, manager.getHandler(), () -> promise.resolve(index));
 
         prepare();
     }
--- a/android/src/main/java/com/guichaguri/trackplayer/service/player/LocalPlayback.java
+++ b/android/src/main/java/com/guichaguri/trackplayer/service/player/LocalPlayback.java
@@ -124,12 +124,6 @@ public class LocalPlayback extends ExoPlayback<SimpleExoPlayer> {
     }
 
     @Override
-    public void move(int index, int newIndex, Promise promise) {
-        Collections.swap(queue, index, newIndex);
-        source.moveMediaSource(index, newIndex, manager.getHandler(), Utils.toRunnable(promise));
-    }
-
-    @Override
     public void removeUpcomingTracks() {
         int currentIndex = player.getCurrentWindowIndex();
         if (currentIndex == C.INDEX_UNSET) return;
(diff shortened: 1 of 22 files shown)


Guichaguri:
@dcvz I still need to work on the JS side, docs and testing, but all of the native code for Android is already there. Let me know what do you think about these changes.

dcvz:
From a quick look, this looks great! I'll work on the iOS implementation of this soon :)

samtoya:
Any progress?

Drazail:
@Guichaguri 
This will be a great addition and a major improvement to the module.
however, ditching the old "by-id" methods will introduce a breaking change, it would be nice if we could keep them. For instance, the architecture of our app revolves around unique IDs we recieve from our back-end service.

internally we have implemented the following methods:

- addByIndex
- removeByIndex
- shuffle
- shuffleFromIndex
- move
- skipbyIndex
- updateMetadataForTrackByIndex
- setRepeatMode
- getRepeatMode

the native logic behind these methods are virtually identical to the methods you've implemented in this branch.

I think it would be a great if we could keep the old methods as a means of backwards compatibility.

-------------------
Also we added the following methods as an improvement to the Q system:

-updateTrack
-updateTrackbyIndex

these methods will update the media source as well as metadata in case we have to change a TrackObject programmatically.

let me know if this isn't considered to be an anti-pattern so I can make a pull request.

skkshoaib:
Hi,
Any updates on this?

Guichaguri:
@Drazail We're working on v2 to improve the API and intentionally make breaking changes. We certainly could implement a backwards-compatible layer, but I don't see a reason why we should.

* Not everyone has unique IDs for their tracks, they would have to implement a system to create IDs for them
* If they do have unique IDs, they wouldn't be able to add the same track twice without creating two separate IDs
* I've seen many cases that they lookup the queue array to find the ID of a track based on index, for then the module to translate the ID back into a index.

That being said, I don't see a reason to keep IDs in an API design standpoint.
The compatibility layer can also be implemented in pure JS, it's not like it needs native code in order to work.

Drazail:
@Guichaguri 

well, backwards compatibility is generally considered to be a major reason to keep the API consistent between different releases.

As for pure JS implementation of a backward compatible layer, even if it was a viable option from a performance standpoint, that's really not something to be delegated to the end user.

Anyways, at this point, our fork and this repo are so different that I understand trying to keep the API similar doesn't worth the effort.

android/src/main/java/com/guichaguri/trackplayer/service/player/LocalPlayback.java
```diff
@@ -120,6 +118,12 @@ public void remove(List<Integer> indexes, Promise promise) {
         }
     }
 
+    @Override
+    public void move(int index, int newIndex, Promise promise) {
+        queue.add(newIndex, queue.remove(index));
+        source.moveMediaSource(index, newIndex, manager.getHandler(), Utils.toRunnable(promise));
+    }
+
```
curiousdustin, line 132: @Guichaguri, Are we allowing using `move()` on the current index. SwiftAudio does not allow this. I would have to dig deeper to understand why.

Also just wondering, does `queue.add(newIndex, queue.remove(index));` actually insert at the correct slot, if `newIndex` is greater than `index`? Trying to learn a little more Java as I go along here. Is there a reason you're not using swap() instead?

Guichaguri: If SwiftAudio doesn't allow it, we can make Android reject that too. It would actually make it more consistent with `remove()` which doesn't allow removing the current track either.

For some reason I didn't even thought about `swap()` when I was writing this, nor that removing an item would shift all following indexes

curiousdustin: I guess I have the same basic question about shuffle. Doesn’t shuffle potentially move the current track?

Guichaguri: You're right, we can make it not move at all and move all the other tracks
Or we can also make it move all following tracks, so that previous ones don't play again, which makes more sense

Guichaguri:
@Drazail I'm still thinking about whether or not we should get rid of IDs, the big advantage is that the IDs stay the same no matter the position of the track in the queue.

So I guess we should keep both. It shouldn't hurt to have the ID-based functions in the library. Making the native functions such as `skip()`, `getTrack()` or `move()` use indexes and implementing functions like `skipToId()`, `getTrackById()`, `moveById()` in pure JS:
```js
async function skipToId(id) {
    const queue = await TrackPlayer.getQueue();
    TrackPlayer.skip(queue.findIndex(track => track.id === id));
}
```

We could also make a separated file just as a v1 compatibility layer built on top of the new API: 
```js
import TrackPlayer from 'react-native-track-player'
TrackPlayer.skip(index)

import TrackPlayer from 'react-native-track-player/v1'
TrackPlayer.skip(id)
```

So updating it will be as easy as changing the imports, so you don't have to update all the code to the new API at once, you can update it as you go. What do you think?

puckey:
Another idea could be to use immer to mutate the queue: https://github.com/immerjs/immer
Allowing for:

```js
TrackPlayer.changeQueue((queue) => {
  queue.pop();
  queue.unshift({ track });
  queue[2].title = "another title";
});
```

Guichaguri:
@puckey That looks good, but how the native code would react to these changes?

We could send the whole queue to the native code and then update track by track, but that's too much overhead if you have a big queue and only want to change the title of a specific track.

It looks like Immer has patches, which we could potentially parse and apply through native code, but that would require too much code and testing on our side, basically having to recreate that library in native code.

We could also implement a simpler diff algorithm that would only send the tracks that had been updated to the native code, that needs more thought but could work.

Another possibility would be to implement our own "queue" class that has the same functions as an Array but apply those changes using the API currently in place.

At this point I think we should create a new issue for discussing utility functions (library compatibility, React classes and hooks, etc)

Drazail:
@Guichaguri 

I do agree that the implementation we used is bloating the API, and the second option you suggested is probably the way to go.

However, there's another idea of the top of my head: restrict ID types to be strings. this can help the module decide which layer ( compatibility layer or the core API) should handle the call, ie:

```TrackPlayer.skipTo("foo")``` skips to "foo" id 
```TrackPlayer.skipTo(3)``` skips to the 4th track in Q

I personally prefer to give users as many options as possible while keeping the API as stable and as simple as possible hence reducing the barrier of entry, but then again, as you have more knowledge of this library than anyone else, this comes down to what you think is the more appropriate behavior.

neoassyrian:
I'm with @drazail all the way. It makes sense to me,especially the string vs int.

gustavo-dev:
Any updates on this?

harerakalex:
> @dcvz I still need to work on the JS side, docs and testing, but all of the native code for Android is already there. Let me know what do you think about these changes.

Well done @Guichaguri, Have you finished the Js part, if yes, when should we expect this PR to be merged?

zeejaydev:
Any update on this yet?

dcvz:
This branch was merged with only the changes to an index based queue.

Other queue methods like:
- `move(index, newIndex)`
- `shuffle()`
- `setRepeatMode(mode)`
- `getRepeatMode()`

will be added separately as a different effort because they are complex and affect the notifications system. At the moment I am trying to prioritise getting a stable build of v2 out.

PR #690, opened 2019-08-06
[iOS] Add swift version to podspec

Bang9:
added Swift version, default version is Swift 4.0(with Auto linking)
It occur Swift compile error when building after auto link, so we need specify Swift version in pod spec (it can be fixed by manually choose Swift version above 4.2, but it is not efficiently)

image

I'm using react-native-track-player@2.0.0-rc13


ovr:
ping @dcvz @guichaguri

Can you merge this onliner fix? This will fix:

```- `react-native-track-player` does not specify a Swift version and none of the targets (`XXXXXXX`) integrating it have the `SWIFT_VERSION` attribute set. Please contact the author or set the `SWIFT_VERSION` attribute in at least one of the targets that integrate this pod.```


And cherry pick to 1.x branch

PR #692, opened 2019-08-06
Create a TrackPlayer module

K0rdan:
This pull request is meant to create a TrackPlayer module and update index.js, Hooks and ProgressComponent modules to avoid require cycles


dcvz:
Thanks @K0rdan we've redone our module for native Typescript support so I believe this issue has now been solved.

PR #704, opened 2019-08-16
support preferredPeakBitRate

ananevam:


Guichaguri:
Can you specify the use-case for exposing this property?

bradleyflood:
Looks like a good idea. If there is more need for this let us know :)

PR #707, opened 2019-08-21
Add prioritizeTimeOverSize setting for Android buffer

lukebaker:
Default is `false` to maintain backward compatibility. If you're configuring minBuffer and maxBuffer, you probably want to set this to `true`. Setting to `true` ensures that your `minBuffer` setting will always be met when possible. When it is false, ExoPlayer has a separate targetBufferSize that may trigger loading of the resource to pause even if minBuffer is not met.

In ExoPlayer, audio tracks have a [default buffer size of ~3.5MB](https://exoplayer.dev/doc/reference/constant-values.html#com.google.android.exoplayer2.C.DEFAULT_AUDIO_BUFFER_SIZE). If I have an audio file that's 12MB and set `minBuffer` to 10 minutes and use the default settings, ExoPlayer will not fill the minimum buffer to 10 minutes. It will stop filling the buffer after it has ~3.5MB of data. However, if I set `prioritizeTimeOverSize` to `true`, then ExoPlayer will fill the buffer until we've reached `minBuffer` rather than stopping at the default audio buffer size. The logic for this can be found in [`shouldeContinueLoading()` in ExoPlayer](https://github.com/google/ExoPlayer/blob/44293e8f45fc62ff918001b4425c44d0ded5ffc9/library/core/src/main/java/com/google/android/exoplayer2/DefaultLoadControl.java#L365-L381).

I would love to see a new 1.1.x release with this change in it. I've left the default as `false` to ensure that folks who upgrade will see the same behavior unless they set `prioritizeTimeOverSize`.

### DIFF
--- a/android/src/main/java/com/guichaguri/trackplayer/service/MusicManager.java
+++ b/android/src/main/java/com/guichaguri/trackplayer/service/MusicManager.java
@@ -120,11 +120,13 @@ public class MusicManager implements OnAudioFocusChangeListener {
         int playBuffer = (int)Utils.toMillis(options.getDouble("playBuffer", Utils.toSeconds(DEFAULT_BUFFER_FOR_PLAYBACK_MS)));
         int backBuffer = (int)Utils.toMillis(options.getDouble("backBuffer", Utils.toSeconds(DEFAULT_BACK_BUFFER_DURATION_MS)));
         long cacheMaxSize = (long)(options.getDouble("maxCacheSize", 0) * 1024);
+        boolean prioritizeTimeOverSize = options.getBoolean("prioritizeTimeOverSize", false);
         int multiplier = DEFAULT_BUFFER_FOR_PLAYBACK_AFTER_REBUFFER_MS / DEFAULT_BUFFER_FOR_PLAYBACK_MS;
 
         LoadControl control = new DefaultLoadControl.Builder()
                 .setBufferDurationsMs(minBuffer, maxBuffer, playBuffer, playBuffer * multiplier)
                 .setBackBuffer(backBuffer, false)
+                .setPrioritizeTimeOverSizeThresholds(prioritizeTimeOverSize)
                 .createDefaultLoadControl();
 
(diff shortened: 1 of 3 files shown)


Guichaguri:
I'm not sure about this. I think we're exposing too many low-level properties that are rarely going to be used. In this case, I don't see an use-case where someone will set `minBuffer` but leave this as `false` intentionally.

Should we automatically set this to true whenever the `minBuffer` property is set?

Also, sorry for taking to long to respond

lukebaker:
> I'm not sure about this. I think we're exposing too many low-level properties that are rarely going to be used. In this case, I don't see an use-case where someone will set `minBuffer` but leave this as `false` intentionally.
> 
> Should we automatically set this to true whenever the `minBuffer` property is set?

For my use-case, that'd be fine. I wanted to err on the side of not introducing breaking changes as I don't know all the scenarios that folks might choose to use `minBuffer`. If you're happy with setting `prioritizeTimeOverSize` to `true` whenever `minBuffer` is set, that's good with me. :+1:

curiousdustin:
@lukebaker can you show where this setting defaults to false?

I was looking at this to potentially create a PR. However, the behavior you describe is not what happens for me. My tracks are buffering past 3.5MB using a long `minBuffer`.

Also when looking at the ExoPlayer code you linked to, it [seems that](https://github.com/google/ExoPlayer/blob/44293e8f45fc62ff918001b4425c44d0ded5ffc9/library/core/src/main/java/com/google/android/exoplayer2/DefaultLoadControl.java#L62) `DEFAULT_PRIORITIZE_TIME_OVER_SIZE_THRESHOLDS` defaults to `true`.

Maybe I'm missing something?

lukebaker:
@curiousdustin thanks for looking into this and patience with my response. I think you're absolutely right: `prioritizeTimeOverSize` defaults to `true` and this pull request is unnecessary. I tested my fork as well as 1.1.8 and 1.2.2 and they all behaved similarly in terms of buffering a ~46 minute / 22MB mp3 file streaming over HTTP.

PR #716, opened 2019-08-29
Send data with playback end event

florianalbrecht:
We sometimes ran into a race condition where self.currentItem was already changed when the the playbackEnd event was handled. That's why we added it as payload to the event (see also https://github.com/dcvz/SwiftAudio/pull/6).


biesbjerg:
Experiencing the same. It makes it impossible to reliably detect when last track in queue has finished playing.

curiousdustin:
@florianalbrecht @biesbjerg , I think v1.2.3 solved this in a different way. Do you have time to check it?

PR #717, opened 2019-08-29
Handle playback stalled event

florianalbrecht:
When losing internet connection and running out of cache while playing a file-based stream on iOS the playback cannot be resumed when the connection is back. This PR handles the playbackStalled event which was added to SwiftAudio in this PR: https://github.com/dcvz/SwiftAudio/pull/7

### DIFF
--- a/ios/RNTrackPlayer/RNTrackPlayer.swift
+++ b/ios/RNTrackPlayer/RNTrackPlayer.swift
@@ -79,6 +79,7 @@ public class RNTrackPlayer: RCTEventEmitter {
             "playback-state",
             "playback-error",
             "playback-track-changed",
+            "playback-stalled",
             
             "remote-stop",
             "remote-pause",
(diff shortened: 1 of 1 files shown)


biomancer:
Could you please give an example on how to handle `playback-stalled` event to be able to resume playback when connection is back?

florianalbrecht:
@biomancer In our app we only close the audio player.. I didn't find a good solution to resume playback in this case.

biomancer:
@florianalbrecht could this be related to https://github.com/jorgenhenrichsen/SwiftAudio/issues/56 ? I see that the fix for this issue is included in your version of SwiftAudio, and I hoped it would help for such cases, it's a pity if it's not :(

florianalbrecht:
@biomancer I don't think this is related because in my case the player didn't change to failure state. It is just firing the AVPlayerItemPlaybackStalled notification and for file-based playback it cannot be resumed after the connection is back (see https://developer.apple.com/documentation/foundation/nsnotification/name/1387661-avplayeritemplaybackstalled).

curiousdustin:
https://github.com/dcvz/SwiftAudio/issues/7 doesn't seem to exist anymore. Do know how we can view this change?

Was this or similar ever added to the main SwiftAudio repo?

jspizziri:
I'm closing this PR due to inactivity. Please let me know if it should be reopened.

PR #722, opened 2019-09-04
Update hooks - reset progress when playback stops

alangumer:


alangumer:
@dcvz can you please include this in the latest tag 2.0.0-rc13?

src/hooks.ts
```diff
@@ -83,6 +83,10 @@ export function useProgress(updateInterval?: number) {
   }
 
   useEffect(() => {
+    if (playerState === State.Stopped) {
+      setState({ position: 0, duration: 0, buffered: 0 });
+      return;
+    }
```
Guichaguri, line 89: It's not the desirable solution, but it does fixes the issue pretty well

curiousdustin:
Was there an issue ticket associated with this? I’m not seeing it. Should a similar fix be applied to dev?

Guichaguri:
This is probably seen as a breaking change, but I think it will be superseded by #935

PR #725, opened 2019-09-04
Add isServiceRunning method to detect if music service is running

biomancer:
Adds method to check if playback service is running, which is required in android to be able to detect if TrackPlayer methods are safe to call as described here https://github.com/react-native-kit/react-native-track-player/issues/473#issuecomment-528120502

### DIFF
--- a/android/src/main/java/com/guichaguri/trackplayer/module/MusicModule.java
+++ b/android/src/main/java/com/guichaguri/trackplayer/module/MusicModule.java
@@ -159,6 +159,11 @@ public class MusicModule extends ReactContextBaseJavaModule implements ServiceCo
     }
 
     @ReactMethod
+    public void isServiceRunning(final Promise promise) {
+        promise.resolve(binder != null);
+    }
+
+    @ReactMethod
     public void destroy() {
         try {
             if(binder != null) {
--- a/index.d.ts
+++ b/index.d.ts
@@ -116,6 +116,7 @@ declare namespace RNTrackPlayer {
 
   export function setupPlayer(options?: PlayerOptions): Promise<void>;
   export function destroy(): void;
+  export function isServiceRunning(): Promise<boolean>;
 
   // Player Queue Commands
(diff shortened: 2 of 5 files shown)


biomancer:
That is not a complete fix for the mentioned issue, but at least will give an ability to write code in a way that app won't crash on android when service has been stopped by OS

Guichaguri:
I think we should reject functions instead of adding a new one for that. But we should also check whether the service *can* be initialized.

curiousdustin:
@Guichaguri , After leaving this on the back burner for several months, I was literally implementing this method by @biomancer now as a stopgap.

Are you able to share any plans for when you would be able to address this in the way you are recommending?

Guichaguri:
I think we should reject all methods when the app is in background and the service is not initialized, but also add a function that returns the same condition, so you can add checks instead of handling rejections. (Something like `isPlayerUsable()` and reject with `can-not-initialize`)

nickarora:
Somewhat shocked at how long this issue has been open.  People are reporting a 20% crash rate on android devices based on this bug alone; and yet new features like shuffled get prioritized...At the very least, allow this guy's PR to get merged...

hiroyaiizuka:
Why this PR has not been merged yet? 

This error still happens....

Kelvne:
Since this was not merged yet + 1 change to list of changes in our fork of this service 🤷‍♂️

mpivchev (approved):
Hey, thanks for all the input on this issue. The reason why we haven't taken it into account until now was because this project stopped being actively worked on for a while. This has changed recently.

This out of the way, we will accept this PR with the pretext that this function will be added to the android rewrite as well, to keep feature parity between the two versions. Since the android rewrite now checks if the service is running before executing most service related functions, this function will most likely be deprecated there.

mpivchev:
We've made a new PR that implements these changes, since this PR's commit history was broken: #1452

Issue #711, opened 2019-08-27
Playback abruptly stops on Android


farshed:
# Configuration

I'm using `react-native-track-player` 1.1.4 with `react-native` 0.59.9

# Issue

The playback randomly freezes on Android even when the app is in foreground. Controls like pause, play & progress bar also freeze and do not respond until the player is reset and a new track is added. (I'm managing my own queue on the JS side so there's always only one track in the native queue at a given time).

I've tested this on a Nexus 5X emulator running Oreo 8.0 and Oppo F11 running Pie 9.0. It should be noted that the problem seems to go away on my Oppo F11 after the **Smart Restriction of the Background Running** (Battery saving) is turned off in the app settings.
Also, the problem didn't seem to exist in the previous version of my app which used `react-native-track-player` 1.1.3.

**EDIT:** I did some digging and found several [similar issues](https://github.com/google/ExoPlayer/issues?utf8=%E2%9C%93&q=is%3Aissue+killed) with the Exoplayer.

# Code

Show us the code you are using

PR #723, opened 2019-09-04
[android] Fix Bad notification for startForeground error

biomancer:
This is updated https://github.com/react-native-kit/react-native-track-player/pull/579 fix by @puckey based on discussion there, fixes https://github.com/react-native-kit/react-native-track-player/issues/534 (or at least the crash mentioned in comment https://github.com/react-native-kit/react-native-track-player/issues/534#issuecomment-484014486 if it's not the one that issue author has encountered)

Issue #534, opened 2019-04-13
Galaxy S10 not working


sideokyou:
I tested on other Android devices and worked fine. but only Samsung Galaxy S10 is not working. After playing music, it crashes. Any idea what needs to be solved?

Guichaguri:
Do you have any crash logs?

ishigamii:
@Guichaguri I am having the same kind of report from user of Samsung Galaxy telling me that the sound is stopped as soon as the app goes in sleep mode (screen closing) it works if the screen is awake. 

For me I have this crash for user of (Galaxy S10+ and Galaxy S10e) : 
```
Fatal Exception: android.app.RemoteServiceException: Bad notification for startForeground: java.lang.RuntimeException: invalid channel for service notification: null
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1871)
       at android.os.Handler.dispatchMessage(Handler.java:106)
       at android.os.Looper.loop(Looper.java:214)
       at android.app.ActivityThread.main(ActivityThread.java:7032)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:494)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:965)
```

sideokyou:
![Screen Shot 2019-04-19 at 10 22 35 AM](https://user-images.githubusercontent.com/47291956/56435394-1690a600-628d-11e9-8e8c-53004658862f.png)


I am still figuring out why it shows this error on this device(galaxy s10) specifically.

I am not sure, but this error seems not related to react-native-track-player.  Commented all TextInput, and set all color properties as a static value such as #e2e2e2. Still showing this error.

Any help? 

Thank you in advance,

PR #729, opened 2019-09-09
Fixed: Notification startForeground and Wifi Lock

xgustavoh:
This update fixes the issue already cited in [#473-Comment](https://github.com/react-native-kit/react-native-track-player/issues/473#issuecomment-507167261), #534  and partially fixes issue #711.

It also replaces Pull: #579 (@minhtc) and #723 (@biomancer and @puckey).

-------
This problem I've been trying to fix for a long time, from Friday to Monday was successful in a Samsung where formerly was disconnecting from Wi-Fi.

Problem #711 can still happen due to some Stop error, which I'm also having, which kills the player and takes a long time to play again.

Note: If you still have a background issue, the issue may be in the notification, as the 'startForeground' command is called again, instead of updating the current notification, I don't know if this is true, but I'm still studying this possibility.


xgustavoh:
I am working on an improvement of this code due to other issues found in the notification. Because of this I am closing this pull request.

PR #772, opened 2019-10-29
Call `playback-track-changed` for last track

high-performance-hery:
Fixes an issue on iOS where `playback-track-changed` isn't called when switching from the penultimate track of the queue to the last track, causing metadata not to update.

This behaves as expected on Android.

### DIFF
--- a/ios/RNTrackPlayer/RNTrackPlayer.swift
+++ b/ios/RNTrackPlayer/RNTrackPlayer.swift
@@ -185,7 +185,7 @@ public class RNTrackPlayer: RCTEventEmitter {
         player.event.playbackEnd.addListener(self) { [weak self] reason in
             guard let `self` = self else { return }
 
-            if (reason == .playedUntilEnd) {
+            if reason == .playedUntilEnd {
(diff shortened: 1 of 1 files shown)


sslash:
Nice one! 🤩

ntk860:
@hery-finimize your solution works but it trigger playback-track-changed one additional time every-time a track changes. I have some business logic on playback-track-changed and I have to do a trick to ignore the extra firing.

octodhruv:
Hi! How come this PR hasn't been merged yet? I think it fixes some pretty important stuff with the library....

high-performance-hery:
Good point @ntk860, https://github.com/react-native-kit/react-native-track-player/pull/772/commits/b991d4e8f54abdcdba31a4e619687ecb91f0c473 should fix it!

matt-dalton:
Any progress on getting this merged?

octodhruv:
what's the status with this? is there anything in particular that's blocking this from being merged? This seems like quite a serious issue with the listeners?

PR #836, opened 2020-01-17
Feature/ios live stream option

Definitely-Not-Vlad:
I've added another `option` for use in `TrackPlayer.updateOptions()` called `isLiveStreamSource` (`bool`).

This signifies to the iOS player that lock screen controls should not show a progress bar and timers, since the source is a live feed (e.g. a radio stream).

I branched this off the `v2.0.0-rc13` tag directly instead of from `v2`'s current state.


nesinervink:
When merged with 1.2.2, in my case it did break remote player functionality.
If player is set up with `isLiveStreamSource: false`, everything works as expected, but once option is upated (or initial value is set) to `true` - remote controls stop working.

curiousdustin:
@nesinervink can you try again with 1.2.3?

curiousdustin:
@Guichaguri is there something similar that can/should be applied on Android?

puckey:
I gave this a try on my fork based on the 1.2.3 branch and the remote controls also show up as disabled for me...
IMG_BD4672D59964-1

puckey:
I was able to get this working on 1.2.3.. Happy to do a PR.. but I am wondering if it doesn't make more sense as a property of the track object.. `live: true` perhaps? or `stream: true`

curiousdustin:
I agree with @puckey, a track property seems to make more sense for this feature.

Guichaguri:
Does it need to be a property? I was thinking about something that worked straight out of the box such as checking whether the duration is zero, which not only makes sense for livestreams, but also for tracks that have unknown duration.

dcvz:
This is a useful addition @Definitely-Not-Vlad -- would you mind rebasing this against the latest main (v2) branch?

dcvz:
It would be great if this was automatically detected and we knew it was a stream -- however, I'm a bit wary about just assuming we know what's best for the user, for example is there some situations where they might not want this functionality?

If so, we aren't really providing a way to opt out of this.. so it might still be best to have the users explicitly opt-in. However, I do think it would be better suited as a track property like (album artwork is) rather than globally on the player config.

gavrichards:
Is this feature being implemented?

ionleu:
Hey guys,
any updates related to this?

gavrichards:
Yes we would really like to see this too! :)

dcvz:
Thanks for following up here again, I'll try to get this in soon as a property on `Track`. if anyone wants to make a PR that incorporates this as a property on Track, that would be very much appreciated!

dcvz:
This just landed in https://github.com/DoubleSymmetry/react-native-track-player/commit/0116789c06458a6b20c5855f1b7c8f8d2eb4fb13. It will go out with the next release -- in the meantime you can try it by pointing your projects against the main branch.

gavrichards:
@dcvz do you know when this is likely to land in a release? :)

PR #841, opened 2020-01-24
refactor(Android): getState() directly returns STATE_NONE when service is disconnected

Rapsssito:
`TrackPlayer.getState()` now directly returns **STATE_NONE** if there is no service running, instead of waiting for the service creation in `waitForConnection()`.


curiousdustin:

PR #900, opened 2020-04-16
Rename PlaylistScreen component function name in example

ChickenAhn:
I think LandingScreen is copied from another component.
Rename this to a proper name.


curiousdustin:
Thanks! 👏

Issue #909, opened 2020-04-26
inconsistent playback-track-changed behaviour between android and ios


mynameisguy:
**Describe the bug**
when last track in queue finishes on iOS **playback-track-changed** is triggered with `nextTrack=null`, while in Android no **playback-track-changed** is fired.
On both platform **playback-queue-ended** event is called.

**To Reproduce**
letting last track in queue finish and check if **playback-track-changed** event is called

**Environment (please complete the following information):**
react: 16.9.0
react-native: 0.61.5
react-native-track-player: 1.2.2

testing on devices

PR #910, opened 2020-04-26
reporting last track changed on android

mynameisguy:
solves #909.
Since track changed seems to be identified while the player is playing the next track, the simplest way to solve this is by reporting changed track right before reporting that the queue ended


curiousdustin:
@guichaguri, what are your thoughts on this? At some point, on iOS, reporting the last track change to null was added. Should android be changed to match?

Guichaguri:
I'm okay with that, in fact, I think the queue ended event becomes useless now, I don't quite remember the reason why we have added it, but there's no reason to fire both events when only one of them can serve for both purposes quite well.

I'll merge this for v1, but I'd like to take the opportunity to open a discussion on removing the queue ended event.

PR #912, opened 2020-04-28
New feature: repeat Mode and Shuffle Mode | Android

ithustle:
I believe that those feature will be a good assets.


curiousdustin:
Thank you for your work on this, however, @Guichaguri has already completed this for Android. 

See #688

PR #925, opened 2020-04-30
Rework docs

puckey:
Continued from https://github.com/react-native-kit/react-native-track-player/pull/506

Unless there are any changes to suggest, this is ready for merging.


docs/API.md
```diff
@@ -209,42 +207,34 @@ class MyPlayerBar extends TrackPlayer.ProgressComponent {
...
+Track Player can be configured using a number of options. Some of these options pertain to the media controls available in the lockscreen / notification and how they behave, others describe the availability of capabilities needed for platform specific functionalities like Android Auto.
 
-## Options
-You can define the rating type, capabilities, icons, etc.
-Every parameter is optional, you don't have to set all of them.
-You can also call this function multiple times, you don't need to specify all the options, just the ones you want to change.
+You can change options multiple times. You do not need to specify all the options, just the ones you want to change.
 
-It's recommended to set the options right before using any other function (other than `setupPlayer`)
+It is recommended to set the options right before using any other function (other than `setupPlayer`)
```
curiousdustin, line 220: @guichaguri, can you provide any background on this recommendation? It seems to suggest doing it at other times could cause problems?

docs/API.md
```diff
@@ -256,9 +246,9 @@ TrackPlayer.updateOptions({
...
     nextIcon: require('./next-icon.png'),
-    icon: require('./notification-icon.png'), // The notification icon
+    icon: require('./notification-icon.png')
 });
 ```
 
 ### Notes
-* Due to the notification not being swipable, we highly recommend you to have a stop button in the notification. The button should `destroy()` the player.
\ No newline at end of file
+* Because you can not stop the player service by swiping the Android notification, we highly recommend you to have a stop button in the notification. The button should `destroy()` the player.
```
curiousdustin, line 254: @guichaguri, is there a simple explanation we could include here for why the notification cannot be swipable?

docs/index.md
```diff
@@ -33,8 +33,8 @@ A fully fledged audio module created for music apps. Provides audio playback, ex
 
 ## Example
 
-If you want to get started with this module, check the [Getting Started](./API.md) page.
-If you want detailed information about the API, check the [Documentation](./Documentation.md).
+If you want to get started with this module, check the [Getting Started](https://react-native-kit.github.io/react-native-track-player/api/) page.
+If you want detailed information about the API, check the [Documentation](https://react-native-kit.github.io/react-native-track-player/documentation/).
```
Guichaguri, line 37: Is there a reason to get back into using full URLs here?

example/service.js
```diff
@@ -32,4 +32,50 @@ module.exports = async function() {
...
+        TrackPlayer.stop();
+        return;
+      }
+
+      // When the event is triggered with ducking set to true,
+      // you should either lower the volume or pause the app.
+      // If it’s a music app, you probably want to lower the volume,
+      // but if it’s a podcast app, you probably want to pause it for a moment.
+      if (ducking) {
+        volumeBeforeDuck = await TrackPlayer.getVolume();
```
Guichaguri, line 52: Manually ducking has been removed in favor of letting the OS do the ducking automatically, the `ducking` parameter has also been removed with that.

docs/Background-Mode.md
```diff
@@ -9,27 +9,29 @@ redirect_from:
 
 # Background Mode
 
+React Native Track Player supports playing audio while your app is in the background on all supported platforms.
+
 ## Android
-The background support works right out of the box, and it will keep playing even after the app is closed. If that is not the desired behavior and you only want to play when the app is open, you can disable it with the `stopWithApp` property in `updateOptions`:
+Background audio playback works right out of the box. By default, the audio will keep playing even after the app is closed. If that is not the desired behavior and you want the audio to stop playing when the app is closed, you can disable it with the `stopWithApp` property in `updateOptions`:
```
curiousdustin, line 15: Can we word this in a way that more clearly states the stopWithApp has to do with when the app is closed/quit, which is different than when it is suspended for multitasking?

curiousdustin:
Thanks. It’s looking great!

puckey:
Thanks for the feedback @curiousdustin & @Guichaguri - will run through it and make some changes

docs/index.md
puckey: Happy to change it – but we are also using full links in the index at the top of README – should I change those into relative links also?

docs/API.md
```diff
@@ -10,169 +10,166 @@ redirect_from:
...
 
-```javascript
-// Adding to the queue
-// You can provide multiple items in one single call
-TrackPlayer.add([track, track2]).then(function() {
-    // The tracks were added
-});
+// Add the tracks to the queue:
+await TrackPlayer.add([track1, track2]);
 ```
```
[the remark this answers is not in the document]
puckey: Yes - do we also need the track durations in there?

docs/Background-Mode.md
puckey: Hmm - would you like to make a proposal for this?

docs/index.md
Guichaguri: Do you mean in the `README.md`? Unfortunately relative links don't work as expected there, but we should use relative links in every `docs/` page where possible

docs/API.md
Guichaguri: I think it was to prevent multiple metadata updates, as the module would show the notification as soon as it starts playing, but it would have to update the notification right after because of the `setOptions` call.

That can be removed or moved into the formal documentation

docs/API.md
Guichaguri: Basically the Android service requires a notification in order to be in the "foreground" state (allowing it to use more system resources without being killed), and that notification is forced to be "on going" (not swipable).

A stop button would be an intuitive and easy to implement solution to allow the user to get rid of the player. I've seen some apps destroying the player when it is closed while paused, which also works pretty well.

docs/Background-Mode.md
Guichaguri: Totally not related to this PR, but I've been thinking about making `stopWithApp` false by default in v2, as it's closest to how iOS behaves and the majority of users also set this to false.
As both iOS and Windows need extra steps for background mode to work, I think it would be more intuitive if Android worked the same way.

docs/API.md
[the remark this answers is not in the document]
Guichaguri: Yeah, there's recurring issues that reports that `getDuration()` returns the wrong value. The answer is that the module streams the track, it doesn't load the full track into memory before playing it, so the duration is pretty much a prediction. Because of that, all streaming services store the duration in their database in order to not have to rely on the duration calculated by the player. There is a `duration` parameter in the track object that will make `getDuration()` return it.

It would be really helpful if you could add that explanation somewhere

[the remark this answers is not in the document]
puckey: Okay. I think the `getDuration()` docs cover your explanation: https://react-native-kit.github.io/react-native-track-player/documentation/#getduration but I will see if I can reword it a bit

docs/API.md
puckey: I opted to remove this.

docs/API.md
puckey: Okay, I extended this note to include your explanation.

puckey:
@Guichaguri Are the linking instructions in `Installation.md` still necessary? As far as I can see, on android linking happens automatically when using the latest version of React Native. On iOS you need to do a `pod install` in the ios directory...

curiousdustin:
@puckey I think it may be good to keep them in there for a bit still. Not everyone is necessarily using 0.60+ yet. With v2, we will start requiring 0.60+. Related is that below 0.60, you didn't necessarily have to use CocoaPods. Some people would avoid using CocoaPods as it was just one more thing that to learn, or go wrong, when you're starting out...

However, I definitely think it needs to be clarified which installation steps are necessary for which RN versions... and if you're using CocoaPods or not.

puckey:
@curiousdustin okay cool - I will see if I can take care of that soon.. lets get this one merged first

puckey:
@Guichaguri are you sure that `remote-ducking` is now working out of the box on android? An initial test has my audio running at the same time as the audio of other apps..

docs/Background-Mode.md
curiousdustin: Here is my stab at it:

> Background audio playback works right out of the box. By default, the audio will continue to play, not only when the app is suspended in the background, but also after the app is closed by the user. If that is not the desired behavior and you want the audio to stop playing when the app is closed, you can disable it with the `stopWithApp` property in `updateOptions`. In this case, the audio will still play while the app is open in the background.

puckey: Perfect - will add it.

curiousdustin:
Also, on iOS, I think we still have to decide when to resume based on `remote-duck`. I think it pauses automatically (iOS does this out of our control).

Perhaps we should leave the remote-duck example out of this PR, are define it better in the formal docs?

curiousdustin:
@puckey Will these changes address everything discussed in https://github.com/react-native-kit/react-native-track-player/issues/505?

Guichaguri:
> @Guichaguri are you sure that `remote-ducking` is now working out of the box on android? An initial test using the example app as my audio running at the same time as the audio of other apps (Youtube in this case)..

Yeah, it works out of the box only for ducking, pausing and stopping the player are still triggered.

Newer Android versions now duck automatically and only triggers the event for pausing or stopping the player. We can choose whether we would like it to trigger a pause event instead of ducking for every interruption, but that's it.
To workaround that, I've decided to remove the `ducking` property from the event, implement auto-ducking for older Android versions and add the `alwaysPauseOnInterruption` option to disable automatic ducking and trigger the event with `paused` set to true instead.

Guichaguri:
Is this ready for merging?

puckey:
Yes go ahead

Guichaguri:
Thank you!

PR #926, opened 2020-05-04
[docs] Clarify casting support in the docs.

curiousdustin:


curiousdustin:
@Guichaguri, I do not have the full background on this module previously supporting casting, and why it was removed and such. I just wanted to clean up references to it in the docs. 

Can you take a look and make sure my changes make sense?

puckey:
We should probably also remove the `Cast-Integration.md` file, which contains outdated instructions.

curiousdustin:
I intentionally left that because @Guichaguri has excluded that file from the docs navigation menu. So, I wasn't sure what his intention with it was.

But I do agree, it is possible to navigate to this file in github browsing view, so it could be confusing to find it. I suggest either removing it, or adding a very clear header in it that states it is not currently supported.

Guichaguri:
I want to add casting back as soon as we merge v2, I've removed it from the module due to the incompatibility and build issues it used to cause in older RN versions, and that's why I started working on an "addon", but I think those are not an issue anymore.

I've never removed because I thought I'd get that back pretty quickly, but it's misleading. Feel free to merge it

curiousdustin:
@Guichaguri , should we remove the `Cast-Integration.md` file as well?

Guichaguri:
Yeah, I can add that back later, there is probably quite a bit I'd like to rewrite anyway

PR #935, opened 2020-05-11
Update hooks

andrekovac:
This PR does two things:

1. Gives the possibility to the `useTrackPlayerProgress` hook to return the correct Player `position` even if it is currently not buffering or playing. The signature of the hook was changed such that it still has the former behavior as a default value. This way it works exactly the same for people how already use it and just adds the possibility to use the hook when e.g. creating a movable Slider.

2. Expose the hooks which are already documented in the Typescript typings file but not yet exposed.

### DIFF
--- a/docs/Documentation.md
+++ b/docs/Documentation.md
@@ -567,14 +567,17 @@ A hook alternative to the [Progress Component](#progresscomponent).
 | bufferedPosition | `number` | The buffered position in seconds |
 | duration         | `number` | The duration in seconds          |
 
-`useTrackPlayerProgress` accepts two arguments:
+`useTrackPlayerProgress` accepts two arguments
 
-1. `interval`: An interval to set the rate (in miliseconds) to poll the track player's progress (default value is `1000` or every second)
-2. `pollTrackPlayerStates`: An optional whitelist of playback states at which the track player's progress is polled.
+| Param                 | Type      | Description                            |
+| --------------------- | --------- | -------------------------------------- |
+| interval              | `number`  | An interval to set the rate (in miliseconds) to poll the track player's progress (default value is `1000` or every second)  |
+| pollTrackPlayerStates | `State[]` | An optional whitelist of playback states at which the track player's progress is polled.  |
 
-  Provide `null` in case you want really poll at the `interval`'s rate, no matter the playback state the player is in.
+More information concerning the `pollTrackPlayerStates` argument:
 
-  By default the hook will only poll while a track is playing or buffering, i.e. the default value for the second argument is:
+- Provide `null` in case you want really poll at the `interval`'s rate, no matter the playback state the player is in.
+- By default the hook will only poll while a track is playing or buffering, i.e. the default value for the second argument is:
 
   ```js
   [
--- a/docs/Documentation.md
+++ b/docs/Documentation.md
@@ -109,7 +109,7 @@ If the player is already initialized, the promise will resolve instantly.
 
 | Param                | Type     | Description   | Default   | Android | iOS | Windows |
 | -------------------- | -------- | ------------- | --------- | :-----: | :-: | :-----: |
-| options              | `object` | The options   | 
+| options              | `object` | The options   |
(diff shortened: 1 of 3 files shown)


curiousdustin:
@andrekovac Could you add the type definition for the add parameter?

@Guichaguri, I'll defer to you on this one.

andrekovac:
@Guichaguri Updated it in this commit: https://github.com/react-native-kit/react-native-track-player/pull/935/commits/9cdcfa8be780870db16ba5b01c38e9a7dc6dd95d

colinricardo:
+1

andrekovac:
I wasted several hours trying to figure out why the position would not update and already began to implement crazy workarounds until I figured out that the `useTrackPlayerProgress` hook does **NOT** return at every interval, but only if the player is in certain playback states.

I now updated the documentation concerning this hook so that it's hopefully more obvious for other.

However, to be honest, the default behavior of the hook should be to return all the time, given by the frequency defined in the `interval` argument. And this restriction on certain playback states should be an optional parameter.

However, I didn't want to introduce a breaking change here. But please consider this for the future. I find the function the way it is now, a bit misleading. Much less so, after this PR is merged. But it might cost some others developers some more headaches in the future.

docs/Documentation.md
```diff
@@ -566,15 +567,30 @@ A hook alternative to the [Progress Component](#progresscomponent).
 | bufferedPosition | `number` | The buffered position in seconds |
 | duration         | `number` | The duration in seconds          |
 
-`useTrackPlayerProgress` accepts an interval to set the rate (in miliseconds) to poll the track player's progress. The default value is `1000` or every second.
+`useTrackPlayerProgress` accepts two arguments:
+
+1. `interval`: An interval to set the rate (in miliseconds) to poll the track player's progress (default value is `1000` or every second)
+2. `pollTrackPlayerStates`: An optional whitelist of playback states at which the track player's progress is polled.
+
+  Provide `null` in case you want really poll at the `interval`'s rate, no matter the playback state the player is in.
```
puckey, line 575: Make it optional instead of asking the user to provide `null`

puckey:
Ah, I understand the reason for the null option now.. ignore my earlier comment.

andrekovac:
@puckey Exactly as you said I'd also recommend to than introduce the breaking change in v2

docs/Documentation.md
```diff
@@ -566,15 +567,30 @@ A hook alternative to the [Progress Component](#progresscomponent).
...
+
+  By default the hook will only poll while a track is playing or buffering, i.e. the default value for the second argument is:
+
+  ```js
+  [
+    TrackPlayer.STATE_PLAYING,
+    TrackPlayer.STATE_BUFFERING,
+  ]
+  ```
+
```
Guichaguri, line 585: Can you format the parameters into a table, just like the other functions?

lib/hooks.js
```diff
@@ -127,34 +128,44 @@ const usePlaybackStateIs = (...states) => {
...
-            TrackPlayer.getDuration()
-        ])
-        setState({ position, bufferedPosition, duration });
+      const [position, bufferedPosition, duration] = await Promise.all([
+        TrackPlayer.getPosition(),
+        TrackPlayer.getBufferedPosition(),
+        TrackPlayer.getDuration(),
+      ]);
+      setState({ position, bufferedPosition, duration });
+    };
```
Guichaguri, line 153: The whole file is indented with 4 spaces, can you keep the same code style?

docs/Documentation.md
andrekovac: I added a commit in which I added them into a table. Because the comment for the second argument is very big, I wrote a note below the table.

lib/hooks.js
andrekovac: Ok, I indented the example.

andrekovac:
@Guichaguri I added two commits to address your code review

curiousdustin:

Issue #617, opened 2019-06-04
[iOS] Can't play track store on Firebase


AylanBoscarino:
# Configuration

Version:
1.1.4

React Native Environment Info:
    System:
      OS: macOS 10.14.5
      CPU: (4) x64 Intel(R) Core(TM) i5-5250U CPU @ 1.60GHz
      Memory: 21.87 MB / 8.00 GB
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 10.16.0 - /usr/local/bin/node
      Yarn: 1.16.0 - /usr/local/bin/yarn
      npm: 6.9.0 - /usr/local/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 12.2, macOS 10.14, tvOS 12.2, watchOS 5.2
      Android SDK:
        API Levels: 28
        Build Tools: 28.0.3
    IDEs:
      Android Studio: 3.4 AI-183.6156.11.34.5522156
      Xcode: 10.2/10E125 - /usr/bin/xcodebuild
    npmPackages:
      react: 16.8.4 => 16.8.4 
      react-native: 0.59.3 => 0.59.8 
    npmGlobalPackages:
      react-native-cli: 2.0.1

# Issue

I'm using the Firebase Storage module to store my audio files. The problem is that Firebase Storage uses %2F to indicate it's internal file structure in the URL and it's being encoded to '/'.

This problem only occurs in iOS, the Android version of the app works as expected.

This is the audio I'm working on: 
https://firebasestorage.googleapis.com/v0/b/hupper-mobile.appspot.com/o/audio_meditation%2Fanxiety%2FTeste_Raffael%20Marfara%CC%81_20%20segundos.mp3?alt=media&token=55aff5c0-1af7-4a8b-b4a2-f450c12f181c

AylanBoscarino:
I found the problem was that the model MediaURL.swift was encoding and decoding the URL it was given. 
I'm not swift programmer  and don't have the time to learn right now so I just commented some lines until it worked.

I forked the library just so I can run it in my project ASAP but I hope to find a proper solution.

```swift
import Foundation

struct MediaURL {
    let value: URL
    let isLocal: Bool
    private let originalObject: Any
    
    init?(object: Any?) {
        guard let object = object else { return nil }
        originalObject = object
        
        if let localObject = object as? [String: Any] {
            let uri = localObject["uri"] as! String
[...]

PR #950, opened 2020-05-20
[iOS] Resolve some issues with provided url formatting.

curiousdustin:
Without these changes the following provided urls were failing on iOS:

```js
url: encodeURI(`file://${RNFS.MainBundlePath}/assets/foo bar.mp3`), // fails to play
url: `file://${RNFS.MainBundlePath}/assets/foo bar.mp3`, // crash on add
url: `file://${RNFS.MainBundlePath}/assets/foo[bar].mp3`, // fails to play
url: require("./assets/foo bar.mp3"), // fails to play
```

With these changes, all of the above work.

(except the last one using `require` with spaces, which still fails in debug mode, but works in release. I feel this is a metro issue)

I am looking for some input on whether this should be changed. I have seen some old discussions about letting the developer handle encoding and such. This change would apply encoding as a last resort.

Here are some related issues:

https://github.com/react-native-kit/react-native-track-player/issues/947 (spaces)
https://github.com/react-native-kit/react-native-track-player/pull/71 (old discussion about taking encoding out)
https://github.com/react-native-kit/react-native-track-player/issues/44 (encoding)
https://github.com/react-native-kit/react-native-track-player/issues/874
https://github.com/react-native-kit/react-native-track-player/issues/794#issuecomment-573353727
https://github.com/react-native-kit/react-native-track-player/issues/794#issuecomment-619060395
https://github.com/react-native-kit/react-native-track-player/issues/167 (spaces, crashing)

---

## Tested with the following values for url

```
require(`./assets/test/foo-bar.mp3`)
require(`./assets/test/foo bar.mp3`) // fails in debug, works in release
require(`./assets/test/foo[bar.mp3`)
require(`./assets/test/fooæbar.mp3`)

file://${RNFS.MainBundlePath}/assets/test/foo-bar.mp3
file://${RNFS.MainBundlePath}/assets/test/foo bar.mp3
file://${RNFS.MainBundlePath}/assets/test/foo[bar.mp3
file://${RNFS.MainBundlePath}/assets/test/fooæbar.mp3
file://${RNFS.MainBundlePath}/assets/test/foo%C3%A6bar.mp3
file://${RNFS.MainBundlePath}/assets/test/foo%C3%A6baræ.mp3 // fails in both debug and release

https://test.domain.com/public/pin/test/foo-bar.mp3
https://test.domain.com/public/pin/test/foo bar.mp3
https://test.domain.com/public/pin/test/foo[bar.mp3
https://test.domain.com/public/pin/test/fooæbar.mp3
https://test.domain.com/public/pin/test/foo%C3%A6bar.mp3
https://test.domain.com/public/pin/test/foo%C3%A6baræ.mp3 // fails in both debug and release

// Test Issue #617
https://firebasestorage.googleapis.com/v0/b/hupper-mobile.appspot.com/o/audio_meditation%2Fanxiety%2FTeste_Raffael%20Marfara%CC%81_20%20segundos.mp3?alt=media&token=55aff5c0-1af7-4a8b-b4a2-f450c12f181c

```

### DIFF
--- a/ios/RNTrackPlayer/Models/MediaURL.swift
+++ b/ios/RNTrackPlayer/Models/MediaURL.swift
@@ -25,7 +25,24 @@ struct MediaURL {
         } else {
             let url = object as! String
             isLocal = url.lowercased().hasPrefix("file://")
-            value = URL(string: url.replacingOccurrences(of: "file://", with: ""))!
+            if isLocal {
+                // When the provided url is local,
+                // remove the file:// prefix,
+                // and attempt to remove any existing encoding.
+                // Both of these can both cause URL(fileURLWithPath: string) to fail
+                var path = url.replacingOccurrences(of: "file://", with: "")
+                path = path.removingPercentEncoding ?? path
+                value = URL(fileURLWithPath: path)
+            } else {
+                // When the provided url is not local, attempt to use it without any modifications
+                if let urlValue = URL(string: url) {
+                    value = urlValue
+                } else {
+                    // If url could not be initialized, try encoding it
+                    print("Warning: The provided url could not be parsed. Attempting to encode url instead.")
+                    value = URL(string: url.addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed)!)!
+                }
+            }
         }
     }
 }
--- a/ios/RNTrackPlayer/Models/MediaURL.swift
+++ b/ios/RNTrackPlayer/Models/MediaURL.swift
@@ -26,7 +26,7 @@ struct MediaURL {
                 url = String(format: "%@.bundle/%@", bundleName, url)
             }
             
-            isLocal = url.lowercased().contains("http") ? false : true
+            isLocal = url.lowercased().hasPrefix("http") ? false : true
             value = RCTConvert.nsurl(url)
         } else {
             let url = object as! String
--- a/ios/RNTrackPlayer/Models/MediaURL.swift
+++ b/ios/RNTrackPlayer/Models/MediaURL.swift
@@ -17,32 +17,21 @@ struct MediaURL {
         guard let object = object else { return nil }
         originalObject = object
         
+        // This is based on logic found in RCTConvert NSURLRequest, 
+        // and uses RCTConvert NSURL to create a valid URL from various formats
         if let localObject = object as? [String: Any] {
-            let uri = localObject["uri"] as! String
-            isLocal = uri.contains("http") ? false : true
-            let encodedURI = uri.addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed)!
-            value = URL(string: encodedURI.replacingOccurrences(of: "file://", with: ""))!
+            var url = localObject["uri"] as? String ?? localObject["url"] as! String
+            
+            if let bundleName = localObject["bundle"] as? String {
+                url = String(format: "%@.bundle/%@", bundleName, url)
+            }
+            
+            isLocal = url.lowercased().contains("http") ? false : true
+            value = RCTConvert.nsurl(url)
         } else {
             let url = object as! String
             isLocal = url.lowercased().hasPrefix("file://")
-            if isLocal {
-                // When the provided url is local,
-                // remove the file:// prefix,
-                // and attempt to remove any existing encoding.
-                // Both of these can both cause URL(fileURLWithPath: string) to fail
-                var path = url.replacingOccurrences(of: "file://", with: "")
-                path = path.removingPercentEncoding ?? path
-                value = URL(fileURLWithPath: path)
-            } else {
-                // When the provided url is not local, attempt to use it without any modifications
-                if let urlValue = URL(string: url) {
-                    value = urlValue
-                } else {
-                    // If url could not be initialized, try encoding it
-                    print("Warning: The provided url could not be parsed. Attempting to encode url instead.")
-                    value = URL(string: url.addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed)!)!
-                }
-            }
+            value = RCTConvert.nsurl(url)
         }
     }
 }
--- a/ios/RNTrackPlayer/Models/Track.swift
+++ b/ios/RNTrackPlayer/Models/Track.swift
@@ -79,7 +79,7 @@ class Track: NSObject, AudioItem, TimePitching, AssetOptionsProviding {
     // MARK: - AudioItem Protocol
     
     func getSourceUrl() -> String {
-        return url.value.absoluteString
+        return url.isLocal ? url.value.path : url.value.absoluteString
     }
     
     func getArtist() -> String? {


puckey:
I attempted to test the `url: require("./assets/foo bar.mp3")` case by importing a local file with spaces but it fails for me:

importing using:
```js
import localTrack from './src/resources/pure with space.m4a';
const track = {
    id: 'local-track',
    url: localTrack,
    title: 'Pure (Demo)',
    artist: 'David Chavez',
    artwork: 'https://i.picsum.photos/id/500/200/200.jpg',
    duration: 28,
  }
```

```
Error: Asset not found: /dev/react-native-track-player/example/src/resources/pure%20with%20space.m4a for platform: ios
    at /dev/react-native-track-player/example/node_modules/metro/src/Assets.js:188:13
    at Generator.next (<anonymous>)
    at asyncGeneratorStep (/dev/react-native-track-player/example/node_modules/metro/src/Assets.js:52:24)
    at _next (/dev/react-native-track-player/example/node_modules/metro/src/Assets.js:72:9)
```

But it looks to me that this error is originating from `react-native` not supporting the requiring of assets with spaces..

curiousdustin:
@puckey, was that a dev or release build, when it failed with spaces?

curiousdustin:
@Guichaguri would be nice to get your input on whether we should make this change. The key point being that this adds a fallback to attempt encoding the provided URL, when iOS fails to create a URL object from the provided URL string.

I had found some previous discussion on this matter: https://github.com/react-native-kit/react-native-track-player/issues/44

Guichaguri:
If I understood the problem correctly:
* Local files need to be decoded in iOS (e.g. `æ.mp3`)
* Remote files need to be encoded in iOS (e.g. `%C3%A6.mp3`)

The problem is that we need to find a common ground for both cases, while still accepting both URLs, is that right?

I can think of three options we can follow:
* Try to get as closest to the official [WHATWG URL spec](https://url.spec.whatwg.org/#example-url-parsing), which makes sense, as all other web APIs are based on that. But that would require the same logic be implemented on Android and iOS
* Use the same logic as other React Native components, such as <Image>. That's already how it is implemented on Android.
* Make that we only accept encoded URLs and add that to the documentation, easier to implement on both sides and less bug prone.

I'm think using the same logic as the <Image> component would make sense and require the least amount of code and documentation.

curiousdustin:
Ok, using the same logic as `<Image>`, _sounds_ straight forward, but after digging around in the RN source code for a half hour, I haven't identified where that logic would be. 🤦 

Do you have any pointers?

Guichaguri:
I've looked for quite a bit, and also didn't found anything for a good half hour too, but finally found this [conversion function](https://github.com/facebook/react-native/blob/ee3994530d4f76695d77c9a21a6b57429bac68aa/React/Base/RCTConvert.m#L141-L157).

At first, I thought it didn't look like it was doing anything to the URL, but then I found [this function](https://github.com/facebook/react-native/blob/ee3994530d4f76695d77c9a21a6b57429bac68aa/React/Base/RCTConvert.m#L79-L126) that deals with encoding for both remote and local URLs

Is there anything we can use from that?

curiousdustin:
Still looking into this, but just have to share this [gem](https://github.com/facebook/react-native/blob/ee3994530d4f76695d77c9a21a6b57429bac68aa/React/Base/RCTJavaScriptLoader.mm#L331):

```
static NSURL *sanitizeURL(NSURL *url)
{
  // Why we do this is lost to time. We probably shouldn't; passing a valid URL is the caller's responsibility not ours.
  return [RCTConvert NSURL:url.absoluteString];
}
```

puckey:
@curiousdustin sorry missed your earlier question – I was testing with a dev build

Guichaguri:
That's great! We should look at some point in unit testing these functions

Does it work with encoded URLs such as `%C3%A6.mp3` (which should point to `æ.mp3`)?

Not that it should work, but I'm also curious about mixing encoded characters with non-encoded ones, like `%C3%A6-æ.mp3` which would be the equivalent of `æ-æ.mp3`

ios/RNTrackPlayer/Models/MediaURL.swift
```diff
@@ -17,32 +17,21 @@ struct MediaURL {
...
-            isLocal = uri.contains("http") ? false : true
-            let encodedURI = uri.addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed)!
-            value = URL(string: encodedURI.replacingOccurrences(of: "file://", with: ""))!
+            var url = localObject["uri"] as? String ?? localObject["url"] as! String
+            
+            if let bundleName = localObject["bundle"] as? String {
+                url = String(format: "%@.bundle/%@", bundleName, url)
+            }
+            
+            isLocal = url.lowercased().contains("http") ? false : true
```
Guichaguri, line 29: I think this would be safer with `.hasPrefix`, as the file path could somehow contain http, such as a random generated filename like this one: `MVHZSTGYNIHTTPSYUJBWVX.mp3`

curiousdustin: I agree, I will change this.

@dcvz added this line, so I was just trying to leave it untouched, but I don't see any notes or comments about why it would be the way it is instead of checking for prefix.

curiousdustin:
Ok, I have made the change you suggested @Guichaguri. I also tested a few more cases.

```
require(`./assets/test/foo-bar.mp3`)
require(`./assets/test/foo bar.mp3`) // fails in debug, works in release
require(`./assets/test/foo[bar.mp3`)
require(`./assets/test/fooæbar.mp3`)

file://${RNFS.MainBundlePath}/assets/test/foo-bar.mp3
file://${RNFS.MainBundlePath}/assets/test/foo bar.mp3
file://${RNFS.MainBundlePath}/assets/test/foo[bar.mp3
file://${RNFS.MainBundlePath}/assets/test/fooæbar.mp3
file://${RNFS.MainBundlePath}/assets/test/foo%C3%A6bar.mp3
file://${RNFS.MainBundlePath}/assets/test/foo%C3%A6baræ.mp3 // fails in both debug and release

https://test.domain.com/public/pin/test/foo-bar.mp3
https://test.domain.com/public/pin/test/foo bar.mp3
https://test.domain.com/public/pin/test/foo[bar.mp3
https://test.domain.com/public/pin/test/fooæbar.mp3
https://test.domain.com/public/pin/test/foo%C3%A6bar.mp3
https://test.domain.com/public/pin/test/foo%C3%A6baræ.mp3 // fails in both debug and release

// Test Issue #617
https://firebasestorage.googleapis.com/v0/b/hupper-mobile.appspot.com/o/audio_meditation%2Fanxiety%2FTeste_Raffael%20Marfara%CC%81_20%20segundos.mp3?alt=media&token=55aff5c0-1af7-4a8b-b4a2-f450c12f181c
```

As you probably guessed, mixing encoding and non-encoding in the filename doesn't work. It doesn't compile if you use require, and it fails to load if using paths/urls.

I also tested the case found in #617, which was one of the causes for waffling back and forth between whether to encode or not. This case works now because the `RCTConvert.nsurl(url)` utility method only tries encoding if the initial url fails to parse into a URL object.

curiousdustin:
@Guichaguri I only tested this on iOS. Should we take this opportunity to verify Android behavior is the same?

What is the easiest way to include audio and image files in the android build that can be accessed with `file://` paths? I haven't needed to do this in my past projects.

Guichaguri:
I think using RNFS is the easiest way to get files in a directory that the app has permission to access.

All the file paths and URLs seem to work on Android, even mixing encoded and non-encoded characters, I haven't had time to test the require though, but I guess we're stuck with RN limitations anyway.

curiousdustin:
Oh, so there isn't just an easy way to include files in a given folder when making a build, and then access them with RNFS? on iOS you just add a given folder to the Xcode project, and those assets and available in the app via the file:// syntax.

Guichaguri:
You're meant to use the app bundle instead, which is different from the system's file system.

TalhaArshad (commented):
It's working

curiousdustin:

PR #952, opened 2020-05-25
Update typescript definitions

andrekovac:
Update type definitions of STATE and PITCH types to be more accurate.

Also run prettier on the index.d.ts file.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,9 +1,8 @@
-import { Component } from 'react';
+import { Component } from "react";
 
 export = RNTrackPlayer;
 
 declare namespace RNTrackPlayer {
-
   export type EventType =
     | "playback-state"
     | "playback-error"
@@ -27,11 +26,7 @@ declare namespace RNTrackPlayer {
     | "remote-dislike"
     | "remote-bookmark";
 
-  export type TrackType =
-    | "default"
-    | "dash"
-    | "hls"
-    | "smoothstreaming";
+  export type TrackType = "default" | "dash" | "hls" | "smoothstreaming";
 
   type ResourceObject = any;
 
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -163,19 +203,57 @@ declare namespace RNTrackPlayer {
...
+  export const STATE_ANDROID_SKIPPING_TO_QUEUE_ITEM: 11;
+
+  // iOS states
+  export const STATE_IOS_NONE: "idle";
+  export const STATE_IOS_PLAYING: "playing";
+  export const STATE_IOS_PAUSED: "paused";
+  export const STATE_IOS_STOPPED: "idle";
+  export const STATE_IOS_CONNECTING: undefined;
+  export const STATE_IOS_BUFFERING: "loading";
+  export const STATE_IOS_READY: "ready";
```
Guichaguri, line 256: These constants are defined by the native code just to prevent them being hardcoded here, I think constants should be their own type, not even assuming that they are going to be numbers or strings

index.d.ts
```diff
@@ -200,10 +278,16 @@ declare namespace RNTrackPlayer {
...
-  
+  export const PITCH_ALGORITHM_LINEAR: typeof PITCH_ALGORITHM_LINEAR_IOS | typeof PITCH_ALGORITHM_ANDROID;
+  export const PITCH_ALGORITHM_MUSIC: typeof PITCH_ALGORITHM_MUSIC_IOS | typeof PITCH_ALGORITHM_ANDROID;
+  export const PITCH_ALGORITHM_VOICE: typeof PITCH_ALGORITHM_VOICE_IOS | typeof PITCH_ALGORITHM_ANDROID;
+
+  export const PITCH_ALGORITHM_LINEAR_IOS: "linear";
+  export const PITCH_ALGORITHM_MUSIC_IOS: "music";
+  export const PITCH_ALGORITHM_VOICE_IOS: "voice";
+
+  export const PITCH_ALGORITHM_ANDROID: undefined;
```
Guichaguri, line 289: These other string constants are only defined in JS, which shouldn't be different between platforms, even if they are not supported. I feel like this is not the proper way to represent constants that are not supported in a specific platform.

dcvz:
thanks for this @andrekovac -- the main branch (v2) has been rewritten in Typescript. I'll close this for now.

PR #957, opened 2020-05-28
[ios] deduce togglePlayPause from play and pause capabilities

puckey:
This adds support for toggling playback for headphones. Started at https://github.com/react-native-kit/react-native-track-player/pull/531, but deducing togglePlayPause from play and pause capabilities instead of adding it as a separate one. Tested using wired apple headphones.

### DIFF
--- a/ios/RNTrackPlayer/Models/Capabilities.swift
+++ b/ios/RNTrackPlayer/Models/Capabilities.swift
@@ -9,7 +9,7 @@
 import Foundation
 
 enum Capability: String {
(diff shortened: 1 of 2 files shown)


puckey:
I was testing my implementation with airpods and noticed I was unable to stop / play.. After looking at Apple's sample code, I figured out they apply the `play` / `pause` together with `togglePlayPlause` - whis gets airpods support working.

curiousdustin:
Looks good to me. Thanks @puckey

curiousdustin:

PR #960, opened 2020-06-01
[ios] fix updating of remote commands

puckey:
I was looking into the following issue: https://github.com/react-native-kit/react-native-track-player/issues/921 where lock screen controls were ending up in a disabled state after calling `TrackPlayer.updateOptions`.

I compared the way `SwiftAudio` does enabling of commands with the way the Apple `NowPlaying` example code does it. Apple disables / reenables commands one by one, while SwiftAudio first disables all commands and then enables the ones that need to be enabled.

This also seems to solve the flashing of controls when calling `TrackPlayer.updateOptions`.

### DIFF
--- a/ios/RNTrackPlayer/Vendor/SwiftAudio/Classes/RemoteCommandController/RemoteCommandController.swift
+++ b/ios/RNTrackPlayer/Vendor/SwiftAudio/Classes/RemoteCommandController/RemoteCommandController.swift
@@ -30,9 +30,11 @@ public class RemoteCommandController {
     }
     
     internal func enable(commands: [RemoteCommand]) {
-        self.disable(commands: RemoteCommand.all())
-        commands.forEach { (command) in
-            self.enable(command: command)
+        RemoteCommand.all().forEach { (command) in
+            self.disable(command: command)
+            if commands.contains(command) {
+                self.enable(command: command);
+            }
         }
     }
     
--- a/ios/RNTrackPlayer/Vendor/SwiftAudio/Classes/RemoteCommandController/RemoteCommandController.swift
+++ b/ios/RNTrackPlayer/Vendor/SwiftAudio/Classes/RemoteCommandController/RemoteCommandController.swift
@@ -32,8 +32,8 @@ public class RemoteCommandController {
     internal func enable(commands: [RemoteCommand]) {
         RemoteCommand.all().forEach { (command) in
             self.disable(command: command)
-            if commands.contains(command) {
-                self.enable(command: command);
+            if let newCommand = commands.first(where: {$0 == command}) {
+                self.enable(command: newCommand);
             }
         }
     }
--- a/ios/RNTrackPlayer/Vendor/SwiftAudio/Classes/RemoteCommandController/RemoteCommand.swift
+++ b/ios/RNTrackPlayer/Vendor/SwiftAudio/Classes/RemoteCommandController/RemoteCommand.swift
@@ -103,7 +103,7 @@ public struct FeedbackCommand: RemoteCommandProtocol {
     }
 }
 
-public enum RemoteCommand {
+public enum RemoteCommand : Equatable {
 
     case play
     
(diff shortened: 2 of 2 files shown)


ios/RNTrackPlayer/Vendor/SwiftAudio/Classes/RemoteCommandController/RemoteCommandController.swift
```diff
@@ -30,9 +30,11 @@ public class RemoteCommandController {
     }
     
     internal func enable(commands: [RemoteCommand]) {
-        self.disable(commands: RemoteCommand.all())
-        commands.forEach { (command) in
-            self.enable(command: command)
+        RemoteCommand.all().forEach { (command) in
```
gaodeng, line 33:
```swift

.skipForward
.skipBackward
.like
.dislike
.bookmark

commands.contains  with these enumeration values may not get the desired results
```

puckey: Thanks – I will test it with these values.

puckey: Yeah that didn't work. I added a commit which should fix these cases. A little verbose, but it does the job.

curiousdustin:
@puckey, unfortunately this doesn't appear to fix the issues for me.

I still get weird behavior with the following test code:

```ts
  TrackPlayer.addEventListener(TrackPlayerEvents.REMOTE_NEXT, async (data) => {
    //take the default options used at initial setup, and change the capabilities
    const options = trackPlayerOptions;
    options.capabilities = [
      TrackPlayer.CAPABILITY_PLAY,
      TrackPlayer.CAPABILITY_PAUSE,
      // TrackPlayer.CAPABILITY_SKIP_TO_NEXT,
      TrackPlayer.CAPABILITY_SKIP_TO_PREVIOUS,
      TrackPlayer.CAPABILITY_SEEK_TO,
    ];

    await TrackPlayer.skipToNext();
    await TrackPlayer.updateOptions(options);
  });
```

Testing in debug mode results in various random behavior. Either some controls will be disabled when they shouldn't be, things like the title and image will disappear, or even sometimes, the player stops playing instead of playing the expected track.

Testing in release mode is slightly better. I don't see the cases where the image disappears, or the player actually stops, but the controls are occasionally still not correctly configured.

curiousdustin:
Also just to re-iterate points made in [other threads](https://github.com/react-native-kit/react-native-track-player/issues/921#issuecomment-629280272), swapping the order of operations here resolves the issue.

```ts
    await TrackPlayer.updateOptions(options);
    await TrackPlayer.skipToNext();
```

Sure would be great if we could resolve the need for this though.

gaodeng:
> Also just to re-iterate points made in [other threads](https://github.com/react-native-kit/react-native-track-player/issues/921#issuecomment-629280272), swapping the order of operations here resolves the issue.
> 
> ```ts
>     await TrackPlayer.updateOptions(options);
>     await TrackPlayer.skipToNext();
> ```
> 
> Sure would be great if we could resolve the need for this though.

Put these remote command  enable & disable tasks on the main thread to execute, it should solve the problem you said

ios/RNTrackPlayer/Vendor/SwiftAudio/Classes/RemoteCommandController/RemoteCommandController.swift
```diff
@@ -30,9 +30,11 @@ public class RemoteCommandController {
...
+            self.disable(command: command)
```
gaodeng, line 34: Can this line  be placed in the else branch, otherwise, the control buttons will still flicker

puckey:
@gaodeng can you give this pr a try instead: https://github.com/react-native-kit/react-native-track-player/pull/962

bohdan-vasylenko:
> Also just to re-iterate points made in [other threads](https://github.com/react-native-kit/react-native-track-player/issues/921#issuecomment-629280272), swapping the order of operations here resolves the issue.
> 
> ```ts
>     await TrackPlayer.updateOptions(options);
>     await TrackPlayer.skipToNext();
> ```
> 
> Sure would be great if we could resolve the need for this though.

I've spent 6h, struggling with my app on IOS 13.x. On track paused (sometimes also on track change) in background, widget switched to default Apple Music app. The reason was that i've put updating options function like below
```js 
TrackPlayer.addEventListener('remote-play', async () => {
    await Player.controls.play();
    await updateCapabilities();
  });
```
after commenting out 
```js
// await updateCapabilities();
```
 all problems with background have gone completely. 
@curiousdustin thank you so much!

jspizziri:
I'm closing this PR as v1 is no longer supported and this has been inactive for quite some time. Please let me know if it should be reopened.

PR #962, opened 2020-06-02
[ios]: run TrackPlayer.updateOptions from main queue

puckey:


ios/RNTrackPlayer/RNTrackPlayer.swift
```diff
@@ -263,24 +263,24 @@ public class RNTrackPlayer: RCTEventEmitter {
...
+            let capabilities = capabilitiesStr.compactMap { Capability(rawValue: $0) }
+            
+            let remoteCommands = capabilities.map { capability in
+                capability.mapToPlayerCommand(jumpInterval: options["jumpInterval"] as? NSNumber,
+                                              likeOptions: options["likeOptions"] as? [String: Any],
+                                              dislikeOptions: options["dislikeOptions"] as? [String: Any],
+                                              bookmarkOptions: options["bookmarkOptions"] as? [String: Any])
+            }
+            self.player.enableRemoteCommands(remoteCommands)
+
```
gaodeng, line 281:
```swift
self.player.remoteCommands = remoteCommands
```
The remoteCommands variable also needs to be set, otherwise, it may be reset by self.enableRemoteCommands(remoteCommands) here  https://github.com/react-native-kit/react-native-track-player/blob/2392845c0ea3663da891835a48a9ea3ded3fa01b/ios/RNTrackPlayer/Vendor/SwiftAudio/Classes/AudioPlayer.swift#L226

gaodeng:
There are some places inside SwiftAudio that call enable disable remote command in the background thread ，which also needs to be fixed

ios/RNTrackPlayer/RNTrackPlayer.swift
puckey: That functionality is disabled in this overridden function: https://github.com/react-native-kit/react-native-track-player/blob/2392845c0ea3663da891835a48a9ea3ded3fa01b/ios/RNTrackPlayer/RNTrackPlayerAudioPlayer.swift#L119

puckey:
@gaodeng can you point to one of these places?

ios/RNTrackPlayer/RNTrackPlayer.swift
gaodeng: > That functionality is disabled in this overridden function:
> 
> https://github.com/react-native-kit/react-native-track-player/blob/2392845c0ea3663da891835a48a9ea3ded3fa01b/ios/RNTrackPlayer/RNTrackPlayerAudioPlayer.swift#L119

You are right, thank you for pointing this out

curiousdustin:
I think there might appear to be, but with our modifications, they don't actually end up doing anything. For instance, `enableRemoteCommands(forItem: item)` is called in `AudioPlayer.load`, but it will have no effect because of our overridden behavior of `enableRemoteCommands`.

However, if this main queue idea ends up being the fix, perhaps we should get rid of this override, and instead modify SwiftAudio to do things on the proper thread.

gaodeng:
> @gaodeng can you point to one of these places?



> @gaodeng can you point to one of these places?

I haven’t seen the overload in RNTrackPlayerAudioPlayer before, after you point it out, I looked at it again, there should be no such problem

gaodeng:
Thank you very much for your efforts. I really look forward to the merging of these two PR https://github.com/react-native-kit/react-native-track-player/pull/962 https://github.com/react-native-kit/react-native-track-player/pull/960 , which should completely solve the problem that the iOS control buttons are randomly disabled.

puckey:
hopefully #962 should be enough on its own

curiousdustin:
@puckey, I ran the same test as before and this seems solid! 🥇 

What are you thoughts on removing the override I added as a partial fix for this issue? It was basically meant to avoid resetting the controls on each track change. Maybe this is still best, because we don't support the track based control changes that SwiftAudio is looking for anyway...

Thoughts?

gaodeng:
What do you think about removing the dependency on swiftaudio? This project has been inactive for almost half a year

curiousdustin:
@gaodeng, are you suggesting removing it entirely and creating our own implementation of an audio player with queue and such?

gaodeng:
Yes. But I feel that we should implement queue with js.
Native code only implements playback and preload, other APIs are implemented with js. This can simplify a lot of work and will be more flexible

curiousdustin:
That deserves consideration. I would suggest opening a new issue with your suggestion for implementing that.

puckey:
@curiousdustin happy to hear it also works for you! I think the override can stay in place, since we do not support track-based changing of controls. Please feel free to merge if you agree.

curiousdustin:
I agree, leave the override in place for now.

I think we should wait for confirmation from others, then merge this.

mcdit:
Hi guys, any chance that this bugfix will tickle down to version 1 (eg 1.2.4. or something) in the near future ?

Before we started using track player we had our own implementation of a player with queue handling in JS. It was unreliable. We used it for audiobooks and after a while listening to it, iOS would simply put the js engine to sleep and prevented it from feeding the player new tracks. The  background stuff on iOS is like a black box and we did not know what caused iOS to do this.

Guichaguri:
> Yes. But I feel that we should implement queue with js.
> Native code only implements playback and preload, other APIs are implemented with js. This can simplify a lot of work and will be more flexible

A lot of people have asked this before, we don't do that because of gapless playback, which ends up in two cases:
* Smooth playback, basically the player can start loading and buffering the next track before the current one ends.
* As already pointed out by @mcdit, the player doesn't stop playing for a second so the OS will not kill it while in background.

If the queue is fully managed by JS, the native code will not be able to know what is next, hence not being able to do gapless playback. We want to give a little more control to the queue in v2 (see #688), but still controlled by the native code.

gaodeng:
> > Yes. But I feel that we should implement queue with js.
> > Native code only implements playback and preload, other APIs are implemented with js. This can simplify a lot of work and will be more flexible
> 
> A lot of people have asked this before, we don't do that because of gapless playback, which ends up in two cases:
> 
> * Smooth playback, basically the player can start loading and buffering the next track before the current one ends.
> * As already pointed out by @mcdit, the player doesn't stop playing for a second so the OS will not kill it while in background.
> 
> If the queue is fully managed by JS, the native code will not be able to know what is next, hence not being able to do gapless playback. We want to give a little more control to the queue in v2 (see #688), but still controlled by the native code.

Gapless playback can be done by adding a preload api to the native code. When all the current track data is loaded, emit a loaded event. The js code decides whether to call the preload api to preload the next track data

dcvz:
@puckey what problem was this solving and is it still needed in v2?

jspizziri:
I'm closing this due to inactivity and the fact that v1 is no longer supported.

Issue #970, opened 2020-06-11
[iOS] Buffer is not working as expected


robinalexandre:
**Describe the bug**
In RNTrackPlayer.swift I added in the setupPlayer method: 
`
let minBuffer: TimeInterval = config["minBuffer"] as? TimeInterval ?? 15
player.bufferDuration = minBuffer
`

`player.bufferDuration` is then passed to `AVPlayerItem -> preferredForwardBufferDuration` which should control the buffer size. However, it seems like the buffer is controlled automatically and I can't do anything about it.

**To Reproduce**
Add in RNTrackPlayer.swift in the setupPlayer method:
`
let minBuffer: TimeInterval = config["minBuffer"] as? TimeInterval ?? 15
player.bufferDuration = minBuffer
`

Add then the minBuffer options to the TrackPlayer.setupPlayer method.

**Environment (please complete the following information):**
react-native info:
`
System:
    OS: macOS 10.15.5
    CPU: (8) x64 Intel(R) Core(TM) i7-6700HQ CPU @ 2.60GHz
    Memory: 1.78 GB / 16.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 12.12.0 - /usr/local/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.13.1 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  Managers:
    CocoaPods: 1.8.4 - /usr/local/bin/pod
  SDKs:
    iOS SDK:
      Platforms: iOS 13.5, DriverKit 19.0, macOS 10.15, tvOS 13.4, watchOS 6.2
    Android SDK: Not Found
  IDEs:
    Android Studio: 3.6 AI-192.7142.36.36.6241897
    Xcode: 11.5/11E608c - /usr/bin/xcodebuild
  Languages:
    Java: 13.0.1 - /usr/bin/javac
    Python: 3.7.3 - /Users/alexandrerobin/anaconda3/bin/python
  npmPackages:
    @react-native-community/cli: ^4.5.1 => 4.5.1 
    react: 16.13.1 => 16.13.1 
    react-native: 0.63.0-rc.1 => 0.63.0-rc.1 
`
react-native-track-player: V2 branch
Tested on real device using iOS 13.5

**Note**
[...]

Issue #440, opened 2019-02-11
[iOS] Buffering not starting


jakobereksen:
With the iOS Player pre-1.1.0 Buffering would Start automatically. With the Versions >= 1.1.0 Buffering does not start at all under iOS but works as expected under Android.

I am using these player options: 

`minBuffer: 100,
  maxBuffer: 10000,
  playBuffer: 101,
  maxCacheSize: 100000`

ntk860:
what do you mean by "buffering not starting" ? When you add the song to the queue, buffering should start right after you call play().

jakobereksen:
i will try to explain it better. The behaviour I wanted to achieve and had success doing so on iOS pre 1.1.0 was to buffer the whole track before playing it. To do so I started the track and paused it directly after, waited until it was fully buffered and resumed playing.

With 1.1.0 this did not work for me anymore. Buffering progress stayed at 0, it did not matter whether the track was playing or not. Weirdly enough, while stopping to work for iOS, 1.1.0 made android work for me.

hrag:
I'm having trouble getting an answer to this, but is buffering even supported on iOS??

We can't set custom buffering options (options.minBuffer, options.maxBuffer etc) for iOS so it raises questions. Also in the platform support matrix it says caching isn't supported.

Maybe the documentation could be a bit more clear around this? I feel like the shortcomings in iOS are not all that well explained.

@Guichaguri ?

samuelstevens9:
I think the playing issue is related to #616 

> unable to play any streams until I set the undocumented waitForBuffer parameter to true in TrackPlayer.setupPlayer:
```
TrackPlayer.setupPlayer({
    minBuffer: 30, // buffer 30 seconds
    waitForBuffer: true
  });
```

To get basic buffer support on iOS add the following to RNTrackPlayer.swift after `setupInterruptionHandling();`
```
        // minBuffer iOS support
        let minBuffer: TimeInterval = config["minBuffer"] as? TimeInterval ?? 15
        player.bufferDuration = minBuffer
```

iOS AudioPlayer Documentation re: [preferredForwardBufferDuration](https://developer.apple.com/documentation/avfoundation/avplayeritem/1643630-preferredforwardbufferduration)

Hope this helps.

PR #973, opened 2020-06-16
[iOS] Pass minBuffer as the buffer size.

knyar:
This has been suggested in a few issues (#440, #970) but it seems that nobody has actually submitted a change.

### DIFF
--- a/docs/Documentation.md
+++ b/docs/Documentation.md
@@ -110,7 +110,7 @@ If the player is already initialized, the promise will resolve instantly.
 | Param                | Type     | Description   | Default   | Android | iOS | Windows |
 | -------------------- | -------- | ------------- | --------- | :-----: | :-: | :-----: |
 | options              | `object` | The options   |
-| options.minBuffer    | `number` | Minimum time in seconds that needs to be buffered | 15 | ✓ | ✗ | ✗ |
+| options.minBuffer    | `number` | Minimum time in seconds that needs to be buffered | 15 | ✓ | ✓ | ✗ |
 | options.maxBuffer    | `number` | Maximum time in seconds that needs to be buffered | 50 | ✓ | ✗ | ✗ |
 | options.playBuffer   | `number` | Minimum time in seconds that needs to be buffered to start playing | 2.5 | ✓ | ✗ | ✗ |
 | options.backBuffer   | `number` | Time in seconds that should be kept in the buffer behind the current playhead time. | 0 | ✓ | ✗ | ✗ |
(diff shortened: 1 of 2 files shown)


robinalexandre:
Did you try your changes ? 
Is it actually working if you set 15 sec on a long mp3 ?

Because for me, as I said in #970, it's not working.

knyar:
I have confirmed that it gets set as the [preferredForwardBufferDuration](https://developer.apple.com/documentation/avfoundation/avplayeritem/1643630-preferredforwardbufferduration) property when a new [AVPlayerItem](https://developer.apple.com/documentation/avfoundation/avplayeritem) is created.

robinalexandre:
I agree that it is set as the `preferredForwardBufferDuration` but the buffer which is then load does not correspond to the value. The question is: where does this problem come from ?

knyar:
Just to confirm: I am seeing this change take effect and restrict the maximum size of the buffer for an `AVPlayerItem` that contains an HLS stream.

@robinalexandre, perhaps in your specific use case `preferredForwardBufferDuration` does not do what you expect it to do? I don't think this is something this library can really fix.

robinalexandre:
I am using a default stream (mp3).
I expect the buffer to act the same for HSL or any over stream.

Could you please try with a mp3 and tell me your results to see if it comes from my side ?

puckey:
Thanks for your PR!

According to the docs you referred to: "If set to `0`, the player will choose an appropriate level of buffering for most use cases."

In which case, would it not make more sense to default to `0` if no minBuffer is passed by the user?

knyar:
`0` would be a reasonable default, however:
- at least for HLS streams `0` seem to mean that the player will keep buffering audio while it has available memory. This will be a big behaviour change compared to the previous buffer size (1 second).
- [Documentation.md](https://github.com/react-native-kit/react-native-track-player/blob/dev/docs/Documentation.md) describes the default to be 15 seconds. This is why I've decided to use this value.

I am happy to change the default to whatever value you think makes the most sense.

ainformatico:
Hey guys! We actually forked and made the same change some time ago. Wanted to open a PR and saw this one :)
Any idea in which version we should expect this to be merged?

Guichaguri:
> `0` would be a reasonable default, however:
> 
> * at least for HLS streams `0` seem to mean that the player will keep buffering audio while it has available memory. This will be a big behaviour change compared to the previous buffer size (1 second).
> * [Documentation.md](https://github.com/react-native-kit/react-native-track-player/blob/dev/docs/Documentation.md) describes the default to be 15 seconds. This is why I've decided to use this value.
> 
> I am happy to change the default to whatever value you think makes the most sense.

I think we should leave the default on iOS as `0`, just like 15 seconds is the default value on Android. These values also impact in memory usage, the default ones have been tested by the Android and iOS teams, which they know works well for their system.

knyar:
I've changed the default to 0.

lode:
I also would love to see this change in. Is there anything blocking releasing this I could help with?

BjoernRave:
same here, would love to see this get merged :)

dcvz:
thanks for this @knyar ! I have made these changes against the main branch (v2) #f16936a4be2c1acef2c930509138e3777544915a

PR #976, opened 2020-06-23
[iOS] Add AirPlay 2 Support

leinelissen:
Hi there guys,

I was using react-native-track-player, and found its Airplay support was already mostly there! Yet, I ran into two hiccups:
1. When setting AirPlay output, the audio volume is disabled on the device
2. The Apple TV to which I output shows it as if it's a video stream, with metadata missing

I consulted the internet, and these two problems seem relatively simple to fix (see [here](https://stackoverflow.com/questions/12221787/how-to-keep-the-slider-of-mpvolumeview-visible-when-connecting-to-apple-tv/14743134) for instance), by setting `allowsExternalPlayback` to false on `AVPlayer`. This (counter-intuitively) sets the AirPlay stream to audio, rather than video, opening up volume controls and settings the proper metadata when streaming to the Apple TV.

Additionally, I made an effort to [set up AirPlay 2](https://developer.apple.com/documentation/avfoundation/airplay_2/getting_airplay_2_into_your_app) properly for this application. Fortunately, most of the work is already done, with the docs only recommending to set a RouteSharingPolicy, which I have implemented. Unfortunately, I have no means if AirPlay 2 is actually working, but I these settings do no harm regardless.

This pull request at the very least improves QOL for those who wish to use react-native-track-player in combination with AirPlay, and I would recommend it be merged. 

I am here for any additional questions.

Cheers,
Lei

### DIFF
--- a/ios/RNTrackPlayer/RNTrackPlayer.swift
+++ b/ios/RNTrackPlayer/RNTrackPlayer.swift
@@ -170,8 +170,15 @@ public class RNTrackPlayer: RCTEventEmitter {
                 sessionCategoryMode = mappedCategoryMode.mapConfigToAVAudioSessionCategoryMode()
         }
         
-        try? AVAudioSession.sharedInstance().setCategory(sessionCategory, mode: sessionCategoryMode, options: sessionCategoryOptions)
-        
+        // Progressively opt into AVAudioSession policies for background audio
+        // and AirPlay 2. 
+        if #available(iOS 13.0, *) {
+            try? AVAudioSession.sharedInstance().setCategory(sessionCategory, mode: sessionCategoryMode, policy: .longFormAudio, options: sessionCategoryOptions)
+        } else if #available(iOS 11.0, *) {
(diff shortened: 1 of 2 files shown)


leinelissen:
Before  | After
------
 |

puckey:
Thanks for your PR! What would we need to do to test Airplay 2?

leinelissen:
@puckey I am still trying to find a programmatic way to do this. However, you could also try and stream to multiple speakers simultaneously. If this works, it proves Airplay 2 is used. I only have a single Airplay 2-capable device though 😅

puckey:
Okay, in that case Airplay 2 support is already there – I have multiple Homepod / Sonos speakers and have tested RNTP on them.. But that does not take away that the way you have set it up is more correct. I will test this on an Apple TV and merge if everything works well.

leinelissen:
Perfect, thank you!

leinelissen:
Are we good to merge? I'd love to switch back to this package for my projects 😅

devpascoe:
May we have this merged?

curiousdustin:
@puckey did you ever get a chance to check this?

curiousdustin:

PR #982, opened 2020-06-26
Option to unduck on pause

mrvisser:
Hey, thanks for this great library. It got me 99% to where I needed to be, except I need to be able to "unduck" background audio sometimes when music is paused or stopped.

I've made that an option for pause, reset and stop in this PR and implemented it for iOS.

Is this something that you're interested in supporting for other mobile devices as well? Maybe this is already implemented somewhere in `dev`, if so don't hesitate to close this in favour of a later release.

This is the first native code I've written, so it's probably pretty rough around the edges :)

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,7 @@
+{
+  "editor.formatOnSave": false,
+  "editor.codeActionsOnSave": {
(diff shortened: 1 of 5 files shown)


mrvisser:
Closing due to inactivity. Probably better ways to represent in API.

PR #992, opened 2020-07-07
[iOS] Update session category without app restart

bang9:


mcmillan:
This would be really helpful for us!

PR #1018, opened 2020-08-12
Fix: EventHandler first arg type

fredericoamsb:


dcvz:
Thanks for this @fredericoamsb! I think this is no longer necessary on v2.

PR #1054, opened 2020-10-22
support windows vnext

ryfow:
Windows support is out of date. This gets things mostly working with 0.63.5 but is waiting on at least
  https://github.com/microsoft/react-native-windows/pull/6285 so that events get passed to JS correctly.

There's a little work to do to re-enable merging track metdata but I thought I'd get the PR started so someone that knows what they're doing can start to look at it.


windows/RNTrackPlayer.sln
```diff
@@ -1,89 +1,289 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.26730.16
+# Visual Studio Version 16
```
ryfow, line 2: I'm frankly not sure about all these changes in the .sln file. It'd be good for someone that knows what they're doing to look at it.

windows/RNTrackPlayer/Logic/MediaManager.cs
```diff
@@ -1,28 +1,23 @@
-using Newtonsoft.Json.Linq;
-using ReactNative.Bridge;
-using ReactNative.Modules.Core;
+using Microsoft.ReactNative.Managed;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
```
asklar, line 3: i'm curious why do you need this objectmodel using?

asklar:
thanks for the contribution @ryfow ! Adding  @dannyvv to take a look at the C# module stuff

windows/RNTrackPlayer/Logic/MediaManager.cs
ryfow: I'm not sure how they got in there. Just removed them. Thanks.

package.json
```diff
@@ -41,10 +41,16 @@
...
-  },
+  "scripts": {},
   "peerDependencies": {
     "react": ">=16.8.6",
     "react-native": ">=0.60.0-rc.2"
+  },
+  "dependencies": {
+    "react": "16.13.1",
+    "react-native": "0.63.2"
+  },
```
NickGerleman, line 52: A community module shouldn't declare these dependencies directly, These should be devDependencies instead.

ryfow: Thanks!

dcvz:
closing this now as we've merged in #1159 #1180 and #1204 to go into the `v2` release.

PR #1073, opened 2020-11-15
Fix(progress): progress reflects previous track

Gregoirevda:
When `needsPoll` is `true` based on a new PLAYER_STATE, you'd expect to have the progress **directly** and then after each `interval`.
This adds the **directly**


curiousdustin (approved):

curiousdustin:

Issue #978, opened 2020-06-25
Track immediately pauses on iOS after calling TrackPlayer.play()


sgerendasy:
**Describe the bug**
I am using RNTP to stream user-recorded audio that is stored in Azure blob storage. After a user records their audio, I upload the audio via `uploadAudio()` and then setup the player and add a track via `loadPlayer()`. The audio plays just fine. However, if the user deletes the recording and then records a new track (which calls `uploadAudio()` again when finished), I can never get the new audio track to play. When I log the player state, I see it immediately goes from 'playing' to 'paused'.

I'm using react-native-community/react-native-audio-toolkit to record the audio. I'm using RNTP for its background capabilities, plus I'm already using RNTP in another area of the same app.

**Environment**
System:
    OS: macOS 10.15.5
    CPU: (8) x64 Intel(R) Core(TM) i7-4850HQ CPU @ 2.30GHz
    Memory: 212.79 MB / 16.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 12.18.0 - /usr/local/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.14.4 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  SDKs:
    iOS SDK:
      Platforms: iOS 13.5, DriverKit 19.0, macOS 10.15, tvOS 13.4, watchOS 6.2
  IDEs:
    Android Studio: 3.5 AI-191.8026.42.35.5900203
    Xcode: 11.5/11E608c - /usr/bin/xcodebuild
  npmPackages:
    react: 16.8.6 => 16.8.6 
    react-native: 0.60.5 => 0.60.5

I'm using react-native-track-player version 1.2.3
I am testing on a real iOS device, iPhone SE 2020.

**Code (simplified)**

    async uploadAudio(filePath){
        const audioURL = [URL OF AUDIO FILE IN BLOB STORAGE]
        await RNFetchBlob.fetch('PUT', audioURL, 
        {
            'x-ms-blob-type': 'BlockBlob',
            'content-type': 'application/octet-stream'
        }, RNFetchBlob.wrap(filePath))
        .then(async (res) => {
            await this._loadPlayer(audioURL);
        });
    }
  
    async loadPlayer(audioURL){
        await TrackPlayer.destroy();
        await TrackPlayer.setupPlayer().then(async () => {
            await TrackPlayer.updateOptions({
[...]

Issue #616, opened 2019-06-04
[ios] playing streams does not work without undocumented waitForBuffer parameter


puckey:
After releasing the Android version of my app, I went over to iOS mode. I was unable to play any streams until I set the undocumented `waitForBuffer` parameter to `true` in `TrackPlayer.setupPlayer`:

```
  const buffer = 0.5;
  await TrackPlayer.setupPlayer({
    playBuffer: buffer,
    minBuffer: buffer * 2,
    maxBuffer: buffer * 2,
    waitForBuffer: true
  });
```

If I do not set this parameter, the stream plays almost directly and then stops again.

maartenvandillen:
On my device it did not play at all without the waitForBuffer parameter. Android is working fine but I had never heard any audio on iOS until I added that.

sebacaccaro:
> On my device it did not play at all without the waitForBuffer parameter. Android is working fine but I had never heard any audio on iOS until I added that.

Same here

puckey:
I was about to add this parameter to the docs, but reading the [docs](https://developer.apple.com/documentation/avfoundation/avplayer/1643482-automaticallywaitstominimizestal) of the underlying `automaticallyWaitsToMinimizeStalling ` option leads me to think we should just default it to true and not have it be configurable at all.

When set to true:

> When this property is true and the player changes from a paused state (rate of 0.0) to a played state (rate > 0.0), the player will try to determine if the current item can play to its end at the currently specified rate. If it determines that it’s likely to encounter a stall, the value of the player’s timeControlStatus will change to AVPlayer.TimeControlStatus.waitingToPlayAtSpecifiedRate and playback will automatically start when the likelihood of stalling has been minimized. A similar condition will occur during playback if the current player item’s playback buffer is exhausted and playback stalls. Playback will automatically resume when the likelihood of stalling has been minimized.

When set to false:

> If the value of this property is false, playback will start immediately when requested as long as the playback buffer is not empty. If the playback buffer becomes empty and playback stalls, the player’s timeControlStatus will switch to AVPlayer.TimeControlStatus.paused and the playback rate will change to 0.0.

Maybe I am missing something, but I can't think of a situation in the context of `rntp` where having the playback stop when the playback buffer is empty makes sense.

@curiousdustin perhaps you have some thoughts on this? Do you set this option to true in your app?

curiousdustin:
I have not used `waitForBuffer` in my app. 

[...]

Issue #601, opened 2019-05-21
( Bug ) Require second Press to play audio from Stream URL


shiroze:
# Configuration

react: 16.8.3
react-native: 0.59.8
react-native-track-player:1.1.4

# Issue
Test on iOS

Streaming URL, no Audio when first press.
The audio play when we press play twice.

Step :
1. Press play and state playing become true
2. Pause and play again then audio is playing

it work fine on build  v1.0.0 and URL File

# Code

```
import React, {Component} from 'react';
import { 
  Platform,
  StyleSheet,
  Text,
  View,
  TouchableOpacity,
  YellowBox
} from 'react-native';
import TrackPlayer from 'react-native-track-player';


const instructions = Platform.select({
  ios: 'Press Cmd+R to reload,\n' + 'Cmd+D or shake for dev menu',
  android:
    'Double tap R on your keyboard to reload,\n' +
    'Shake or press menu button for dev menu',
});

export default class App extends Component {
  constructor() {
    super();
    this.state = {
      isplaying: false
    };

    YellowBox.ignoreWarnings(['Setting a timer']);
	}

  componentWillMount () {
    TrackPlayer.setupPlayer();
  }

  componentDidMount() {
    TrackPlayer.updateOptions({
      stopWithApp: true,
      capabilities: [
        TrackPlayer.CAPABILITY_PLAY,
        TrackPlayer.CAPABILITY_STOP
      ],
      compactCapabilities: [
        TrackPlayer.CAPABILITY_PLAY,
      ]
    });
  }

  onPlay = async () => {
    const currentTrack = await TrackPlayer.getCurrentTrack();
    this.setState({
      isplaying: !this.state.isplaying
    });

    if (currentTrack == null && this.state.isplaying) {
      await TrackPlayer.reset();
      await TrackPlayer.add({
        id: "1",
        url: "https://live.hunter.fm/country",
        title: "Track Title",
        artist: "Track Artist",
        artwork: "https://www.playradio.rs/images/play_button.png"
      });
      await TrackPlayer.setVolume(1);

      await TrackPlayer.play();
    } else {
      await TrackPlayer.stop();
    }
    // { 
    //   "id": "1111",
    //   "url": "https://drive.google.com/uc?export=download&id=1AjPwylDJgR8DOnmJWeRgZzjsohi-7ekj",
    //   "title": "Longing",
    //   "artist": "David Chavez",
    //   "artwork": "https://picsum.photos/200"
    // },
  }

[...]

Issue #594, opened 2019-05-17
Pressing play when using network pauses and requires a second press


chrise86:
# Configuration
```
React Native Environment Info:
    System:
      OS: macOS 10.14.4
      CPU: (4) x64 Intel(R) Core(TM) i5-4288U CPU @ 2.60GHz
      Memory: 714.62 MB / 16.00 GB
      Shell: 5.3 - /bin/zsh
    Binaries:
      Node: 10.15.1 - ~/.nvm/versions/node/v10.15.1/bin/node
      Yarn: 1.16.0 - /usr/local/bin/yarn
      npm: 6.4.1 - ~/.nvm/versions/node/v10.15.1/bin/npm
      Watchman: 4.1.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 12.2, macOS 10.14, tvOS 12.2, watchOS 5.2
      Android SDK:
        API Levels: 23, 24, 25, 26, 27
        Build Tools: 23.0.1, 23.0.3, 25.0.0, 25.0.1, 25.0.2, 25.0.3, 26.0.0, 26.0.1, 26.0.2, 26.0.3, 27.0.1, 27.0.3, 28.0.0
        System Images: android-23 | Google APIs Intel x86 Atom, android-26 | Google APIs Intel x86 Atom
    IDEs:
      Android Studio: 2.3 AI-162.4069837
      Xcode: 10.2.1/10E1001 - /usr/bin/xcodebuild
    npmPackages:
      react: 16.8.3 => 16.8.3
      react-native: 0.59.5 => 0.59.5
```

Using `react-native-track-player` version `1.1.4`.

# Issue
When running the example code on iOS, but using a network stream, pressing play results in the following state changes:

1. playing
2. ready
3. loading
4. paused

This means you have to press play a second time to actually get it to play. Is there something I'm missing?

# Code

```jsx
import React, { Component } from 'react'
import { StyleSheet, Text, View } from 'react-native'
import TrackPlayer from 'react-native-track-player'
import { Player } from './src/components/Player'

interface Props {}

interface State {
  playbackState: TrackPlayer.State
}

export default class App extends Component<Props, State> {
  onStateChanged?: TrackPlayer.EmitterSubscription

  state: Readonly<State> = {
    playbackState: TrackPlayer.STATE_NONE
  }

  componentWillMount() {
    this.onStateChanged = TrackPlayer.addEventListener('playback-state', (data) => {
[...]

PR #1084, opened 2020-12-10
Add tip about `waitForBuffer`  for network media

cmlarsen:
I ran into the issue where the audio would pause immediately after buffering a stream and would require a second `play()` call to start playing.  After searching the issues found that a lot of people #601 #594 #616 #978.  While there is some discussion if this should be enabled by default or perhaps there is a deep issue. However, until it is resolve, (for some of us at least) this seems to fix the issue and making a note in a docs would be handy.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -18,6 +18,10 @@ TrackPlayer.setupPlayer().then(() => {
     // The player is ready to be used
 });
 ```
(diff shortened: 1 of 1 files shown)


curiousdustin:
Thank you for the suggestion. 

I have added `waitForBuffer` documentation in the `setupPlayer` section instead, with a similar note.

See https://github.com/react-native-kit/react-native-track-player/commit/a19abb1dfbe4c74b67d4cb28f44a3e8cc5708f4c

PR #1162, opened 2021-05-07
Fix TypeScript definitions

elliotdickison:
The useTrackPlayerProgress hook can take `null` as the second argument
    to trigger polling regardless of the player's state, but the type
    definition for the hook does not allow null. This change adds null as
    an allowed value to the type definition.

The TypeScript definitions do not currently include the STATE_CONNECTING
    player state. This change adds that constant to the definitions.


dcvz:
Thanks for this @elliotdickison I think this is no longer necessary on v2 :)

PR #1175, opened 2021-05-29
[iOS] playback-metadata-received support

dcvz:


almightybob:
Is this feature available as of a certain ios Version? I have implemented it and some users complain that it does not work on their old ios devices?

PR #1179, opened 2021-05-31
Removing redundant links in Readme

samadritsarkar2:
Removed the link to old docs (react-native-kit.github.io) as it was creating confusion. 
Also, the same was already updated with new docs link and this was just redundant.


dcvz:
Good catch @samadritsarkar2 !

PR #1181, opened 2021-06-03
Update Media controls capabilities on example

jonra1993:
On documentation, old constants are used which can cause problem to new people who are trying to use this library

image


dcvz:
thanks for this @jonra1993 we've gone and updated the documentation to show the exported types instead of the constant values!

PR #1196, opened 2021-07-01
Fix add() unable to insert before first track

austinried:
Calling `TrackPlayer.add(tracks, 0)` in order to add new tracks before the first track in the queue is currently broken due to 0 being falsy, thus defaulting to -1, and the tracks being added after the last track in the queue instead.  This change fixes that by explicitly checking for `undefined`, passing -1 in that case, and otherwise passing the number (including 0) on to the native code.


dcvz:
nice catch @austinried ! thanks for the fix

Issue #886, opened 2020-03-24
<iOS> Background player still exists after destroying track-player


sarykemmoun:
# Configuration

`react-native-track-player` version: 2.0.0-rc13

# Issue
I use track player to render a recording.
It works really fine, But after I'm destroying the player on iOS,
After locking the device the audio stream indeed stopped, but I still see the notification.
# Code

```
    const leavePlayer = async () => {
        await TrackPlayer.stop();
        await TrackPlayer.destroy();
    }
```

After executing this function the player destroys, But after locking the screen - this is the result.
![File](https://user-images.githubusercontent.com/28590349/77432188-699ca280-6de6-11ea-9150-7628e614392b.jpg)
Do you have any idea why does it happen? Thanks.

curiousdustin:
Does this happen only if you are viewing the lock screen when your leavePlayer function is called?

Does it happen if you are in the app when the function is called, and then press power button twice to view lock screen?

curiousdustin:
Related: 

https://github.com/react-native-kit/react-native-track-player/issues/825

https://github.com/jorgenhenrichsen/SwiftAudio/issues/83

marioot:
I can confirm this bug.
Destroying the player doesn't destroy the external notification player.

marioot:
Calling TrackPlayer.reset() before destroying fixes the issue. At least it's a workaround for now.

curiousdustin:
@marioot If you are viewing the lock screen when calling `reset()` and `destroy()`, does the notification go away?

waheedakhtar694:
@marioot calling reset before destroy does not work for me, @curiousdustin Any update on this?

nickmcmillan:
Experiencing the same issue here. In the return statement for the useEffect (the unmount) I've tried the following to no avail, the notification still persists on iOS. Android works as expected.
```
TrackPlayer.remove(trackId)
TrackPlayer.reset()
TrackPlayer.destroy()
```

PR #1212, opened 2021-07-26
Remove player from notification when destroying track player on iOs

MAVERlCK:
There were an issue destroying player on iOs ( tracked here [#886](https://github.com/DoubleSymmetry/react-native-track-player/issues/886) ), the destroy method didn't seemed to be linked to any code (except the print).

So I just followed instructions from a SwiftAudio issue (https://github.com/jorgenhenrichsen/SwiftAudio/issues/83#issuecomment-575966626) and seems to be working now for iOs.


uzegonemad:
@MAVERlCK @dcvz I think the destroy function also needs to set `hasInitialized = false`. Without it you won't be able to call setupPlayer again.

PR #1220, opened 2021-08-03
Update App.tsx

mehr30na:
Make example work on ios


dcvz:
@mehr30na what's wrong with the current example? does it not work for you?

PR #1232, opened 2021-08-17
Fix the problem of violating react hook rules

lyswhut:
This PR did two things:
1. Fixed the problem of violating react hook rules
2. In order to avoid similar problems in the future, I added the official [eslint-plugin-react-hooks](https://www.npmjs.com/package/eslint-plugin-react-hooks) eslint extension of react to check for problems that violate the rules of react hooks.

### DIFF
--- a/src/hooks.ts
+++ b/src/hooks.ts
@@ -75,6 +75,13 @@ export function useProgress(updateInterval?: number) {
   const [state, setState] = useState<ProgressState>({ position: 0, duration: 0, buffered: 0 })
   const playerState = usePlaybackState()
   const stateRef = useRef(state)
+  const isUnmountedRef = useRef(true)
+  useEffect(() => {
+    isUnmountedRef.current = false
+    return () => {
+      isUnmountedRef.current = true
+    }
+  }, [])
 
   const getProgress = async () => {
     const [position, duration, buffered] = await Promise.all([
@@ -82,6 +89,8 @@ export function useProgress(updateInterval?: number) {
       TrackPlayer.getDuration(),
       TrackPlayer.getBufferedPosition(),
     ])
+    // After the asynchronous code is executed, if the component has been uninstalled, do not update the status
+    if (isUnmountedRef.current) return
 
     if (
       position === stateRef.current.position &&
--- a/src/hooks.ts
+++ b/src/hooks.ts
@@ -58,7 +58,7 @@ export const useTrackPlayerEvents = (events: Event[], handler: Handler) => {
     )
 
     return () => subs.forEach(sub => sub.remove())
-  }, events)
+  }, [events])
 }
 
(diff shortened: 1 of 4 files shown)


src/hooks.ts
```diff
@@ -75,13 +75,22 @@ export function useProgress(updateInterval?: number) {
   const [state, setState] = useState<ProgressState>({ position: 0, duration: 0, buffered: 0 })
   const playerState = usePlaybackState()
   const stateRef = useRef(state)
+  const isUnmountedRef = useRef(true)
```
dcvz, line 78: ah, quite nice! you have tested this to work?
do you have an example app where I can see this failing before the fix?

src/hooks.ts
```diff
@@ -75,13 +75,22 @@ export function useProgress(updateInterval?: number) {
...
+  }, [])
 
   const getProgress = async () => {
     const [position, duration, buffered] = await Promise.all([
       TrackPlayer.getPosition(),
       TrackPlayer.getDuration(),
       TrackPlayer.getBufferedPosition(),
     ])
+    // After the asynchronous code is executed, if the component has been uninstalled, do not update the status
+    if (isUnmountedRef.current) return
```
dcvz, line 93: is there a need for this to happen after the async call? Did you find that the un-mounting happens while the async call is still executing?

dcvz (approved):
Thank you for this contribution @lyswhut ! It looks pretty solid!
I've left some questions for things I'd like to understand a bit better.

src/hooks.ts
dcvz: Actually, we probably want a similar fix for all hooks.

src/hooks.ts
lyswhut: Yes, this problem may occur when the status is updated after all asynchronous calls.
Although it may be rare, see: <https://www.debuggr.io/react-update-unmounted-component/>

src/hooks.ts
dcvz: I merged this, and applied the fix for `usePlaybackState`

lyswhut: > ah, quite nice! you have tested this to work?
> do you have an example app where I can see this failing before the fix?

This problem happened occasionally in my application before, and it can work normally after applying this fix

---

> Actually, we probably want a similar fix for all hooks. Do you feel up for making those changes @lyswhut?

Okay, I will take time to check other asynchronous hooks when I have time

dcvz: No need I've applied the fix now also to `usePlaybackState` here: https://github.com/DoubleSymmetry/react-native-track-player/commit/abd2670156bd77a07c0fb7d909a8137fe93f010d

Thanks for the PR! It'll be out soon with the next version.

PR #1253, opened 2021-09-06
Reset initialization flag on destroy

sreten-bild:
According to docs, calling `destroy()`:
`Destroys the player, cleaning up its resources. After executing this function, you won't be able to use the player anymore, unless you call setupPlayer() again.`

However since `hasInitialized` wasn't reset to false, it didn't run any setup code. This PR fixes the issue.

I noticed this while working on an app that plays and records sound multiple times. Since my recording code was messing with the shared instance, I needed a way to "restart" the track player.


dcvz (approved):
Thanks for this @sreten-bild !

PR #1255, opened 2021-09-08
Enable the use of rendering extensions for ExoPlayer

austinried:
This change allows ExoPlayer to be relied on and built locally in projects with extensions to support additional media formats, primarily the ffmpeg extension.  In the media player app I am building I wanted to add support for ALAC playback, which required this.

I have tested it both with and without the actual extension being present and it does not appear to cause any issues if you don't have extensions, it only prefers the extension for playback rendering if the audio format is both not supported by the system and also has an extension present to handle it.  You can read more about why these changes were needed to add support for extensions here: https://github.com/google/ExoPlayer/blob/r2.11.4/extensions/ffmpeg/README.md#using-the-extension

And if you want you can also see the changes I made to my app to use and build ExoPlayer locally to support this: https://github.com/austinried/subtracks/pull/15


jspizziri:
@austinried can you verify if this is still needed in `v2.2.0-rc3`? There's [a massive Android refactor](https://github.com/doublesymmetry/react-native-track-player/issues/1371) going on that would negate all of these changes when merged.

I'm going to preemptively close this PR because of that, but if it's an issue in `2.2.0-rc3` please let me know and we can discuss a PR against the `kotlin` refactor.

austinried:
@jspizziri According to the ExoPlayer docs it sounds like it should be (ALAC listed as supported by the ffmpeg extension still), I'll do some tests and see what happens though.

austinried:
@jspizziri Well I've done some tests and it seems like I'm getting an ALAC file to play, which I'm a little suspicious of since I don't see any info on ExoPlayer or your KotlinAudio adding built-in support for the ffmpeg extension or ALAC itself.  Do you know happen to know why this might be working now?

jspizziri:
@austinried , nothing concrete, with the exception that the underlying version of exoplayer has been upgraded (technically, it's using a doublesymmetry fork of exo) and that all the android code has been rewritten in the latest version.

In terms of how to proceed, it sounds like it's working for now. But if it breaks again in the future please let us know and we can work towards getting it addressed.

That work for you?

PR #1271, opened 2021-09-21
Unify the states used in events + getState

dcvz:


android/src/main/java/com/guichaguri/trackplayer/service/MusicService.kt
```diff
@@ -177,7 +178,7 @@ class MusicService : HeadlessJsTaskService() {
 
                 emit(MusicEvents.PLAYBACK_STATE, bundle)
             }
-        }
+            .launchIn(serviceScope)
 
         serviceScope.launch {
             event.audioItemTransition.collect {
```
mpivchev, line 184: I think we should decide if we should use `collect` or `forEach`, and use one on all observers? @dcvz

dcvz: Ah I actually didn't mean to include this in this PR but yes we should decide, I'd definitely opt for this way though..

but this is purely based on our experience in OneBeat

dcvz: but let's scrap this conversation for now.. let's stick to collect.

PR #1273, opened 2021-09-21
Fix JS Service

dcvz:
This fixes the issue where the `TrackPlayer.registerPlaybackService(() => require('./service'));` was no longer being recognised for play/pause and ducking.

### DIFF
--- a/android/src/main/java/com/guichaguri/trackplayer/module/MusicModule.kt
+++ b/android/src/main/java/com/guichaguri/trackplayer/module/MusicModule.kt
@@ -143,8 +143,13 @@ class MusicModule(private val reactContext: ReactApplicationContext?) :
         playerSetUpPromise = promise
         playerOptions = Arguments.toBundle(data)
 
-        if (!isServiceBound)
-            reactContext?.bindService(Intent(reactContext, MusicService::class.java), this, Context.BIND_AUTO_CREATE)
+        if (!isServiceBound) {
+            val intent = Intent(reactContext, MusicService::class.java)
+            reactContext?.startService(intent)
+            intent.action = MusicService.CONNECT
+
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/guichaguri/trackplayer/module/MusicModule.kt
```diff
@@ -143,8 +143,13 @@ class MusicModule(private val reactContext: ReactApplicationContext?) :
         playerSetUpPromise = promise
         playerOptions = Arguments.toBundle(data)
 
-        if (!isServiceBound)
-            reactContext?.bindService(Intent(reactContext, MusicService::class.java), this, Context.BIND_AUTO_CREATE)
+        if (!isServiceBound) {
+            val intent = Intent(reactContext, MusicService::class.java)
+            reactContext?.startService(intent)
+            intent.action = MusicService.CONNECT
```
mpivchev, line 149: So you need this action for the JS service to work?

dcvz: This I'm not completely sure of. But the `onBind` method of the previous implementation checked if it was explicitly this intent or not before trying to create a MusicBinder() and that seemed safer.

Let me know if you think that's incorrect. The important part was the `reactContext?.startService(intent)`

mpivchev: I think you can omit the action. We only ever start the service from the module, and via only this intent, so we don't need to specifically check for intents.

mpivchev (changes requested):
I think you can omit the `intent.action = MusicService.CONNECT`. We only ever start the service from the module, and via only this intent, so we don't need to specifically check for intents.

PR #1284, opened 2021-09-28
Replace old Android implementation with KotlinAudio

mpivchev:


bradleyflood:
Looking great guys 👍🏼

docs/Installation.md
```diff
@@ -21,10 +21,31 @@ yarn add react-native-track-player
 
 **2. (iOS) Enable Swift Modules**
 
-Because the iOS module uses Swift, if the user is using a standard react-native application they'll need to add support for Swift in the project. This can be easily by adding a swift file to the Xcode project -- could be called `dummy.swift` and saying yes when prompted if you'd like to generate a bridging header.
+Because the iOS module uses Swift, if the user is using a standard react-native application they'll need to add support for Swift in the project. This can be done easily by adding a swift file to the Xcode project -- could be called `dummy.swift` and saying yes when prompted if you'd like to generate a bridging header.
 
 ![Importing Swift](https://i.imgur.com/CBqBcWs.png)
 
+**3. (Android) Enable Kotlin**
+Because the Android module uses Kotlin, if the user is using a standard react-native application they'll need to add support for Kotlin in the project. This can be done easily by updating your `android/build.gradle` to include:
```
bradleyflood, line 29: So my `android/build.gradle` file says..
```
// NOTE: Do not place your application dependencies here; they belong
// in the individual module build.gradle files
```
Should we instead be recommending our users to add the dependency to `android/app/build.gradle` instead? Or will that not work?

src/trackPlayer.ts
```diff
@@ -28,7 +28,8 @@ async function setupPlayer(options: PlayerOptions = {}): Promise<void> {
   return TrackPlayer.setupPlayer(options || {})
 }
 
-function destroy() {
+function destroy(): Promise<void> {
+  emitter.removeAllListeners()
```
jenshandersson, line 32: Should we call this for iOS as well or only Android?

PR #1287, opened 2021-09-29
Respect given index when calling add

dcvz:
Should be used with KotlinAudio branch of the same name.

### DIFF
--- a/android/src/main/java/com/guichaguri/trackplayer/module/MusicModule.kt
+++ b/android/src/main/java/com/guichaguri/trackplayer/module/MusicModule.kt
@@ -187,22 +187,18 @@ class MusicModule(private val reactContext: ReactApplicationContext?) :
 
         val tracks: MutableList<Track> = mutableListOf()
         val bundleList = Arguments.toList(data)
-        if (bundleList != null) {
-            bundleList.forEach { trackBundle ->
-                if (trackBundle is Bundle) {
-                    try {
-                        tracks.add(
-                            Track(reactApplicationContext, trackBundle, RatingCompat.RATING_HEART)
-                        )
-                    } catch(ex: Exception) {
-                        callback.reject("invalid_track_object", ex)
-                    }
-                } else {
-                    callback.reject("invalid_track_object", "Track was not a dictionary type")
-                }
-            }
-        } else {
+
+        if (bundleList == null) {
             callback.reject("invalid_parameter", "Was not given an array of tracks")
+            return
+        }
+
+        bundleList.forEach {
+            if (it is Bundle) {
+                tracks.add(Track(reactApplicationContext, it, RatingCompat.RATING_HEART))
+            } else {
+                callback.reject("invalid_track_object", "Track was not a dictionary type")
+            }
         }
 
         when {
--- a/android/src/main/java/com/guichaguri/trackplayer/module/MusicModule.kt
+++ b/android/src/main/java/com/guichaguri/trackplayer/module/MusicModule.kt
@@ -185,21 +185,42 @@ class MusicModule(private val reactContext: ReactApplicationContext?) :
     fun add(data: ReadableArray?, insertBeforeIndex: Int, callback: Promise) {
         if (verifyServiceBoundOrReject(callback)) return
 
+        val tracks: MutableList<Track> = mutableListOf()
         val bundleList = Arguments.toList(data)
-        val tracks: List<Track> = try {
-            Track.createTracks(
-                reactApplicationContext, bundleList, RatingCompat.RATING_HEART
-            )!!
-
-
-        } catch (ex: Exception) {
-            callback.reject("invalid_track_object", ex)
-            return
+        if (bundleList != null) {
+            for (trackBundle in bundleList) {
+                if (trackBundle is Bundle) {
+                    try {
+                        tracks.add(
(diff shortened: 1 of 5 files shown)


android/src/main/java/com/guichaguri/trackplayer/module/MusicModule.kt
```diff
@@ -185,21 +185,42 @@ class MusicModule(private val reactContext: ReactApplicationContext?) :
...
-            Track.createTracks(
-                reactApplicationContext, bundleList, RatingCompat.RATING_HEART
-            )!!
-
-
-        } catch (ex: Exception) {
-            callback.reject("invalid_track_object", ex)
-            return
+        if (bundleList != null) {
+            for (trackBundle in bundleList) {
```
mpivchev, line 191: you can use `forEach`

android/src/main/java/com/guichaguri/trackplayer/module/MusicModule.kt
```diff
@@ -185,21 +185,42 @@ class MusicModule(private val reactContext: ReactApplicationContext?) :
...
+                        )
+                    } catch(ex: Exception) {
+                        callback.reject("invalid_track_object", ex)
+                    }
+                } else {
+                    callback.reject("invalid_track_object", "Track was not a dictionary type")
+                }
+            }
+        } else {
+            callback.reject("invalid_parameter", "Was not given an array of tracks")
```
mpivchev, line 205: It looks like there are quite a few nested calls here. Maybe it can be simplified.

android/src/main/java/com/guichaguri/trackplayer/module/MusicModule.kt
```diff
@@ -185,21 +185,42 @@ class MusicModule(private val reactContext: ReactApplicationContext?) :
...
+                    add(tracks)
+                    callback.resolve(null)
+                }
+            }
+            else -> {
+                musicService.apply {
+                    add(tracks, insertBeforeIndex)
+                    callback.resolve(null)
+                }
+            }
```
mpivchev, line 223: `If else` might be more suitable here. Up to you.

dcvz: The kotlin compiler suggested I use when instead of if's :)

android/src/main/java/com/guichaguri/trackplayer/module/MusicModule.kt
dcvz: Do you have a suggestion?

android/src/main/java/com/guichaguri/trackplayer/module/MusicModule.kt
```diff
@@ -185,21 +185,42 @@ class MusicModule(private val reactContext: ReactApplicationContext?) :
...
-        } catch (ex: Exception) {
-            callback.reject("invalid_track_object", ex)
-            return
+        if (bundleList != null) {
+            for (trackBundle in bundleList) {
+                if (trackBundle is Bundle) {
+                    try {
+                        tracks.add(
+                            Track(reactApplicationContext, trackBundle, RatingCompat.RATING_HEART)
+                        )
```
mpivchev, line 196: It seems your AS is wrapping code very aggressively. You can fix this by going to Preferences -> Code Style -> Kotlin -> Wrapping and Braces -> set Hard wrap at to 999

android/src/main/java/com/guichaguri/trackplayer/module/MusicModule.kt
mpivchev: I usually consider `when`s more like a `switch`. And in this case where we have different types of conditioning I prefer to use `if else`. But you can leave it like this if you prefer.

android/src/main/java/com/guichaguri/trackplayer/module/MusicModule.kt
dcvz: This is a code style. Let's comment on these when we want to put a practice in place.

android/src/main/java/com/guichaguri/trackplayer/module/MusicModule.kt
dcvz: I do what the compiler tells me to do :) Since its about code style and we haven't yet decided on a type let's keep it like this.

android/src/main/java/com/guichaguri/trackplayer/module/MusicModule.kt
mpivchev: You can do something like this:

```kotlin
        if (bundleList == null) {
            callback.reject("invalid_parameter", "Was not given an array of tracks")
            return
        }

        bundleList.forEach {
            if (it is Bundle) {
                tracks.add(Track(reactApplicationContext, it, RatingCompat.RATING_HEART))
            } else {
                callback.reject("invalid_track_object", "Track was not a dictionary type")
            }
        }
```

You can flip the null check and also remove the `try catch` as I don't think anything can throw an exception inside `Track`

dcvz: cool. I like this! I had the try catch cause it was there before and wasn't sure if something could throw.

mpivchev (approved):
LGTM :)

PR #1289, opened 2021-09-29
Remove old module files

mpivchev:


android/src/main/AndroidManifest.xml
```diff
@@ -1,26 +0,0 @@
-<?xml version="1.0" encoding="utf-8"?>
```
dcvz, line 1: I would not delete this file. I would instead remove the line from the example AndroidManifest that specifies the service:
```xml
<service
            android:name="com.guichaguri.trackplayer.service.MusicService"
            android:enabled="true" />
```

Plus we need the `<receiver android:name="androidx.media.session.MediaButtonReceiver">` to be registered.

Either way I think its best that the lib has this than asking the apps to add this to their Manifest files. Most people are generating RN projects via CLI and they shouldn't have to modify this to make the player work.

example/android/app/src/main/AndroidManifest.xml
```diff
@@ -23,7 +23,7 @@
         </activity>
 
         <service
```
dcvz, line 25: Following up with my other comment, let's remove this line in favour of having the Manifest at the RN lib level.

dcvz (changes requested):
Yay for cleanup and renaming! Some changes required!

android/build.gradle
```diff
@@ -63,32 +63,6 @@ dependencies {
     //noinspection GradleDynamicVersion
     implementation 'com.facebook.react:react-native:+'
 
-    def exoPlayerVersion = safeExtGet("exoPlayerVersion", '2.11.4')
```
dcvz, line 66: On the top of this file there's more variables related to dash, his and smooth streaming that don't seem necessary anymore (are unused).

android/build.gradle
```diff
@@ -63,32 +63,6 @@ dependencies {
...
-
-    // ExoPlayer SmoothStreaming
-    if (smoothstreaming) {
-        implementation "com.google.android.exoplayer:exoplayer-smoothstreaming:${exoPlayerVersion}"
-    } else {
-        compileOnly "com.google.android.exoplayer:exoplayer-smoothstreaming:${exoPlayerVersion}"
-    }
-
     // Make sure we're using androidx
     implementation "androidx.core:core-ktx:1.7.0-alpha01"
```
dcvz, line 67: have you reviewed the existing imports. Are they all required?

android/src/main/AndroidManifest.xml
```diff
@@ -1,26 +0,0 @@
...
-<manifest xmlns:android="http://schemas.android.com/apk/res/android"
-          package="com.guichaguri.trackplayer">
-
-    <uses-permission android:name="android.permission.WAKE_LOCK" />
-    <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
-
-    <application>
-
-        <!-- The main service, handles playback, playlists and media buttons -->
-        <service android:name=".service_old.MusicService" android:enabled="true" android:exported="true">
```
dcvz, line 11: This one should change from `.service_old` to `.service` though.

android/build.gradle
```diff
@@ -13,14 +13,9 @@ buildscript {
 apply plugin: 'com.android.library'
 apply plugin: 'kotlin-android'
 
-boolean dash = false, hls = false, smoothstreaming = false
-
 File file = file('../../../track-player.json')
```
dcvz: all of this file stuff should also go away, we're not using any of it.

example/android/app/build.gradle
```diff
@@ -115,21 +115,21 @@ def jscFlavor = 'org.webkit:android-jsc:+'
...
  * and the benefits of using Hermes will therefore be sharply reduced.
  */
-def enableHermes = project.ext.react.get("enableHermes", false);
+def enableHermes = project.ext.react.get("enableHermes", false)
 
 android {
     ndkVersion rootProject.ext.ndkVersion
 
-    compileSdkVersion rootProject.ext.compileSdkVersion
+    compileSdkVersion 31
```
dcvz, line 127: These should be updated at the example root project level: `/example/android/build.gradle`

example/android/app/build.gradle
```diff
@@ -115,21 +115,21 @@ def jscFlavor = 'org.webkit:android-jsc:+'
...
 
     defaultConfig {
         applicationId "com.example"
-        minSdkVersion rootProject.ext.minSdkVersion
-        targetSdkVersion rootProject.ext.targetSdkVersion
+        minSdkVersion 21
+        targetSdkVersion 31
```
dcvz, line 132: These should be updated at the example root project level: `/example/android/build.gradle`

android/src/main/java/com/doublesymmetry/trackplayer/Utils.kt
```diff
@@ -16,11 +16,10 @@ import com.google.android.exoplayer2.upstream.RawResourceDataSource
 /**
  * @author Guichaguri
```
dcvz, line 17: there's still a few scattered around like this that we could completely remove. Unless they're still original files and you'd like to keep them.

android/build.gradle
```diff
@@ -12,27 +12,17 @@ buildscript {
...
-    dash = json.dash ?: dash
-    hls = json.hls ?: hls
-    smoothstreaming = json.smoothstreaming ?: smoothstreaming
-}
-
-def safeExtGet(prop, fallback) {
-    rootProject.ext.has(prop) ? rootProject.ext.get(prop) : fallback
-}
+//
+//def safeExtGet(prop, fallback) {
```
dcvz, line 16: I would restore this and continue using it below.

android/build.gradle
```diff
@@ -12,27 +12,17 @@ buildscript {
...
+//    rootProject.ext.has(prop) ? rootProject.ext.get(prop) : fallback
+//}
 
 android {
-    compileSdkVersion safeExtGet("compileSdkVersion", 28)
```
dcvz, line 31: I would restore this, we can just ask it to use 31 instead.

android/build.gradle
```diff
@@ -12,27 +12,17 @@ buildscript {
...
+    compileSdkVersion 31
 
     defaultConfig {
-        minSdkVersion safeExtGet("minSdkVersion", 16) // RN's minimum version
```
dcvz, line 34: same for these two.

dcvz (changes requested):
A few more changes! I'd also love to review this after the merge conflicts are fixed. Could you rebase the branch before the next review so I can give it one final check?

android/build.gradle
dcvz: Here's another lib as a reference: https://github.com/react-native-video/react-native-video/blob/master/android/build.gradle

android/src/main/java/com/doublesymmetry/trackplayer/Utils.kt
mpivchev: Yes we still use the `Utils`

android/src/main/AndroidManifest.xml
```diff
@@ -1,26 +1,26 @@
...
+          package="com.doublesymmetry.trackplayer">
 
     <uses-permission android:name="android.permission.WAKE_LOCK" />
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
 
     <application>
 
         <!-- The main service, handles playback, playlists and media buttons -->
-        <service android:name=".service_old.MusicService" android:enabled="true" android:exported="true">
+        <service android:name="com.doublesymmetry.trackplayer.service.MusicService" android:enabled="true" android:exported="false">
```
dcvz, line 11: did you change the `exported` to false on purpose?

dcvz (commented):
Just one important question!

android/src/main/AndroidManifest.xml
mpivchev: Yes. We don't want other random apps to use our service.

dcvz: aha, got it. oki.

dcvz (approved):
LGTM 👍

PR #1292, opened 2021-09-30
Fix support for custom media types + additional track options

dcvz:
Fixes AEX-40


android/src/main/java/com/guichaguri/trackplayer/model/TrackType.kt
```diff
@@ -1,26 +0,0 @@
-package com.guichaguri.trackplayer.model
```
dcvz, line 1: There was no reason to have this duplicated class when we can use the library one directly.

android/src/main/java/com/guichaguri/trackplayer/model/Track.kt
```diff
@@ -35,97 +36,33 @@ class Track(context: Context, bundle: Bundle, ratingType: Int) : TrackMetadata()
...
+                userAgent,
+                resourceId
+            )
         )
     }
 
     init {
         resourceId = Utils.getRawResourceId(context, bundle, "url")
         uri = if (resourceId == 0) {
+            resourceId = null
```
dcvz, line 59: It looks like KotlinAudio expects this to be null, if its not to be used, so in a case that we're not using a local asset I set this back to null, so its not used when we call `toAudioItem`

mpivchev (approved):
Nice 👍

PR #1307, opened 2021-10-14
Feat: Remove previous track on queue

ithustle:
This feature will help to anyone who want implements their own shuffle mode. The reason of this method was built its because `.remove()` for iOS doesn't work as expected when trying to remove bulk tracks.

### DIFF
--- a/ios/RNTrackPlayer/RNTrackPlayer.swift
+++ b/ios/RNTrackPlayer/RNTrackPlayer.swift
@@ -355,6 +355,13 @@ public class RNTrackPlayer: RCTEventEmitter {
         resolve(NSNull())
     }
 
+    @objc(removePreviousTracks:rejecter:)
+    public func removePreviousTracks(resolve: RCTPromiseResolveBlock, reject: RCTPromiseRejectBlock) {
+        print("Removing Previous tracks")
+        player.removePreviousItems()
+        resolve(NSNull())
+    }
+
     @objc(removeUpcomingTracks:rejecter:)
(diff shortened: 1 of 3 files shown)


dcvz:
Thanks for the PR @ithustle , will take a look at it soon!

rogerespinoza:
Hello, maybe it will help you  
`tracksRemoveIndex.forEach(async (element) => {
     await TrackPlayer.remove([0]);
  });`

ithustle:
@bradleyflood, yes off course!

bradleyflood (changes requested):
Good point @jspizziri I hadn’t seen your earlier message. 

Need to confirm functionality and add Android support. 

@dcvz

dcvz:
@bradleyflood @jspizziri @ithustle I'm not opposed to this addition but there's some things to keep in mind:

1. We need to add support for other platforms too.

2. It's a bit more complicated to remove previous tracks because it changes the index of the currently playing track.. so we'll need to make sure this doesn't break something. SwiftAudioEx already has support for this but not Android. Though we already support removing indexes smaller than the current one.. so there's precedence.

3. We'll want to make this change against the current `main`

jspizziri:
@ithustle Looks like you're cleared for takeoff. I can assist with initial review and provide some implementation guidance along the way should you need it.

ithustle:
@jspizziri I appreciate that. I'd like to start updating the doc. I couldn't find the MD file.

jspizziri:
@ithustle You can find it under `docs/docs/api/functions/queue.md`

jspizziri:
@ithustle I'm closing this since you opened another PR concerning the same feature with docs. Let's keep all commits concerning a feature on the same PR.

Issue #1301, opened 2021-10-08
npm ERR: Could not resolve dependency


besh81:
Hi all,

just started a new react native app (0.66.0) and trying to install track player:

this is the command I used:
`npm install --save react-native-track-player`

this is the error:
```
npm ERR! code ERESOLVE
npm ERR! ERESOLVE unable to resolve dependency tree
npm ERR!
npm ERR! While resolving: rml@1.7.0
npm ERR! Found: react-native@0.66.0
npm ERR! node_modules/react-native
npm ERR!   react-native@"0.66.0" from the root project
npm ERR!   peer react-native@">=0.60.0-rc.2" from react-native-track-player@2.1.1
npm ERR!   node_modules/react-native-track-player
npm ERR!     react-native-track-player@"*" from the root project
npm ERR!
npm ERR! Could not resolve dependency:
npm ERR! peer react-native@"^0.65.0" from react-native-windows@0.65.4
npm ERR! node_modules/react-native-windows
npm ERR!   peer react-native-windows@">=0.63.0" from react-native-track-player@2.1.1
npm ERR!   node_modules/react-native-track-player
npm ERR!     react-native-track-player@"*" from the root project
npm ERR!
npm ERR! Fix the upstream dependency conflict, or retry
npm ERR! this command with --force, or --legacy-peer-deps
npm ERR! to accept an incorrect (and potentially broken) dependency resolution.
```

I'm not sure but seems that react-native-windows is missing, isn't it?
If so why should I install it even for android app ???

Here react-native info:
```
info Fetching system and libraries information...
System:
    OS: Windows 10 10.0.19043
    CPU: (20) x64 Intel(R) Core(TM) i9-10900 CPU @ 2.80GHz
    Memory: 21.92 GB / 31.84 GB
  Binaries:
    Node: 16.10.0 - C:\Program Files\nodejs\node.EXE
    Yarn: Not Found
    npm: 7.24.0 - C:\Program Files\nodejs\npm.CMD
    Watchman: Not Found
  SDKs:
    Android SDK:
      API Levels: 29, 30, 31
      Build Tools: 30.0.2, 31.0.0
      System Images: android-29 | Google APIs Intel x86 Atom, android-29 | Google Play Intel x86 Atom
      Android NDK: Not Found
    Windows SDK: Not Found
  IDEs:
    Android Studio: Version     2020.3.0.0 AI-203.7717.56.2031.7678000
[...]

PR #1324, opened 2021-11-04
fix(package.json): react-native-windows

jspizziri:
Make `react-native-windows` and optional dependency. Fixes https://github.com/DoubleSymmetry/react-native-track-player/issues/1301


dcvz:
Thanks for this @jspizziri !

Issue #1235, opened 2021-08-19
Could not find com.google.android.exoplayer:exoplayer-core:2.11.4.


giorodri0584:
**Describe the bug**
Hi I install react native track player on a fresh project
 and Im getting this error on android only:
Could not find com.google.android.exoplayer:exoplayer-core:2.11.4.
Searched in the following locations:
       - https://repo.maven.apache.org/maven2/com/google/android/exoplayer/exoplayer-core/2.11.4/exoplayer-core-2.11.4.pom
       - https://dl.google.com/dl/android/maven2/com/google/android/exoplayer/exoplayer-core/2.11.4/exoplayer-core-2.11.4.pom
       - https://www.jitpack.io/com/google/android/exoplayer/exoplayer-core/2.11.4/exoplayer-core-2.11.4.pom

**To Reproduce**
yarn add react-native-track-player

**Environment (please complete the following information):**
Run `react-native info` in your project and share the content.
System:
  Binaries:
    Node: 14.17.5 - /usr/local/bin/node
    Yarn: 1.22.10 - /usr/local/bin/yarn
    npm: 7.20.6 - /usr/local/bin/npm
    Watchman: 2021.08.02.00 - /usr/local/bin/watchman
  Managers:
    CocoaPods: 1.10.2 - /usr/local/bin/pod
  SDKs:
    iOS SDK:
      Platforms: iOS 14.5, DriverKit 20.4, macOS 11.3, tvOS 14.5, watchOS 7.4
    Android SDK: Not Found
  IDEs:
    Android Studio: 2020.3 AI-203.7717.56.2031.7583922
    Xcode: 12.5.1/12E507 - /usr/bin/xcodebuild
  Languages:
    Java: 15.0.2 - /usr/bin/javac
  npmPackages:
    @react-native-community/cli: Not Found
    react: 17.0.2 => 17.0.2
    react-native: 0.65.0 => 0.65.0
    react-native-macos: Not Found
  npmGlobalPackages:
    *react-native*: Not Found
What `react-native-track-player` version are you using? 2.0.3
Are you testing on a real device or in the simulator?  android emulator
Which OS version are you running? MacOs Big Sur

**Code**
I get the error running the app

tried adding different maven url but none work.

dcvz:
Do you also get this issue when running the example?

Tucker2015:
Same issue here with fresh project.

dcvz:
[...]

PR #1335, opened 2021-11-23
added jcenter to build grade

richardsoncarvalho:
In new installations of react native, the android is giving an error due to the lack of jcenter()

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -3,6 +3,7 @@ import groovy.json.JsonSlurper
 buildscript {
     repositories {
         mavenCentral()
+        jcenter()
         google()
     }
     dependencies {
@@ -53,6 +54,7 @@ repositories {
     }
 
     mavenCentral()
+    jcenter()
     google()
 }
 


jspizziri:
@richardsoncarvalho , I believe that you should make this MR against the `feature/kotlinaudio` branch as that is effectively the bleeding edge at this point. Also IMHO, I would add some comments above the `jcenter()` additions to specify which packages rely on jcenter and at what point we'll be able to remove it.

jspizziri:
@mpivchev , some form of this would be good to get into an RC release. Adding `jcenter` to the lib build.gradle should remove the dependency from user projects and would allow it to be removed easily once all `jcenter` packages have been migrated to `maven`.

This particular MR needs to be rebased on the `feature/kotlinaudio` branch and could probably use some better comments. If you're down with that approach I can submit a fresh MR that addresses both of those issues.

Issue reference https://github.com/DoubleSymmetry/react-native-track-player/issues/1235#issuecomment-921266932

jspizziri:
I've created https://github.com/DoubleSymmetry/react-native-track-player/pull/1456 which solves some of the issues I mentioned in my comment regarding this MR.

jspizziri:
@dcvz should this PR get closed since this is a non-issue in the `2.2.0` releases?

jspizziri:
@richardsoncarvalho , this issue is addressed in the `2.2.0` release. You can install `2.2.0-rc3` today to verify that it's been resolved.

PR #1366, opened 2022-01-12
Fix: Ambient session not working

grubicv:
.longForm / .longFormAudtio can't be used with the .ambient session category - this PR fixes that issue as it uses .default policy instead of longForm

### DIFF
--- a/ios/RNTrackPlayer/RNTrackPlayer.swift
+++ b/ios/RNTrackPlayer/RNTrackPlayer.swift
@@ -195,9 +195,9 @@ public class RNTrackPlayer: RCTEventEmitter {
         // Progressively opt into AVAudioSession policies for background audio
         // and AirPlay 2.
         if #available(iOS 13.0, *) {
-            try? AVAudioSession.sharedInstance().setCategory(sessionCategory, mode: sessionCategoryMode, policy: .longFormAudio, options: sessionCategoryOptions)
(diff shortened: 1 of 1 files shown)


bradleyflood:
Great, thanks @grubicv. As a JavaScript dev I have no idea what this is change is brining/improving - could you explain the what and why in some more detail here?

Thanks.

grubicv:
@bradleyflood basically you want to use ambient session category on iOS devices when you want to play a sound without affecting the current media being played (such as apple music using avaudiosession) and only when the app is open - in our case it's playing a beat playlist as a backing track for apple music. 
https://developer.apple.com/documentation/avfaudio/avaudiosession/category/1616560-ambient

The issue is that iOS doesn't allow you to use ambient as a session category if you put a policy for longForm/longFormAudio since that one is used for playing tracks on the main audio route so it's basically conflicting (you don't want to play it as a regular track, but rather as an ambient sound, but you're trying to route it to the longForm route). In that case an exception occurs on that line stating that and basically blocks you from playing sounds in ambient mode since it can't be set. Therefore you need to use the default policy route that's used for playing ambient audio, so I just added a simple condition that uses .default policy route if user wants to use ambient session category, and longForm/longFormAudio for other things.

bradleyflood (approved):

PR #1375, opened 2022-01-14
Update feature_request.md

bradleyflood:
I'd like to be more explicit about asking for help. We are not paid to developer everyone's feature requests!


dcvz:
Love it! Sounds like a good addition. Not only that but we'll always try to make additions that make sense as a whole to the library

dcvz (approved):

PR #1381, opened 2022-01-17
Refactor example app

puckey:
During testing, I noticed a bug in the example app where it would show the wrong currently playing track when you quit the app and start it again. While I was fixing this bug, I also tidied up the code a bit.

- start with currently playing track, even if app was killed in the meantime
- simplify isLoading check
- remove unnecessary check for Event.PlaybackTrackChanged
- avoid useCallback recalcs by passing `isPlaying` boolean instead of State
- remove unnecessary awaits in onSlidingComplete and togglePlayback
- always check if the component is still mounted after awaits
- remove duplicate usePlaybackState call

(Also, let me know if you would prefer the PR go directly to the master branch)

### DIFF
--- a/example/App.tsx
+++ b/example/App.tsx
@@ -15,6 +15,7 @@ import TrackPlayer, {
   Event,
   RepeatMode,
   State,
+  Track,
   usePlaybackState,
   useProgress,
   useTrackPlayerEvents,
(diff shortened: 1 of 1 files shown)


dcvz:
Thanks @puckey ! Really appreciate this

PR #1397, opened 2022-01-27
Code Fix: add resolves index on ios

formula1:
Important to note, I only tested this on the react-native-track-player example with a few items and setting the index. Looks like it worked ok but it's not a great substitute for a proper test. I probably should write a test to ensure both android and ios are returning the expected index but I'd like to continue my project for now.

### DIFF
--- a/ios/RNTrackPlayer/RNTrackPlayer.swift
+++ b/ios/RNTrackPlayer/RNTrackPlayer.swift
@@ -331,7 +331,7 @@ public class RNTrackPlayer: RCTEventEmitter {
             tracks.append(track)
         }
 
-        var index
+        var index: Int = 0;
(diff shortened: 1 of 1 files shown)


formula1:
Related to #1399

formula1:
Also wanted to add, I have never programmed in swift before so if I'm doing anything wrong I am open to learning

bradleyflood (approved):

bradleyflood:
Looks fine to me. @dcvz what do you think?

formula1:
@dcvz, please o.o

Maybe he's busy.

dcvz (approved):
Thanks for the fix @formula1 ! LGTM 👍

PR #1404, opened 2022-02-08
Fix bug that causes the notification play/pause button to get stuck

mpivchev:
The actual fix is in the KotlinAudio repo: https://github.com/DoubleSymmetry/KotlinAudio/pull/16


dcvz (approved):
LGTM 👍

PR #1411, opened 2022-02-22
Fix tracks not being removed from queue

mpivchev:


android/src/main/java/com/doublesymmetry/trackplayer/module/MusicModule.kt
```diff
@@ -273,22 +275,23 @@ class MusicModule(reactContext: ReactApplicationContext) : ReactContextBaseJavaM
...
+                        return@getCurrentTrackIndex
+                    } else if (index >= 0 && index < queue.size) {
+                        musicService.remove(listOf(index))
+                    }
                 }
             }
         }
 
-        if (indexes.isNotEmpty())
-            musicService.remove(indexes)
```
mpivchev, line 290: We are trying to remove indexes before `getCurrentTrackIndex` has finished collecting all the indexes.

android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
```diff
@@ -157,6 +173,10 @@ class MusicService : HeadlessJsTaskService() {
         handler.post { player.add(items, atIndex) }
     }
 
+    fun remove(index: Int) {
+        remove(listOf(index))
+    }
+
```
mpivchev, line 179: Added some helper functions for single indexes

android/src/main/java/com/doublesymmetry/trackplayer/module/MusicModule.kt
```diff
@@ -273,22 +297,23 @@ class MusicModule(reactContext: ReactApplicationContext) : ReactContextBaseJavaM
...
-                if (index == it) return@getCurrentTrackIndex
-                if (index >= 0 && index < queue.size) {
-                    indexes.add(index)
+
+        if (trackList != null) {
+            for (track in trackList) {
+                val index = if (track is Int) track else track.toString().toInt()
+
+                // We do not allow removal of the current item
+                musicService.getCurrentTrackIndex {
```
dcvz, line 306: Hm.. should we really be calling this async method over and over again then? Maybe its best to just get the current index once then do all comparisons against that?

since we can't await or suspend maybe something like:
```kotlin
if (trackList != null) {
  musicService.getCurrentTrackIndex {
    val indexes: MutableList<Int> = ArrayList()
    for (track in trackList) {
      val index = if (o is Int) o else o.toString().toInt()
      if (index == it) return@getCurrentTrackIndex
        if (index >= 0 && index < queue.size) {
          indexes.add(index)
    }
    
    if (indexes.isNotEmpty())
      musicService.remove(indexes)
    }
}
```

_This was written in the browser so might not be perfect.. but you get the idea_

mpivchev: Yeah sure, this is a nice optimization :)

mpivchev: @dcvz try now

dcvz (approved):
LGTM 👍

PR #1413, opened 2022-02-22
Feat/seek initial time

jspizziri:
Based on the [conversation and decisions made on this PR](https://github.com/DoubleSymmetry/react-native-track-player/pull/713#issuecomment-1036286082). This PR implements the following interface:

```ts
interface TrackPlayer {
  public async skipToNext(initialTime?: number);
  public async skipToPrevious(initialTime?: number);
  public async skip(index: number, initialTime?: number);
}
```

### DIFF
--- a/android/src/main/java/com/doublesymmetry/trackplayer/module/MusicModule.kt
+++ b/android/src/main/java/com/doublesymmetry/trackplayer/module/MusicModule.kt
@@ -375,25 +375,40 @@ class MusicModule(reactContext: ReactApplicationContext) : ReactContextBaseJavaM
     }
 
     @ReactMethod
-    fun skip(index: Int, callback: Promise) {
+    fun skip(index: Int, initialTime: Float, callback: Promise) {
         if (verifyServiceBoundOrReject(callback)) return
 
         musicService.skip(index)
+
+        if (initialTime >= 0) {
+            musicService.seekTo(initialTime)
+        }
+
         callback.resolve(null)
     }
 
     @ReactMethod
-    fun skipToNext(callback: Promise) {
+    fun skipToNext(initialTime: Float, callback: Promise) {
         if (verifyServiceBoundOrReject(callback)) return
         musicService.skipToNext()
+
+        if (initialTime >= 0) {
+            musicService.seekTo(initialTime)
+        }
+
         callback.resolve(null)
     }
 
     @ReactMethod
-    fun skipToPrevious(callback: Promise) {
+    fun skipToPrevious(initialTime: Float, callback: Promise) {
         if (verifyServiceBoundOrReject(callback)) return
 
         musicService.skipToPrevious()
+
+        if (initialTime >= 0) {
+            musicService.seekTo(initialTime)
+        }
+
         callback.resolve(null)
     }
 
--- a/docs/Documentation.md
+++ b/docs/Documentation.md
@@ -173,7 +173,7 @@ Removes one or more tracks from the queue.
 | ------ | -------- | ------------- |
 | tracks | `array` of track indexes or a single one | The tracks that will be removed |
 
-#### `skip(index)`
+#### `skip(index, initialTime)`
 Skips to a track in the queue.
 
 **Returns:** `Promise`
@@ -181,16 +181,26 @@ Skips to a track in the queue.
 | Param  | Type     | Description     |
 | ------ | -------- | --------------- |
 | index  | `number` | The track index |
+| initialTime | `number` | **Optional.** Sets the initial playback for the track you're skipping to. |
 
-#### `skipToNext()`
+#### `skipToNext(initialTime)`
 Skips to the next track in the queue.
 
 **Returns:** `Promise`
 
-#### `skipToPrevious()`
+| Param  | Type     | Description     |
+| ------ | -------- | --------------- |
+| initialTime | `number` | **Optional.** Sets the initial playback for the track you're skipping to. |
+
+#### `skipToPrevious(initialTime)`
 Skips to the previous track in the queue.
 
 **Returns:** `Promise`
+
+| Param  | Type     | Description     |
+| ------ | -------- | --------------- |
+| initialTime | `number` | **Optional.** Sets the initial playback for the track you're skipping to. |
+
 
 #### `reset()`
 Resets the player stopping the current track and clearing the queue.
--- a/ios/RNTrackPlayer/RNTrackPlayer.swift
+++ b/ios/RNTrackPlayer/RNTrackPlayer.swift
@@ -299,7 +299,7 @@ public class RNTrackPlayer: RCTEventEmitter {
             reject("player_not_initialized", "The player is not initialized. Call setupPlayer first.", nil)
             return
         }
-        
+
         print("Destroying player")
         self.player.stop()
         self.player.nowPlayingInfoController.clear()
@@ -362,7 +362,7 @@ public class RNTrackPlayer: RCTEventEmitter {
             index = trackIndex.intValue
             try? player.add(items: tracks, at: trackIndex.intValue)
         }
-        
+
         resolve(index)
     }
 
@@ -393,8 +393,13 @@ public class RNTrackPlayer: RCTEventEmitter {
         resolve(NSNull())
     }
 
-    @objc(skip:resolver:rejecter:)
-    public func skip(to trackIndex: NSNumber, resolve: RCTPromiseResolveBlock, reject: RCTPromiseRejectBlock) {
+    @objc(skip:initialTime:resolver:rejecter:)
+    public func skip(
+        to trackIndex: NSNumber,
+        initialTime: Double,
+        resolve: RCTPromiseResolveBlock,
+        reject: RCTPromiseRejectBlock
+    ) {
         if !hasInitialized {
             reject("player_not_initialized", "The player is not initialized. Call setupPlayer first.", nil)
             return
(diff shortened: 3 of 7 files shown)


dcvz:
Thanks for working on this @jspizziri ! As far as I can see, for the methods:

```ts
public async skipToNext(initialTime?: number);
public async skipToPrevious(initialTime?: number);
public async skip(index: number, initialTime?: number);
```

Internally you're just immediately seeking after you skip.. I think we might not actually have to modify these methods to take an initial time. It seems like we can already just do this directly from the JS side:

```ts
TrackPlayer.skipToNext()
TrackPlayer.seek(XXX)
```

jspizziri:
@dcvz that is true. I'm not entirely sure if their 100% equivalent due to the RN bridge execution time but I'm. No expert there. Doing it this way seems closer to the metal, but like I said, not an expert. What I will say is that it's definitely a nice convenience for what I believe is a very common use case, however if we don't want to support it I can understand that, and at that point my only potential concern would be the bridge execution time.

dcvz:
@jspizziri could you do some tests of this? I think it should work.

jspizziri:
@dcvz I'm not 100% sure how to set up a reliable test for something like this but I'll do some more research on the react native/js bridge and see what I can find.

Anecdotally, I had a project with RNTP before where seek after skip was working, but the audio would play from position 0 for a fraction of a second before the seek happened. It could've been related to the async loading issue, but that's the kind of behavior I'm concerned about.

In the meantime I'm going to pull out all the mergeable work from this PR so that it doesn't get held up and submit another PR for it.

dcvz:
Merged @jspizziri

jspizziri:
@dcvz lol, you merged that one so quick I thought there was a bug with GitHub because when I went to go look for it I didn't see it in open PRs so I created a duplicate. 🤣

jspizziri:
Just pushed a trimmed version of this PR rebased on the updates in `main`.

jspizziri:
So the fundamental question is: 

```ts
await TrackPlayer.skip();
// Can there ever be a significant delay between the execution of the above and below calls 
// which would result in them being executed in the native modules with enough delay to
await TrackPlayer.seek(x);
```

Alright, this is an answer I don't feel super confident in, but here goes:

Based on my (admittedly limited) understanding of [how the JS event loop works](https://developer.mozilla.org/en-US/docs/Web/JavaScript/EventLoop) in combination with some of [the documentation on RN Performance](https://reactnative.dev/docs/performance), my understanding of RN's use of their JS threads for execution, and my understanding around how JS fundamentally async and non-blocking architecture works.

It seems entirely theoretically possible that a back to back call

jspizziri:
So the fundamental question is: 

```ts
await TrackPlayer.skip();
// Can there ever be a significant delay between the execution of the above and below calls 
// which would result in them being executed in the native modules with enough delay to
// cause a noticeable stutter in playback for a user.
await TrackPlayer.seek(x);
```

This is an answer I don't feel super confident in, but here goes (I can't find much canonical information to make this more definitive):

TL;DR; based on my limited understanding, I believe it's theoretically possible.

Based on my (admittedly limited) understanding of [how the JS event loop/queue works](https://developer.mozilla.org/en-US/docs/Web/JavaScript/EventLoop) in combination with some of [the documentation on RN Performance](https://reactnative.dev/docs/performance), my understanding of RN's use of their JS threads for execution (specifically the use of a single thread for business logic execution), and my understanding around how JS fundamentally async and non-blocking architecture works.

It seems entirely theoretically possible that a back to back calls could get placed on the JS execution queue with several other calls/operations getting queued between them (e.g things happening asynchronously elsewhere in the app). If there are several operations that require some noticeable amount of time to process that get queued between the two calls then the native side will start playing and then execute the seek once the other queued JS operations have been processed, thus resulting in a delay from the users perspective.

If the above understanding is correct it could be very challenging (at least for someone with my current level of understanding) to contrive a reproducible example where this is occurring, as it would be depending upon so many other factors of execution in the application & subsequently the JS thread in RN.

A more definitive answer would require input from someone who is far more deeply versed and experienced in the JS runtime and the React Native bridge than I am or will be able to become on a reasonable timeline for the purposes of this PR. Perhaps someone from the community can chime in.

Thoughts?

dcvz:
Overall it sounds completely plausible and even expected that we cannot guarantee the calls run right after each other.. but I think fundamentally that does not change whether or not it would work.

I guess what might be possible as you mentioned is that maybe you hear audio for a bit before it actually seeks. But I'd honestly say, let's try to example project and attempt to skip then seek on it and see if we experience this. This might still give us an idea of how "likely/often/easily" this happens, or if it might just be in case we get some bigger call in between the two that causes a bigger delay before we seek.

Down the line this is something that the TurboModules changes might actually help us fix.

dcvz:
We might also be able to ask the people on that feature request thread to give this a spin and see how it performs in their applications, so we have a concrete problem to solve and that we know is occurring regularly.

jspizziri:
@dcvz I'm okay with that approach given your reluctance to change the API.

I'm going to upgrade to `v2.2.0-rc1` and see if I notice anything in the app that we're working on now.

I also think that we can safely close https://github.com/DoubleSymmetry/react-native-track-player/pull/713 at this point which will hopefully push people over to this MR and may help with getting feedback from folks. I'll let people on that issue know and perhaps you can close it.

dcvz:
I'm definitely reluctant to change API's as those need to be maintained in the future. But that doesn't mean we may not have to make changes, I'd first just like to see if it ends up being an issue in practice first.

That sounds like a good plan of action. I'll close it some time after your post.

jspizziri:
@dcvz alright here's what I found after testing my app:

```ts
await TrackPlayer.skipToNext();
await TrackPlayer.seekTo(x);
```

This approach is consistently causing the stutter that I feared. The issue presented itself consistently (on iOS) when one track was playing and then I switched to another. When playing from cold start it seemed to work fine.

I then installed my forked version locally and attempted the `TrackPlayer.skiptoNext(initialTime)` resulted in everything working as I would hope (no stutter). Obviously, this is only one data point, but I did run multiple tests and tried to isolate as many variables as possible. 

While it's possible that I'm getting _lucky_ in the tests with `skipToNext(initialTime)` and it happens to work I think that's unlikely given the perfect consistency over several tests. All in all, my tests do seem to indicate that the theory bears out and is actually far easier to reproduce than I had anticipated.

I know that it's not the result you were hoping for, but I promise I'm not intentionally trying to skew the results in my favor 😄 . I do know it's just one data point though.

jspizziri:
@dcvz 

Here's what I've found in my project after around 2hrs of testing different scenarios.

## Things That Work

```ts
await TrackPlayer.skip(0, initialTime);
await TrackPlayer.play();
```

```ts
await TrackPlayer.skip(0);
await TrackPlayer.seekTo(initialTime);
await TrackPlayer.play();
```

## Things That Don't

Stutters almost every time when skipping between tracks.

```ts
await TrackPlayer.skip(0);
await TrackPlayer.play();
await TrackPlayer.seekTo(initialTime);
```

dcvz:
Sounds like it won't really work the way I want it too 🥲 We may have to incorporate something like this in the end..

I'll do some thinking as well wether we'd be able to make it work. But I think probably not

dcvz:
@jspizziri I actually have one more request. Could we make this PR against the `kotlinaudio` branch?

Have you already been trying to use the new betas/rc? We're getting ready to release that soon so it'd be great to have this there. Android side would have to be remade otherwise.

jspizziri:
@dcvz , I'll do that today.

I only started using the 2.2.x version yesterday since that's where the SwiftAudioEx patch landed.

jspizziri:
@dcvz I pushed up a function version of the kotlin implementation.

As you're looking at it there were a few questions I had:

1. I wasn't sure if you wanted to leave the `seekTo(initialTime)` behavior at the module level, or if you wanted to push that into the respective service layer for each `skip` call. LMK and I can adjust that (for both iOS and Android).
2. I wasn't sure how the player behaved if a user called `skipToNext` but there wasn't anything else in the queue. It doesn't seem like there's a success or failure response sent back. One potential concern I have is if they `skipToNext(initialTime)` but there's nothing in the queue. If it's a silent failure the current implementation would presumably result in the seeking to the track they're on currently.

meetwithtaha:
Unable to remove tracks when screen unmount. it only removes when app restart after kill. @dcvz

jspizziri:
> Unable to remove tracks when screen unmount. it only removes when app restart after kill. @dcvz 

@meetwithtaha is this relevant to this PR in some way? It doesn't seem to be relevant at a glance so can you please explain in more detail?

jspizziri:
@dcvz I know the world and Europe are crazy right now, so I hope all is well with you. Assuming that everything ok I just wanted to ping you on this PR.

dcvz:
Thanks for following up again and being patient @jspizziri. It's been a bit hectic for us in the past weeks! I'll follow up on this later today or tomorrow morning/

dcvz:
Didn't get a chance yet @jspizziri but have not forgotten :) hoping to tomorrow.

jspizziri:
Thanks @dcvz , appreciate you keeping me in the loop!

martinmidtsund:
You asked for feedback from the other PR, sorry I'm a bit late, hectic here as well! 😅  So this is only input, I have not been able to test the code yet, sorry.

If this now works for both platforms, as you mention @jspizziri :

```
await TrackPlayer.skip(0);
await TrackPlayer.seekTo(initialTime);
await TrackPlayer.play();
```

Then that is sufficient for me, then we can use this to implement the same functionality in our app as `skip(0, initialTime)` would, wouldn't we? By listening to `Event.PlaybackTrackChanged` and seekTo if initialTime is set on the track? At least it is sufficient for our use case right now, though this will perhaps change in the future.

Great work!

jspizziri:
@dcvz just wanting to check in on this. Thanks!

jspizziri:
Just pinging this MR. Thanks!

dcvz (approved):
LGTM 👍 Thanks for this @jspizziri !
Sorry for the longer wait

dcvz:
@jspizziri looks like we need one rebase, then I'll merge it in! Thanks a lot and sorry for the delay

jspizziri:
@dcvz great news! And no worries at all, I know how it can go sometimes.

I just rebased and pushed, I'm going to run one more build & test just to make sure all is well and I'll let you know.

jspizziri:
@dcvz alrighty everything looks good from my end. Let me know if you need anything else from me.

jspizziri:
@dcvz just wanted to bump this a bit. Everything should be good to go!

dcvz:
Finally merged @jspizziri ! Thanks a lot for this and your patience :)
I think in a different thread you expressed your interest in being a contributor and I see there's already a problem you want to tackle. Feel free to reach out to me via Discord or email - david [at] doublesymmetry (com) and we can find a way to involve you!

PR #1423, opened 2022-03-01
Copy doc strings to main api functions

alpha0010:
Documentation comments makes code completion in TypeScript editors nicer.


mpivchev (approved):
Thanks :)

PR #1427, opened 2022-03-03
Don't round times to integer

jenshandersson:
We need to get the accurate position, not just integers. And I don't see any reason to round it to ints.

### DIFF
--- a/android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
+++ b/android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
@@ -224,7 +224,7 @@ class MusicService : HeadlessJsTaskService() {
     }
 
     fun seekTo(seconds: Float) {
-        handler.post { player.seek(seconds.toLong(), TimeUnit.SECONDS) }
+        handler.post { player.seek((seconds * 1000).toLong(), TimeUnit.MILLISECONDS) }
     }
 
     fun getCurrentTrackIndex(callback: (Int) -> Unit) {
--- a/android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
+++ b/android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
@@ -212,7 +212,7 @@ class MusicService : HeadlessJsTaskService() {
     }
 
     fun skip(index: Int) {
-        handler.post { player.jumpToItem(index) }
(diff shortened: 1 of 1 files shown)


jenshandersson:
I just realised your KotlinAudio library only takes integers as well, what's the reasoning for removing this much accuracy in a audio library?
https://github.com/DoubleSymmetry/KotlinAudio/blob/002f10c931ba7f34d0c2b7b59b2fb445919f649c/kotlin-audio/src/main/java/com/doublesymmetry/kotlinaudio/players/BaseAudioPlayer.kt#L192

mmmoussa:
I tested this PR in my app and it is working as desired.

android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
```diff
@@ -224,7 +224,7 @@ class MusicService : HeadlessJsTaskService() {
     }
 
     fun seekTo(seconds: Float) {
-        handler.post { player.seek(seconds.toLong(), TimeUnit.SECONDS) }
+        handler.post { player.seek((seconds * 1000).toLong(), TimeUnit.MILLISECONDS) }
```
mpivchev, line 227: This can stay the same, no? We are still seeking only to seconds.

mmmoussa: Why do you want to only seek to seconds? In my use case I would like to be able to seek to milliseconds.

mmmoussa: Since the technical support for it is already here, it doesn't make sense to block off reasonable use cases unless there is a tradeoff I'm unaware of.

mpivchev: We are fine with seeking to milliseconds. I'm just wondering if we should not take milliseconds as parameter instead of seconds as `Float`

mpivchev:
We had an internal discussion about this. We will keep the parameter as `seconds: Float` as that is how the old android version used to work, and that is how the API is currently documented to work as. In V3 of RNTP, we will introduce milliseconds as a parameter for those that want to use that instead of decimal seconds.

Since this PR fixes the rounding, which was a separate problem from this, I will approve it.
Thanks for contributing 😇

PR #1436, opened 2022-03-14
Some small refactoring/add a TODO

mpivchev:


android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
```diff
@@ -28,7 +28,9 @@ import java.util.concurrent.TimeUnit
 class MusicService : HeadlessJsTaskService() {
     private lateinit var player: QueuedAudioPlayer
     private val handler = Handler(Looper.getMainLooper())
+
     var stopWithApp = false
+        private set
```
mpivchev, line 33: I think it's good for this to only be settable inside the `MusicService`, as `updateOptions` is the function that is supposed to change this flag.

android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
```diff
@@ -438,9 +440,10 @@ class MusicService : HeadlessJsTaskService() {
         return MusicBinder()
     }
 
-    fun destroyIfAllowed(ignoreStopWithAppSetting: Boolean = false) {
+    // TODO: #AEX-45 forceDestroy is needed when calling destroy() manually. Find an alternative solution that does not require a second flag.
```
mpivchev, line 443: Added a todo and a separate internal ticket for this, since I think there can be a better solution for this :)

PR #1441, opened 2022-03-22
Add missing promise resolve in iOS destroy method

mmmoussa:
Fixes https://github.com/DoubleSymmetry/react-native-track-player/discussions/1264#discussioncomment-2407544 on iOS.

Before the change:
`await TrackPlayer.destroy()` never resolves

After the change:
`await TrackPlayer.destroy()` resolves and following code continues execution

### DIFF
--- a/ios/RNTrackPlayer/RNTrackPlayer.swift
+++ b/ios/RNTrackPlayer/RNTrackPlayer.swift
@@ -293,6 +293,7 @@ public class RNTrackPlayer: RCTEventEmitter {
         self.player.nowPlayingInfoController.clear()
         try? AVAudioSession.sharedInstance().setActive(false)
         hasInitialized = false
+        resolve(NSNull())
     }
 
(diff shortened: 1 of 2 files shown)


mpivchev:
@dcvz On Android we are also not resolving the callback:

```kotlin
    @ReactMethod
    fun destroy(callback: Promise) {
        if (verifyServiceBoundOrReject(callback)) return

        unbindService()
        musicService.destroyIfAllowed(true)
    }
```

Is this something we missed?

dcvz:
@mpivchev it does seem like something we missed. Looks like we only added a rejection but not completion.

dcvz:
@mmmoussa this is against the wrong branch -- you want to target the `kotlinaudio` branch

jspizziri:
I created an PR that adds the outstanding android fix and is based off of kotlinaudio https://github.com/doublesymmetry/react-native-track-player/pull/1472.

PR #1456, opened 2022-03-30
feat(android): move the jcenter dependency into the android library t…

jspizziri:
…o prevent installation issues

https://github.com/DoubleSymmetry/react-native-track-player/issues/1235#issuecomment-921266932

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -46,6 +46,11 @@ repositories {
 
     mavenCentral()
     google()
+
+    // FIXME: this can be removed once exoplayer has been migrated to a maven
+    // release. See more details here:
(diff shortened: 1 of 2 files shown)


jspizziri:
@mpivchev I think this MR should supersede this one: https://github.com/DoubleSymmetry/react-native-track-player/pull/1335

LMK what you think.

dcvz:
@jspizziri does the `kotlinaudio` branch require `jcenter()`? I think this is only a problem for `main` currently since its based on an old release only available in `jcenter()`

jspizziri:
@dcvz good question. I _thought_ I had run into this while using the `kotlinaudio` branch but let me check again and make sure. I'll let you know what I find.

jspizziri:
@dcvz I just double checked and you're correct. I must've been switching between branches when I ran into this or something.

PR #1457, opened 2022-03-30
Replace Handler with Main scope

dcvz:
Replaces our Handler usage with the existing `MainScope` -- we were already mix and matching, so this goes for just one.


mpivchev (approved):
LGTM! I don't remember if there were issues when using a scope instead of handler with ExoPlayer, but if everything works fine, let's go with it.

PR #1458, opened 2022-03-30
Feature in v1 to declare whether or not to erase notifications on each init

makgabri:
Added support in 1.2.7 to declare whether or not you want to erase all notifications.


jspizziri:
@makgabri , v1 is no longer supported. Please upgrade to v2. As a result we are not currently accepting feature additions for v1.

PR #1464, opened 2022-04-07
Update installation troubleshooting

lostchopstik:
* Added step from issue #1235 to resolve issue on RN 0.65+


jspizziri:
@lostchopstik this issue is actually resolved in v2.2.0. Because a release is very close and all energies are devoted to that effort this isn't going to get merged (again, its not an issue in that release).

Issue #1469, opened 2022-04-13
Artist should not be required for TrackPlay functionality on iOS


JDMathew:
**Describe the Bug**
It seems on iOS artist is required when adding a track otherwise `await TrackPlayer.add(tracks)` will not resolve. This is not the case on Android. According to the Type artist is not required. If artist is required this is undocumented. 

We could update the docs to make it required on iOS but in some cases you may not know the artist and so it would be better if the underlying issue is fixed. 

Please see reference to Track Base Type here:
https://github.com/doublesymmetry/react-native-track-player/blob/0af6e2f7b8f4935c0464195d893e61b03d619097/src/interfaces.ts#L202-L213

**Steps To Reproduce**
How can someone else reproduce this bug?
Add a Track without artist on iOS - promise will not resolve

**Environment Info:**
running `react-native-track-player` version `2.1.3`
I this is an issue on `2.2.0-rc2` as well

Linked issue confirming behaviour: #1468

jspizziri:
@JDMathew thanks for the report! Would you be able to take a crack at fixing the issue?

jspizziri:
@JDMathew the fix was pretty easy so I went ahead and implemented it.

JDMathew:
@jspizziri, thanks for adding a PR fixing this. I left a [comment](https://github.com/doublesymmetry/react-native-track-player/pull/1481#issuecomment-1104253728) on your PR just letting you know that title should also not be required according to the Track interface. https://github.com/doublesymmetry/react-native-track-player/blob/90fc0e76b47a741b5fed04846fff3c89913ed66d/src/interfaces.ts#L202-L213

jspizziri:
@JDMathew this fix just got merged. I'm going to close this. It will be available in the next release (likely `2.2.0-rc4`).

PR #1481, opened 2022-04-19
fix(ios): fix an issue on iOS which required both artist & title to be provided on tracks

jspizziri:
According to the interface `artist` should not be required on a `Track` object. This PR updates this on iOS to make the interface uniform and in doing so closes https://github.com/doublesymmetry/react-native-track-player/issues/1469


JDMathew:
Looks good to me.

jspizziri:
@JDMathew pushed an update. Feel free to take it for a spin to verify it resolves your issue. You can test with the example project (make sure you run `yarn add react-native-track-player@file:..`).

dcvz (approved):
LGTM 👍

PR #1487, opened 2022-04-22
Documentation Update

jspizziri:


dcvz:
@jspizziri looks great! Let's point this to `main` though -- we'll rebase the `kotlinaudio` branch against it before the next release.

jspizziri:
@dcvz repointed at `main`. I also added two more commits; one about contributing (which probably needs to most feedback) and another updating the Discord info.

PR #1492, opened 2022-04-24
docs(workflow): update the build directory as it isn't contextualized…

jspizziri:
… by the working-directory

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -36,7 +36,7 @@ jobs:
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           # Build output to publish to the `gh-pages` branch:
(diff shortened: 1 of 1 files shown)


jspizziri:
@dcvz hopefully this will fix the the docs publish.[ It looks like it's the problem based on this log](https://github.com/doublesymmetry/react-native-track-player/runs/6143136340?check_suite_focus=true#step:6:61). I had assumed that that command would also be contextualized by the `working-directory` that is set in the action, but it doesn't look like it is.

dcvz (approved):

dcvz:
@jspizziri I also had to change some settings in the repo. Looks like this pushes to a deviated gh-pages branch so we have to set that as the source + ask it to read the root of the repo and not the docs folder.

It works now!

PR #1479, opened 2022-04-18
Warn about potential crash in "Getting Started"

austinried:
I encountered a crash recently in the latest `v2.2.0-rc3` related to calling a method on `TrackPlayer` before setup was complete and thought it should maybe be documented more clearly.  The details are in #1475 but the gist is I had a race condition between calling `getQueue` and `setupPlayer`, and that didn't used to crash the app but now it does and that seems to be intentional.

Not super sure on the wording here, but I thought it would make sense to flip the order of what's "first" since you always want to register the playback service, and then calling `setupPlayer` comes later.

PR #1494, opened 2022-04-24
docs(guides): add some guides for common usecases of the app

jspizziri:


dcvz:
@jspizziri looks like PR isn't done yet but seems we're adding docs for things only in 2.2.0 -- we'll have to wait until we've done the versioning split off I mentioned in #1479

jspizziri:
For sure, I'm mostly just writing the guides right now. The necessary functionality hasn't even been merged haha. I just want a pr out there so I don't forget where I put it.

I was thinking that I can set up versioning as part of my progress event PR and then this could be rebased on that before merging.

jspizziri:
Btw, it automatically requests a review from you when a pr is submitted, and you have to submit it before you can mark it as a draft, so that's why that happened that way.

dcvz:
It didn't tag me. I just keep an eye on PR's now that there's not so many 😁

PR #1495, opened 2022-04-24
Docs/versioned

jspizziri:
This PR is based off of https://github.com/doublesymmetry/react-native-track-player/pull/1473 and contains versioned documetation for 2.1.0 and 2.2.0, in addition it adds some common Guides for the functionality added there.


jspizziri:
Note: this shouldn't be merged until the aforementioned PR is finalized and merged and depending on what happens it may need to be rebased.

docs/docs/guides/offline-playback.md
```diff
@@ -0,0 +1,41 @@
+---
+sidebar_position: 1
+---
+
+# Offline Playback
```
dcvz, line 5: Not a big deal but this could exist also in 2.1.X

dcvz (commented):
Honestly now I wished we'd done this as a stand-alone against `kotlinaudio` -- would allow us to merge now without waiting for Android approval of your progress feature -- then rebase the progress branch.

Now this or the progress PR will have a lot of code changes not necessarily tied to the feature.

Would you be ok with doing that? It might be easy to just cherry pick the last two commits. We can keep the guides there.. and only add the progress one as part of your progress PR.

Changes themselves look good though. Perhaps we want to name it `2.1.3` after the latest release instead of just `2.1.0`. Let me know what you think about the above.

jspizziri:
Ah yea, that would've been better! Have a head cold giving me all these great ideas.

I'll move it over in a way that makes more sense.

RE: the version - patch versions shouldn't really have docs that are different from one another (which is why I did 2.1.0). I'd recommend: `2.1`, `2.1.x`, or `2.1.0`. Thoughts?

dcvz:
2.1 would also be ok :) I know docs don't change much across versions but at least the user would feel like they're looking at the docs for the latest version they might use

PR #1496, opened 2022-04-24
docs(2.2): add versioned docs for 2.2 release

jspizziri:


jspizziri:
@dcvz here you go, this should just be the isolated versioning of 2.1/2.2.

jspizziri:
Also note, I made the most current version `2.2.0-rc3` to make it clear that 2.2.0 is not officially released, but this is something that should be updated to `2.2` eventually if we want to keep it consistent.

jspizziri:
I take it back, this won't get deployed into main (or the github pages site) until 2.2.0 is officially released based on how we're doing it. So I'm leaving it as-is.

dcvz (approved):
LGTM :+1:

dcvz:
@jspizziri one small note though.. I do actually want to merge `kotlinaudio` into `main` after the doc split so both documentations are uploaded. However, as far as I understand from the docosaurus docs the android rewrite should be published under /next, no? Therefore signifying that it's the development version

jspizziri:
https://github.com/doublesymmetry/react-native-track-player/pull/1497

PR #1498, opened 2022-04-25
Warn about potential crash in "Getting Started"

austinried:
Same changes as  #1479 but updated for the new docs format.

### DIFF
--- a/docs/docs/basics/getting-started.md
+++ b/docs/docs/basics/getting-started.md
@@ -4,19 +4,8 @@ sidebar_position: 2
 
 # Getting Started
 
-First of all, you need to set up the player. This usually takes less than a
-second:
-
-```ts
-import TrackPlayer from 'react-native-track-player';
-
-await TrackPlayer.setupPlayer({})
-// The player is ready to be used
-```
-
-You also need to register a [playback service](./playback-service.md) right after
-registering the main component of your app:
-
+## Starting off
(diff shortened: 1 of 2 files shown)


docs/docs/basics/getting-started.md
```diff
@@ -4,19 +4,8 @@ sidebar_position: 2
...
-
-await TrackPlayer.setupPlayer({})
-// The player is ready to be used
-```
-
-You also need to register a [playback service](./playback-service.md) right after
-registering the main component of your app:
-
+## Starting off
+First, you need to register a [playback service](#playback-service) right after registering the main component of your app (typically in your `index.js` file at the root of your project):
```
jspizziri, line 8: I believe we should be pointing at `./playback-service.md` instead of `#playback-service`.

jspizziri (changes requested):

docs/docs/basics/getting-started.md
austinried: Ah forgot about that, I'll switch it back

jspizziri:
@austinried this lgtm. My only other thought is that technically this applies to 2.1 too which would be updated in the versioned_docs folder.Would you be able to do that too?

austinried:
@jspizziri no problem, just pushed that change.

jspizziri (approved):
@dcvz LGTM. You'll need to merge.

PR #1512, opened 2022-04-26
chore(example): upgrade example project to react-native@0.68.1

jspizziri:


dcvz:
@jspizziri haven't fully looked yet but seems like we have to update the CI workflow to make them build the app

jspizziri:
@dcvz I updated the workflow to use node 16 which is currently in LTS. The next LTS is 18 which was released earlier this month so I didn't feel like upgrading to that one yet (I haven't even used it yet).

You'll notice I also migrated the example tracks to the docs website as Google Drive has an access limit which causes issues during testing.

dcvz (approved):
LGTM 👍 great idea to move the tracks to an actual host!

Issue #1489, opened 2022-04-22
[Feature Request]  Add  Search To Documentation Site


jspizziri:
Once [the docusaurus refactor launches](https://github.com/doublesymmetry/react-native-track-player/pull/1488) we just need to do this:

https://docusaurus.io/docs/search#using-algolia-docsearch

jspizziri:
I've submitted the application with DocuSearch. Now we wait.

PR #1520, opened 2022-04-27
docs(search): enable search 🚀

jspizziri:
https://github.com/doublesymmetry/react-native-track-player/issues/1489

### DIFF
--- a/docs/docusaurus.config.js
+++ b/docs/docusaurus.config.js
@@ -130,6 +130,12 @@ const config = {
         theme: lightCodeTheme,
         darkTheme: darkCodeTheme,
       },
+      algolia: {
+        appId: 'CYSR91W7SU',
+        apiKey: 'a5ab4af3642fdd7fbc07030593017797',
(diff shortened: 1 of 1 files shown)


docs/docusaurus.config.js
```diff
@@ -130,6 +130,12 @@ const config = {
         theme: lightCodeTheme,
         darkTheme: darkCodeTheme,
       },
+      algolia: {
+        appId: 'CYSR91W7SU',
+        apiKey: 'a5ab4af3642fdd7fbc07030593017797',
```
dcvz, line 135: Is there any danger to including these hard coded? Should they maybe be injected via ENV variables?

jspizziri: They're all public keys so there's no security risk at all. They make that clear in the docs. If people took them and added them to their site all they would be doing is adding search for our docs on their site.

I took a look at other projects using this setup and they didn't bother to use env either.

Issue #1445, opened 2022-03-28
Android `TrackPlayer.reset()` resolving too quickly


JDMathew:
**Describe the Bug**
There seems to be an issue with `TrackPlayer.reset()` resolving it's promise too quickly. 

According to documentation `reset()`:
> Resets the player stopping the current track and clearing the queue.

This means the queue should be empty after calling `TrackPlayer.reset()` i.e `getQueue() // should retrun []`

After adding a track to the queue, If I run the following:
``` javascript
const resetTrack = async () => {
    try {
      await TrackPlayer.reset()
      const queue = await TrackPlayer.getQueue()
        console.log('queue', queue)
    } catch (error) {
      console.log('Error from reset', error)
    }
  }
```

No error is thrown, however:

```
console.log('queue', queue) // returns the queue and not an empty array []
```

If I instead, force a delay before calling `getQueue()` then the correct result is returned:

For example:

``` javascript
 const resetTrack = async () => {
    try {
      await resetPlayBack()
      setTimeout(async () => {
        const queue = await TrackPlayer.getQueue()
        console.log('queue', queue)
      }, 100)
    } catch (error) {
      console.log('error', error)
    }
  }
``` 

Then
```
console.log('queue', queue) // returns an empty array []
```

Furthermore it seems strange to me that after resetting `getCurrentTrack()` would be 0 and that the playbackState is left in a paused state (its previous state) and not None. I.e:
``` javascript
const currentTrackIndex = await TrackPlayer.getCurrentTrack() // 0
await TrackPlayer.getTrack(currentTrackIndex ?? 0) // null
const playBackState = usePlaybackState() // 3 (Paused)
```

Surely `playBackState` should be None (i.e enum of 0) and not Paused (3). If it is Paused then it would appear that a track can be played but when running `play()` nothing will happen. `play()` also silently fails even though there is no track for it to run i.e `getTrack()` is null but `await TrackPlayer.play().catch(console.log)`, the promise resolves but nothing is played. instead of being rejected. Is this the desired behaviour?

**Environment Info:**
This issue was found on react-native-track-player `^2.2.0-rc2`. 
```console
System:
    OS: macOS 12.2.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 178.75 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.14.0 - ~/.nvm/versions/node/v16.14.0/bin/node
[...]

Issue #1515, opened 2022-04-27
TrackPlayer.add return type is incorrect


josemiguel-alvarez:
**Describe the Bug**
TrackPlayer.add return type is `Promise<void>`.

Link to the TrackPlayer.add function: https://github.com/doublesymmetry/react-native-track-player/blob/3ee71f1d8584a1c33b3fc8a418cb7c1de4423a0e/src/trackPlayer.ts#L82

It should be a number as specified in the documentation.

Link to the documentation: https://react-native-track-player.js.org/docs/api/functions/queue#addtracks-insertbeforeindex

jspizziri:
@josemiguel-alvarez would you like to submit an MR for this?

josemiguel-alvarez:
> @josemiguel-alvarez would you like to submit an MR for this?

Sure! I'll open it today.

jspizziri:
Looking at this a little more deeply it looks like the Kotlin service needs to be updated too.

jspizziri:
@josemiguel-alvarez 

This will need to be updated in the [`MusicModule.kt`](https://github.com/doublesymmetry/react-native-track-player/blob/main/android/src/main/java/com/doublesymmetry/trackplayer/module/MusicModule.kt#L287) on lines `287` & `293`. Please [reference how iOS is behaving](https://github.com/doublesymmetry/react-native-track-player/blob/main/ios/RNTrackPlayer/RNTrackPlayer.swift#L354)

jspizziri:
@josemiguel-alvarez I ended up taking this one on. I noticed [this issue in the process](https://github.com/doublesymmetry/react-native-track-player/issues/1522) which we'll address soon.

josemiguel-alvarez:
> @josemiguel-alvarez I ended up taking this one on. I noticed [this issue in the process](https://github.com/doublesymmetry/react-native-track-player/issues/1522) which we'll address soon.
> 
> [Here's the PR](https://github.com/doublesymmetry/react-native-track-player/pull/1523)

Thank you Jacob, I haven't had time yet to work on this.

PR #1523, opened 2022-04-28
fix(android): make the TrackPlayer.add method consistent across JS, i…

jspizziri:
…OS, and Android

https://github.com/doublesymmetry/react-native-track-player/issues/1515


jspizziri:
@dcvz this should be ready for your review. In addition to moving where the `scope.launch` is I removed all the callback params from the MusicService since they _can_ all be sync AFAIK.

android/src/main/java/com/doublesymmetry/trackplayer/module/MusicModule.kt
```diff
@@ -303,255 +321,295 @@ class MusicModule(reactContext: ReactApplicationContext) : ReactContextBaseJavaM
...
+
                 for (track in trackList) {
                     val index = if (track is Int) track else track.toString().toInt()
 
                     // We do not allow removal of the current item
-                    if (index == it) {
+                    if (index == currentIndex) {
                         Timber.e("This track is currently playing, so it can't be removed")
-                        return@getCurrentTrackIndex
+                        continue
```
jspizziri, line 334: Not sure if this should be a `continue` or a `break`

jspizziri: Based on iOS a continue is correct:

https://github.com/doublesymmetry/react-native-track-player/blob/main/ios/RNTrackPlayer/RNTrackPlayer.swift#L366

jspizziri:
I believe this PR also fixes https://github.com/doublesymmetry/react-native-track-player/issues/1445

jspizziri:
I retriggered the iOS build this morning and it's working. Must've been an external dependency issue with the build process.

jspizziri:
Fixing conflicts here. Do not merge.

jspizziri:
Alright, this has been rebased on `main` and the android tests are passing. The iOS build is failing due to a `429 (Too Many Requests` error when attempting to clone a dependency from git. Probably because of how many builds we're running. If this gets triggered at a later point it should succeed.

dcvz:
Thanks @jspizziri haven't had a chance to look yet! But this is something I can take over if it goes past today :)

dcvz:
Just updating here: It's a busy week and I'm the only RNTP atm due to vacations and other absences so things will be a bit slower these weeks.

mpivchev (approved):
LGTM

android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
```diff
@@ -51,7 +51,8 @@ class MusicService : HeadlessJsTaskService() {
     private var notificationCapabilities: List<Capability> = emptyList()
     private var compactCapabilities: List<Capability> = emptyList()
 
-    fun setupPlayer(playerOptions: Bundle?, promise: Promise?) {
+    @MainThread
```
mpivchev, line 54: All Android services run in the main thread, so this is not needed I think. Was there a specific reason for these annotations?

mpivchev (changes requested):

android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
jspizziri: I added these at the request of @dcvz , I believe the thought was to notify people looking at the code that those methods would be on the main thread.

Happy to remove if you give me the clearance.

android/src/main/java/com/doublesymmetry/trackplayer/module/MusicModule.kt
```diff
@@ -230,8 +240,10 @@ class MusicModule(reactContext: ReactApplicationContext) : ReactContextBaseJavaM
     fun destroy(callback: Promise) {
         if (verifyServiceBoundOrReject(callback)) return
```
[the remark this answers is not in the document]
jspizziri: @mpivchev do you have what judgment should I use to make that decision? Happy to do it, but not sure how I should determine what other ones this should be done to, or is it just this one? The logic I used in the modifications here was to wrap the smallest amount of code possible in the "async" `scope.launch`.

android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
dcvz: so we're no longer using `scope.launch` inside the Service but rather at the Module level -- so annotating them with `@MainThread` is just so the IDE knows that when the Module calls them.. it should be within a main thread scope.

dcvz: The reason for this is that at the module level if you did something like:
```kotlin
// pseudo code
player.add(tracks) // this gets called at the service level in a coroutine.. so the next line will be called immediately without actually waiting for the add method to be called.. 
promise.complete()
```

android/src/main/java/com/doublesymmetry/trackplayer/module/MusicModule.kt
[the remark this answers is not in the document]
dcvz: `verifyServiceBoundOrReject` is a simple boolean check so its no big deal to put it in the scope

jspizziri:
@dcvz @mpivchev I've just pushed a new commit updating the `= scope.launch`. There are a few instances I didn't change that based on the judgment that could've been. The reason I didn't is that those are all `override` functions and there were some issues with the return types which with my poor kotlin-foo I didn't have a clear answer for so I took the path of least resistance.

LMK your thoughts. I'll squash once we have an approved version.

android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
```diff
@@ -223,113 +220,126 @@ class MusicService : HeadlessJsTaskService() {
...
-    fun getCurrentTrackIndex(callback: (Int) -> Unit) = scope.launch {
-        callback(player.currentIndex)
-    }
+    @MainThread
+    fun getCurrentTrackIndex(): Int = player.currentIndex
 
-    fun getRate(callback: (Float) -> Unit) = scope.launch {
-        callback(player.playbackSpeed)
-    }
+    @MainThread
```
mpivchev, line 298: I still think we should remove these here, since services always run on main thread 
cc @dcvz

mpivchev (changes requested):

android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
jspizziri: @dcvz this is the only thing hanging up this PR. Would you be able to take a look?

jspizziri: @mpivchev technically someone modifying the codebase could call these service methods from any thread they wanted to when extending the code. I think these annotations are a good safety in the event that that happens.

My 2 cents is to merge this and be done with it.

dcvz (approved):
LGTM 👍

Issue #1511, opened 2022-04-26
Lockscreen controls shows artwork of previous track, if current track has empty artwork


Strate:
**Describe the Bug**
If current track has no artwork, and previous track contained any artwork - lockscreen controls shows artwork of previous track

**Steps To Reproduce**
Add two track in queue, first with artwork, second without artwork

**Environment Info:**
```
System:
    OS: macOS 12.3.1
    CPU: (10) arm64 Apple M1 Max
    Memory: 4.16 GB / 64.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 14.19.1 - /opt/homebrew/bin/node
    Yarn: 1.22.18 - /opt/homebrew/bin/yarn
    npm: 8.5.5 - /opt/homebrew/bin/npm
    Watchman: 2022.03.21.00 - /opt/homebrew/bin/watchman
  Managers:
    CocoaPods: 1.11.3 - /opt/homebrew/bin/pod
  SDKs:
    iOS SDK:
      Platforms: DriverKit 21.4, iOS 15.4, macOS 12.3, tvOS 15.4, watchOS 8.5
    Android SDK: Not Found
  IDEs:
    Android Studio: 2021.1 AI-211.7628.21.2111.8193401
    Xcode: 13.3/13E113 - /usr/bin/xcodebuild
  Languages:
    Java: 11.0.14.1 - /usr/bin/javac
  npmPackages:
    @react-native-community/cli: Not Found
    react: 17.0.2 => 17.0.2
    react-native: https://github.com/chatium/react-native/archive/refs/tags/chatium-18.tar.gz => 0.66.3
    react-native-macos: Not Found
  npmGlobalPackages:
    *react-native*: Not Found
```
_Paste the exact `react-native-track-player` version you are using:_ 2.1.3
_Real device? Or simulator?_ Both of them
_What OS are you running?_ Android & iOS

**How I can Help**

For android fix is simple:
```patch
diff --git a/node_modules/react-native-track-player/android/src/main/java/com/guichaguri/trackplayer/service/metadata/MetadataManager.java b/node_modules/react-native-track-player/android/src/main/java/com/guichaguri/trackplayer/service/metadata/MetadataManager.java
index 42dc212..b616ee8 100644
--- a/node_modules/react-native-track-player/android/src/main/java/com/guichaguri/trackplayer/service/metadata/MetadataManager.java
[...]

PR #1528, opened 2022-04-29
fix(ios, track): fix issues with the setting & unsetting of track …

jspizziri:
…artwork

allow for the setting of empty artwork, use else conditions in Track.getArtwork to prevent multiple
callbacks from firing

https://github.com/doublesymmetry/react-native-track-player/issues/1511


jspizziri:
This PR relies on [an underlying change being made to SwiftAudioEx](https://github.com/doublesymmetry/SwiftAudioEx/pull/17). Before it's complete SwiftAudioEx will need to release a new patch and that `podspec` in this repository will need to be updated to use that patched version.

bradleyflood (approved):
LGTM

jspizziri:
This should be ready to merge now.

Issue #1225, opened 2021-08-11
[iOS] unable to resume track after first play


AdamSheaHewett:
**Describe the bug**
On iOS, it seems that TrackPlayer.stop() clears the queue and also sets the state to State.None

I add a track as normal, and am able to play it, but the moment I use the stop() function, the queue is cleared, state is State.None, and the player just stops working.

**To Reproduce**
Using v2.0.0 on iOS, add a track, TrackPlayer.play(), TrackPlayer.stop(), then try to TrackPlayer.play() again. It won't work.

**Environment (please complete the following information):**
- `react-native-track-player` version 2.0.0
- Testing on iOS simulator
System:
    OS: macOS 11.5.1
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 233.77 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 14.15.4 - ~/.nvm/versions/node/v14.15.4/bin/node
    Yarn: 1.22.10 - ~/.nvm/versions/node/v14.15.4/bin/yarn
    npm: 6.14.10 - ~/.nvm/versions/node/v14.15.4/bin/npm
    Watchman: 2021.06.07.00 - /usr/local/bin/watchman
  Managers:
    CocoaPods: 1.10.1 - /usr/local/bin/pod
  SDKs:
    iOS SDK:
      Platforms: iOS 14.5, DriverKit 20.4, macOS 11.3, tvOS 14.5, watchOS 7.4
    Android SDK: Not Found
  IDEs:
    Android Studio: 2020.3 AI-203.7717.56.2031.7583922
    Xcode: 12.5.1/12E507 - /usr/bin/xcodebuild
  Languages:
    Java: 11.0.11 - /usr/bin/javac
  npmPackages:
    @react-native-community/cli: Not Found
    react: 17.0.2 => 17.0.2 
    react-native: 0.64.1 => 0.64.1 
    react-native-macos: Not Found
  npmGlobalPackages:
    *react-native*: Not Found

**Code**
Please, share the code that is causing the issue
```
// Play button
await TrackPlayer.add(url);
await TrackPlayer.play(); // Plays fine

// Stop button
await TrackPlayer.stop();
const tracks: Track[] = await playControls.getQueue();
console.log('queue', tracks); // []
```

This only happens on iOS

dcvz:
[...]

PR #1529, opened 2022-04-30
fix(ios, stop, event.playbackprogressupdated): make the iOS API the s…

jspizziri:
…ame as the Android API for stop

additionally, fix an issue with Event.PlaybackProgressUpdated causing exceptions when fired after
certain queue manipulations

https://github.com/doublesymmetry/react-native-track-player/issues/1225


bradleyflood (approved):
Makes sense to me.

dcvz (approved):

PR #1569, opened 2022-06-14
Add disabled option in hooks

minhchienwikipedia:
- In case we use `usePlayerProgress` or `usePlaybackState` in the list item, if we put these hooks in the item component it will rerender all of the items. So I just changed it to have an option `disabled` to set state only in case we want to get the value.
- And return `playerState` value in `usePlayerProgress` as well, we don't need to call two hooks to get this value, just return it.

### DIFF
--- a/src/hooks.ts
+++ b/src/hooks.ts
@@ -3,42 +3,54 @@ import { useEffect, useState, useRef } from 'react'
 import TrackPlayer from './trackPlayer'
 import { State, Event, ProgressState } from './interfaces'
 
-/** Get current playback state and subsequent updates  */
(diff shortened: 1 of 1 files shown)


jspizziri:
Closing, as without further explanation this PR seems counterproductive to me.

jspizziri:
You shouldn't be using useProgess in a list item like that. It only gives you the current progress of the currently playing track. It's really only suitable for the player interface. You should be caching your progress for each item and reading it in your list.

See here for more detail

https://react-native-track-player.js.org/docs/next/guides/saving-progress

Issue #1585, opened 2022-06-30
v2.2.0-rc4 introduces fatal crash on iOS when calling TrackPlayer.reset()


mmmoussa:
**Describe the Bug**
Calling `TrackPlayer.reset()` while playing a track on iOS causes a fatal crash in v2.2.0-rc4. Crash was not present on v2.2.0-rc3 or below.

**Steps To Reproduce**
1. Add a track to the queue
2. Start playing the track
3. Call `TrackPlayer.reset()`

**Code To Reproduce**
```
await TrackPlayer.add(listOfTracks);
await TrackPlayer.play();
await TrackPlayer.reset();
```

**Environment Info:**
Tested on iOS simulator with RNTP version v2.2.0-rc4 and RN version v0.68.2

**How I can Help**
Dug into this with the old and new version. The crash happens on https://github.com/doublesymmetry/react-native-track-player/blob/main/ios/RNTrackPlayer/RNTrackPlayer.swift#L825. 

With both rc3 and rc4, when calling `TrackPlayer.reset()`, I see `onCurrentIndexChanged` of `QueuedAudioPlayer` in SwiftAudioEx get executed twice, from which `handleAudioPlayerQueueIndexChange` gets called. In rc3, the `nextIndex` param has a value of nil both times, but in rc4, the second time it has a value of 0 which triggers a crash because `player.items` is empty. This value of `0` traces back to the `newIndex` param of `onCurrentIndexChanged` and is clearly the bug, but is not the root cause, which I have not had a chance to dig deeper into yet. I'm hoping one of the maintainers can save some time with an understanding of what changed recently that broke this.

In the meantime, v2.2.0-rc4 is not usable on iOS unfortunately.

mmmoussa:
I think this change (providing a default of 0) might be the cause of the issue: https://github.com/doublesymmetry/SwiftAudioEx/commit/af803339dcc231725cf4b6c96487df3e1571bdd4#r77334509. My Swift knowledge is very weak and I'm not sure what the necessary fix is.

marcfrankel:
We have also noticed this issue in RC4, might try rolling back to RC3 in the meantime

jspizziri:
@DennisdeWitNL

This is the reset issue, which is caused by an out of range exception. The PR attached to this ticket addresses that issue and has been merged into main. You can confirm by building the example project off of the current main branch

jspizziri:
[...]

PR #1589, opened 2022-07-05
fix(ios): fix an issue with `reset()` causing a crash on iOS due to o…

jspizziri:
…ut of range exception

https://github.com/doublesymmetry/react-native-track-player/issues/1585


mmmoussa:
Tested this in my app and seems to work without any side effects

mpivchev (approved):
LGTM!

kapscudi:
hi, 
I'm still getting the issue TrackPlayer.destroy() and TrackPlayer.reset() without using them I've got no app crash.
I'm using react native 0.69.1 & 2.2.0-rc4

thanx in advance for the help.

PR #1593, opened 2022-07-09
Forward media session events to playback service

doughsay:
Requires https://github.com/doublesymmetry/KotlinAudio/pull/32


doughsay:
**Note**: Jump forward/backward don't receive the configured `interval`. This is an existing issue though and not introduced in this change. Would love some insight into why this is.

mpivchev:
This looks like it's obsolete at this point. If it's not, please reopen this.

PR #1594, opened 2022-07-10
feat: type support for event listener

Spice-Z:
I want event listener's parameter type, but that was `any`!
So
- Defined `interface EventListenerProp` 
- Added generics  to `function addEventListener` to enable type inference in listener function
  - This Generics receives `enum Event value`
  
 Now, listener's param is not any! (from example code)
スクリーンショット 2022-07-10 19 00 41

(Generics might be removed, but I don't figure out how to do it. :sob:)

### DIFF
--- a/src/interfaces.ts
+++ b/src/interfaces.ts
@@ -172,6 +172,79 @@ export enum Event {
   RemoteBookmark = 'remote-bookmark',
 }
 
+export interface EventListenerProp {
+  [Event.PlaybackState]: {
+    state: State
+  }
+  [Event.PlaybackError]: {
+    code: string
+    message: string
+  }
+  [Event.PlaybackQueueEnded]: {
+    track: number
+    position: number
+  }
+  [Event.PlaybackTrackChanged]: {
+    track: number
+    position: number
+    nextTrack: number
+  }
+  [Event.PlaybackMetadataReceived]: {
+    source: string
+    title: string | null
+    url: string | null
+    artist: string | null
+    album: string | null
+    date: string | null
+    genre: string | null
+  }
+  [Event.PlaybackProgressUpdated]: {
+    position: number
+    duration: number
+    buffer: number
+    track: number
+  }
+  [Event.RemotePlay]: null
+  [Event.RemotePlayId]: {
+    id: string
+  }
+  [Event.RemotePlaySearch]: {
+    query: string
+    focus: string
+    title: string
+    artist: string
+    album: string
+    date: string
+    playlist: string
+  }
+  [Event.RemotePause]: null
+  [Event.RemoteStop]: null
+  [Event.RemoteSkip]: {
+    index: number
+  }
+  [Event.RemoteNext]: null
+  [Event.RemotePrevious]: null
+  [Event.RemoteJumpForward]: {
+    interval: number
+  }
+  [Event.RemoteJumpBackward]: {
+    interval: number
+  }
+  [Event.RemoteSeek]: {
+    position: number
+  }
+  [Event.RemoteSetRating]: {
+    rating: boolean | number
+  }
+  [Event.RemoteDuck]: {
+    paused: boolean
+    permanent: boolean
+  }
+  [Event.RemoteLike]: null
+  [Event.RemoteDislike]: null
+  [Event.RemoteBookmark]: null
+}
+
 export enum TrackType {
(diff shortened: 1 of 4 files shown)


src/trackPlayer.ts
```diff
@@ -62,8 +63,8 @@ function registerPlaybackService(factory: () => ServiceHandler) {
   }
 }
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-function addEventListener(event: Event, listener: (data: any) => void) {
+type EventListener<T extends Event> = (data: EventListenerProp[T]) => void
```
jspizziri, line 66: We need to ensure that this is a backward-compatible change. Meaning, we shouldn't require an event type to be passed (needs a default value). People should be able to opt-into this if they want the type safety of it but not be required to update their code when they upgrade to the newest version.

src/interfaces.ts
```diff
@@ -172,6 +172,79 @@ export enum Event {
...
+  [Event.PlaybackMetadataReceived]: {
+    source: string
+    title: string | null
+    url: string | null
+    artist: string | null
+    album: string | null
+    date: string | null
+    genre: string | null
+  }
+  [Event.PlaybackProgressUpdated]: {
```
jspizziri, line 201: Please see the dedicated interface that already exists for this (`ProgressUpdateEvent`). More generally, please abstract all the event structures into their own interfaces and use them in this typemap.

jspizziri (changes requested):
@Spice-Z great work!

I left some feedback. Additionally, I'd love to have something in the docs that shows people how they can take advantage of this type safety.

src/trackPlayer.ts
dcvz: Ah yeah, this one might be a bit tricky.. the change is nice! but it will start throwing typescript errors for those that update.. any ideas on things we can do @Spice-Z ?

jspizziri:
@Spice-Z any updates on this PR? I'd love to merge it if we can fix the BC issue.

puckey:
We can infer the type instead:

```ts
function addEventListener<T extends Event>(event: T, listener: (event: EventDataByEvent[T]) => void) {
  return emitter.addListener(event, listener)
}

interface EventDataByEvent {
  [Event.PlaybackState]: {
    state: State
  }
  [Event.PlaybackError]: {
    code: string
    message: string
  }
  [Event.PlaybackQueueEnded]: {
    track: number
    position: number
  }
  [Event.PlaybackTrackChanged]: {
    track: number
    position: number
    nextTrack: number
  }
  [Event.PlaybackMetadataReceived]: {
    source: string
    title: string | null
    url: string | null
    artist: string | null
    album: string | null
    date: string | null
    genre: string | null
  }
  [Event.PlaybackProgressUpdated]: {
    position: number
    duration: number
    buffer: number
    track: number
  }
  [Event.RemotePlay]: null
  [Event.RemotePlayId]: {
    id: string
  }
  [Event.RemotePlaySearch]: {
    query: string
    focus: string
    title: string
    artist: string
    album: string
    date: string
    playlist: string
  }
  [Event.RemotePause]: null
  [Event.RemoteStop]: null
  [Event.RemoteSkip]: {
    index: number
  }
  [Event.RemoteNext]: null
  [Event.RemotePrevious]: null
  [Event.RemoteJumpForward]: {
    interval: number
  }
  [Event.RemoteJumpBackward]: {
    interval: number
  }
  [Event.RemoteSeek]: {
    position: number
  }
  [Event.RemoteSetRating]: {
    rating: boolean | number
  }
  [Event.RemoteDuck]: {
    paused: boolean
    permanent: boolean
  }
  [Event.RemoteLike]: null
  [Event.RemoteDislike]: null
  [Event.RemoteBookmark]: null
}
`

jspizziri:
@puckey I like that approach, but just to be clear. This would require a generic argument to be passed into `addEventListener` still would it not? Meaning `addEventListener(...)` would cause a ts error since you're not specifying an argument?

puckey:
@jspizziri No it wouldn't – the type is inferred:

Screenshot 2022-08-10 at 16 11 50

puckey:
See here for my additional changes to this PR: https://github.com/doublesymmetry/react-native-track-player/pull/1640

jspizziri:
closing in favor of the new PR.

PR #1595, opened 2022-07-10
chore: add CI task - check-typescript

Spice-Z:
I found main branch has some format error in TypeScript code.
To prevent new error injection, I added a CI task that check TS code.

[Here](https://github.com/Spice-Z/react-native-track-player/pull/2), I confirmed CI task failed when TS code has error.

### DIFF
--- a/.github/workflows/platform-validation.yml
+++ b/.github/workflows/platform-validation.yml
@@ -87,3 +87,22 @@ jobs:
         run: |-
           cd example/ios
           xcodebuild build -workspace example.xcworkspace -scheme example -destination 'platform=iOS Simulator,name=iPhone 12 Pro'
+  check-typescript:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v2
+      - name: Setup Node
+        uses: actions/setup-node@v2
+        with:
+          node-version: 16.x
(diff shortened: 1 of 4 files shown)


.github/workflows/platform-validation.yml
```diff
@@ -87,3 +87,22 @@ jobs:
...
+      - name: Cache JS Dependencies
+        uses: c-hive/gha-yarn-cache@v1
+      - name: Install Yarn
+        run: npm install --global yarn
+      - name: Install Library Dependencies
+        run: yarn install
+      - name: Check Lint & Format
+        run: |
+          npm run lint:ci
+          npm run format:ci
```
dcvz, line 108: `format` is usually used for fixing the linting issues right? will the workflow then try to commit them?

Spice-Z: I think , it depends on how the project define `lint` and `format`.

RNTP's package.json define like this.
```
"lint": "eslint \"src/**/*.ts\"",
"format": "prettier --write \"src/**/*.ts\"",
```

Eslint fixing and Prettier fixing are not always same, so I thought it's better to check both of them.

I added `lint:ci` and `format:ci` to package.json, those command just check code, don't commit any diff.

dcvz (approved):
LGTM 👍

PR #1597, opened 2022-07-11
Fix/jump intervals

jspizziri:
Jump intervals are not correctly being passed to JS because they are `Double`'s and not `Int`'s.

### DIFF
--- a/android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
+++ b/android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
@@ -391,19 +391,21 @@ class MusicService : HeadlessJsTaskService() {
                     is STOP -> emit(MusicEvents.BUTTON_STOP)
                     is FORWARD -> {
                         Bundle().apply {
-                            val interval = (latestOptions?.getInt(FORWARD_JUMP_INTERVAL_KEY) ?: latestOptions?.getInt(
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
```diff
@@ -391,19 +391,19 @@ class MusicService : HeadlessJsTaskService() {
                     is STOP -> emit(MusicEvents.BUTTON_STOP)
                     is FORWARD -> {
                         Bundle().apply {
-                            val interval = (latestOptions?.getInt(FORWARD_JUMP_INTERVAL_KEY) ?: latestOptions?.getInt(
-                                    BACKWARD_JUMP_INTERVAL_KEY)) ?: 15
+                            val interval = (latestOptions?.getDouble(FORWARD_JUMP_INTERVAL_KEY) ?: latestOptions?.getDouble(
+                                    BACKWARD_JUMP_INTERVAL_KEY)) ?: 15.0
```
mpivchev, line 395: Let's maybe put these magic numbers in constants

mpivchev (commented):
LGTM :) Just one comment

jspizziri:
@mpivchev done, I also caught and fixed another bug where the default value was never actually being passed (because `getDouble` returns 0 if no value is found)

mpivchev (approved):

PR #1607, opened 2022-07-20
ci(platform-validation): skip platform validation if only documentati…

jspizziri:
…on is modified

https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks


jspizziri:
@dcvz I _believe_ this should bypass the `platform-validation` job when only docs are modified. See the link in the description form more info. The only thing I'm unsure about is what happens if both `docs/**` and some actual code are modified in the same PR. In that case we do want to run actual validation. I think this does the trick but I don't know how to verify it without merging first.

PR #1610, opened 2022-07-20
ci(platform-validation-skip): add specific build jobs to platform-val…

jspizziri:
…idation-skip

existing implementation still expects the build-ios and build-android jobs:
https://github.com/doublesymmetry/react-native-track-player/pull/1604

### DIFF
--- a/.github/workflows/platform-validation-skip.yml
+++ b/.github/workflows/platform-validation-skip.yml
@@ -18,7 +18,15 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  build:
+  build-android:
+    runs-on: ubuntu-latest
+      steps:
+        - run: 'echo "No build required"'
+  build-ios:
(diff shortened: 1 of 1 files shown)


jspizziri:
@dcvz I tested the workflow and something doesn't seem quite right still. I think it might require all the jobs to be present (see the linked PR as an example).

jspizziri:
@dcvz correct, that's how the docs say to do it. I thought that it just depended on the pipeline name `platform-validation`, but it looks like it depends on specific jobs as well. Here's the correct test PR i created https://github.com/doublesymmetry/react-native-track-player/pull/1611

jspizziri:
I'm thinking if we get it working we should do the same thing for the `.github/workflows` directory too. That ios build takes a painful amount of time

jspizziri:
@dcvz correct.

PR #1615, opened 2022-07-20
ci(platform-validation-skip): update paths for skip workflow to expli…

jspizziri:
…citly include docs/**


jspizziri:
@dcvz alright... here's my most recent attempt. I wish I knew of a way to test this kind of stuff without merging it, but AFAIK it's not possible.

PR #1617, opened 2022-07-20
ci(platform-validation-skip): fix a yml syntax error and simplify the…

jspizziri:
… jobs

### DIFF
--- a/.github/workflows/platform-validation-skip.yml
+++ b/.github/workflows/platform-validation-skip.yml
@@ -18,15 +18,7 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  build-android:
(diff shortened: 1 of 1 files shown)


jspizziri:
@dcvz this actually works finally. I tested it on my fork before creating this PR. It also works correctly if both code and docs are modified simultaneously.

jspizziri:
Turns out the big issue is that I'm an idiot... and had some invalid yml in there.

jspizziri:
What are your thoughts and disabling the builds for changes to the `.github` directory?

jspizziri:
Probably not worth it. We could probably disable them for the `example` dir though right?

PR #1619, opened 2022-07-20
ci(platform-validation-skip): mirror the specific jobs from platform-…

jspizziri:
…validation.yml


jspizziri:
@dcvz no idea why when I tested this my repo behaved differently when I ran it against my repo than this one. But this one seems to require identical jobs within the action.

This... will be my last attempt to make this work regardless.

PR #1625, opened 2022-07-27
Replace stopWithApp key with stoppingAppPausesPlayback in typescript sources and documentations

nicomontanari:


mpivchev:
Thanks for the addition :)

nicomontanari:
You are welcome! 😃

jspizziri:
@nicomontanari @mpivchev , Sorry I'm late on this, but can we create an alias for `stopWithApp` that points to `stoppingAppPausesPlayback` so we don't create a breaking change?

nicomontanari:
@jspizziri it may be an idea, if you decide to do it I could make these changes

PR #1629, opened 2022-08-01
Provide an example for RemoteDuck behaviour

hvihvi:
Implementing a standard use-case that pauses when another app takes over wasn't obvious.

The check "was paused by duck" has to be handled in the callback, providing an example of working code that does this might help understand edge cases.

Our initial "naive" implementation was :
```js
  TrackPlayer.addEventListener(Event.RemoteDuck, (event) => {
    if (event.paused) {
      TrackPlayer.pause();
    } else {
      TrackPlayer.play();
    }
  });
```

However this would start playing the backgrounded track when the "ducked" state ends, regardless of the track being initially paused by human interaction or by the duck event.

### DIFF
--- a/docs/docs/api/events.md
+++ b/docs/docs/api/events.md
@@ -169,3 +169,5 @@ On Android, the volume may also be lowered on an transient interruption without 
 | --------- | --------- | -------------------------------------------- |
 | paused    | `boolean` | On Android when `true` the player should pause playback, when `false` the player may resume playback. On iOS when `true` the playback was paused and when `false` the player may resume playback. |
 | permanent | `boolean` | Whether the interruption is permanent. On Android the player should stop playback.  |
+
+Implementation examples can be found in the [example project](https://github.com/doublesymmetry/react-native-track-player/blob/main/example/src/services/PlaybackService.ts).


docs/docs/api/events.md
```diff
@@ -169,3 +169,20 @@ On Android, the volume may also be lowered on an transient interruption without
...
 | paused    | `boolean` | On Android when `true` the player should pause playback, when `false` the player may resume playback. On iOS when `true` the playback was paused and when `false` the player may resume playback. |
 | permanent | `boolean` | Whether the interruption is permanent. On Android the player should stop playback.  |
+
+Example (pauses when another app takes over):
+```js
+  // service.js
+  let wasPausedByDuck = false;
+  TrackPlayer.addEventListener(Event.RemoteDuck, async (event) => {
+    const state = await TrackPlayer.getState();
+    if (event.paused && state === State.Playing) {
```
jspizziri, line 179: @hvihvi Thanks for the contribution!

Question: I've not worked with `RemoteDuck` before, however reading the documentation it looks like it behaves slightly differently on iOS vs Android. Have you taken that into account here?

https://react-native-track-player.js.org/docs/api/events#remoteduck

Another thought is that this could probably also be implemented directly in the `example` project too.

hvihvi: My bad, the example project already provides an even better implementation, I should have looked there first. 😅

I amended the commit to link to the example's implementation instead if that helps, but it would also make sense to drop it and leave the doc as it is.

mpivchev (approved):
LGTM

Issue #1613, opened 2022-07-20
Android events always return a position of 0


jspizziri:
As originally documented on https://github.com/doublesymmetry/react-native-track-player/issues/1389

Android currently always returns `position = 0` on all track-changed and queue-ended events.

jspizziri:
@mpivchev before this can be closed kotlinaudio needs to be released and updated here and we need to update the hard coded 0 to pull from the event

mpivchev:
@jspizziri doing this atm

PR #1643, opened 2022-08-11
Add a proper position to the AudioItemTransition event 

mpivchev:
This is a followup to https://github.com/doublesymmetry/KotlinAudio/pull/36 which fixes https://github.com/doublesymmetry/react-native-track-player/issues/1613


android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
```diff
@@ -323,20 +325,21 @@ class MusicService : HeadlessJsTaskService() {
...
+                    }
                 }
             }
         }
 
         scope.launch {
             event.audioItemTransition.collect {
                 Bundle().apply {
-                    putDouble(POSITION_KEY, 0.0)
+                    putDouble(POSITION_KEY, (it?.oldPosition ?: 0).toSeconds())
```
mpivchev, line 342: We now provide the proper old position for this event.

jspizziri (approved):
LGTM! Thank you!

PR #1646, opened 2022-08-11
[ios] Remove destroy

dcvz:
Missed before the 3.0.0 release

### DIFF
--- a/ios/RNTrackPlayer/RNTrackPlayer.swift
+++ b/ios/RNTrackPlayer/RNTrackPlayer.swift
@@ -468,7 +468,6 @@ public class RNTrackPlayer: RCTEventEmitter, AudioSessionControllerDelegate {
         }
 
         player.stop()
-        player.reset()
         player.nowPlayingInfoController.clear()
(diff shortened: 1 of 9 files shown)


jspizziri:
@dcvz We'd need to remove the method from JS too.

jspizziri (dismissed):
LGTM, I don't have all the perspective but the docs and Js seem to check out.

mpivchev:
Also attempts to fix https://github.com/doublesymmetry/react-native-track-player/issues/1647

ios/RNTrackPlayer/RNTrackPlayer.swift
```diff
@@ -482,6 +468,7 @@ public class RNTrackPlayer: RCTEventEmitter, AudioSessionControllerDelegate {
         }
 
         player.stop()
+        player.reset()
```
mpivchev, line 471: This is a private function of SwiftAudio. Did you mean to put something else here? @dcvz

mpivchev (approved):
LGTM :)

Issue #1418, opened 2022-02-24
Duplicate package names


mpivchev:
Hi, thanks for this great library.
I am working in a project that uses a other players that use old exoplayer dependencie and I having these kind of errors:

```
 Duplicate class com.google.android.exoplayer2.ui.WebViewSubtitleOutput$1 found in modules jetified-exoplayer-ui-2.14.2-runtime (com.google.android.exoplayer:exoplayer-ui:2.14.2) and jetified-exoplayer-ui-r2.16.4-runtime (com.github.DoubleSymmetry.ExoPlayer:exoplayer-ui:r2.16.4)

 Duplicate class com.google.android.exoplayer2.video.spherical.CameraMotionListener found in modules jetified-exoplayer-core-2.14.2-runtime (com.google.android.exoplayer:exoplayer-core:2.14.2) and jetified-exoplayer-core-r2.16.4-runtime (com.github.DoubleSymmetry.ExoPlayer:exoplayer-core:r2.16.4)
```

It is because Kotlin audio is using the same "com.google.android.exoplayer2" package name, maybe could be it changed for future versions to avoid the conflict ?

_Originally posted by @SystemBlad in https://github.com/DoubleSymmetry/react-native-track-player/discussions/1264#discussioncomment-2185744_

hammadwahab12:
This issue is still open in 2.2.0-rc3 version @mpivchev can you please let me know when this can be fixed. Thanks

rajpootathar:
Still facing the same issue on my end (with 2.2.0-rc3 version).
Further [details ](https://github.com/DoubleSymmetry/react-native-track-player/discussions/1264#discussioncomment-2264733)

mpivchev:
Hi, currently working on #1447. I will pick this issue up after this. If one of you wants to contribute to this issue so we can get it out sooner, we would greatly appreciate it :)

hammadwahab12:
Hello @mpivchev, any update ?

mpivchev:
@hammadwahab12 It's still in our to-do list. As mentioned before, if you want this resolved quicker, you are welcome to contribute.

jspizziri:
Track this PR which should ultimately resolve this issue:  https://github.com/doublesymmetry/KotlinAudio/pull/33

KrisLau:
@jspizziri Thanks for all the awesome work!

stereodenis:
I downgraded to 2.1.3, and it works with react-native-video v5.2.0

[...]

PR #1648, opened 2022-08-13
docs: :memo: Tourbleshoot Android: Duplicate package names #1418

Hannes1:
A temporary workaround for people that have errors with conflicting exoplayers


mpivchev:
This should not be a problem in 3.1.0 anymore. However, if you think otherwise, please reopen this issue.

Issue #1659, opened 2022-08-20
Types of events on version 3.1.0 are incorrect


zzorba:
**Describe the Bug**
It looks like version 3.1.0 added typescript types for the callback passed to the `useTrackPlayerEvents` hook.  Unfortunately, the types do not seem to match the actual code (at least on iOS).

The differentiator field claims to be 'event', however when logging it it turns out it should actually be 'type'
https://github.com/doublesymmetry/react-native-track-player/commit/958a33018d319ce8c30e90590f532c505e8707a2

``` 
 LOG  {"nextTrack": 0, "position": 0, "type": "playback-track-changed"}
 LOG  {"state": "connecting", "type": "playback-state"}
 LOG  {"nextTrack": 0, "position": 0, "track": 0, "type": "playback-track-changed"}
```

Unclear if Android behaves differently, but at least on iOS it doesn't seem to match as expected.

*Steps To Reproduce**
Try to use the events with their documented types, and see that it does not work.  Change the fields to 'type' instead of event works.

**Code To Reproduce**
```
  useTrackPlayerEvents([Event.PlaybackTrackChanged, Event.PlaybackState],
   (event) => {
      console.log(event);
   })
```

**Environment Info:**
react-native-track-player: 3.1.0

**How I can Help**
I'd be happy to submit a PR, but unclear if there is a longer term plan to move to 'event' instead of 'type' as the differentiator for specialization.

zzorba:
opened this preemptively: https://github.com/doublesymmetry/react-native-track-player/compare/main...zzorba:react-native-track-player:patch-1

zzorba:
It is the same on both platforms. I think a mistake was just made when documenting the types in the index.d.at file.

zzorba:
Yes, I confirmed that both platforms see the differentiator as 'type' not 'event'.  Same code is working great on both players with v3.1.0, no need for platform specialization

jspizziri:
I created a PR from @zzorba 's branch.

jspizziri:
Looks like I'm confused about this one. My bad! I believe this can be closed now via https://github.com/doublesymmetry/react-native-track-player/pull/1662

PR #1664, opened 2022-08-22
fix(ts): fix an issue with event callback type declarations

jspizziri:
The field that differentiates events appears to be 'type' not 'event'


jspizziri:
@dcvz this is from [the work @zzorba did here] (https://github.com/doublesymmetry/react-native-track-player/issues/1659#issuecomment-1221395406) but it looks like a PR never got created for it.

jspizziri:
@dcvz weird... if that's the case why is there a diff showing up for it? GitHub must be confused somehow.

PR #1667, opened 2022-08-23
docs(v3): add minimum react native version

hvihvi:
When upgrading from v2.1.3 to v3.1 an error occurs that requires an upgrade from 30 to 31 for these 2 in `android/build.gradle`:

```
        compileSdkVersion = 31
        targetSdkVersion = 31
```

According to https://react-native-community.github.io/upgrade-helper/?from=0.67.4&to=0.68.0 this is done when upgrading to react-native v0.68.0.

### DIFF
--- a/docs/versioned_docs/version-3.1/v3-migration.md
+++ b/docs/versioned_docs/version-3.1/v3-migration.md
@@ -17,6 +17,7 @@ how other popular apps (Spotify, Soundcloud, Google Podcasts, etc.) work.
 - The `destroy` function does not exist anymore. 
 - The `stopWithApp` flag turns into `stoppingAppPausesPlayback` 
     - [More information here](https://github.com/doublesymmetry/react-native-track-player/pull/1447#issuecomment-1195246389)
+- The react-native version has to be 0.68 or higher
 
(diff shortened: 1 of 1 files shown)


jspizziri:
@hvihvi good catch. Question though: couldn't users just upgrade their `compileSdkVersion` and `targetSdkVersion`'s without upgrading to RN 0.68?

hvihvi:
> @hvihvi good catch. Question though: couldn't users just upgrade their `compileSdkVersion` and `targetSdkVersion`'s without upgrading to RN 0.68?

You're right, I got confused, the RN 0.68 upgrade is not necessary.

Just in case someone else faces the same error, this was the error:

```
* What went wrong:
Execution failed for task ':app:checkDebugAarMetadata'.
> A failure occurred while executing com.android.build.gradle.internal.tasks.CheckAarMetadataWorkAction
   > The minCompileSdk (31) specified in a
     dependency's AAR metadata (META-INF/com/android/build/gradle/aar-metadata.properties)
     is greater than this module's compileSdkVersion (android-30).
     Dependency: androidx.core:core-ktx:1.8.0.
```

It was fixed by upgrading from 30 to 31 in `android/build.gradle`:
```
        compileSdkVersion = 31
        targetSdkVersion = 31
```

PR #1668, opened 2022-08-23
Intercept events

mpivchev:


android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
```diff
@@ -381,7 +385,7 @@ class MusicService : HeadlessJsTaskService() {
         }
 
         scope.launch {
-            event.onMediaSessionCallbackTriggered.collect {
+            event.onPlayerActionTriggeredExternally.collect {
```
mpivchev, line 388: This observer should now be collecting events when media session OR notification triggers player events. Make sure to put some breakpoints here and test it.

jspizziri:
@mpivchev (cc @doughsay) I just tested RNTP. I tested the following events:

- `RemotePlay`
- `RemotePause`
- `RemoteNext`
- `RemotePrevious`

All of them were working as I expected 🎊 . I tested both the notification player and via connected Bluetooth headphones.

Here are some things I noticed that did not seem to be working:

- It seems the scrubbing/seeking the progress bar in the notification player isn't working (even with `Capability.SeekTo` enabled).
- I'm receiving duplicate `RemoteJumpForward` and `RemoteJumpBackward` events (hit the button on the notification player once, and I get 2 events). 
  - NOTE: the jump isn't being handled by KotlinAudio, which is consistent with iOS.
- On the notification player, the Fast Forward & Rewind buttons (double arrows right and left) seem to be triggering the `RemoteJump` events, rather than `RemoteSeek` which I would expect

mpivchev:
> Here are some things I noticed that did not seem to be working:
> 
> * It seems the scrubbing/seeking the progress bar in the notification player isn't working (even with `Capability.SeekTo` enabled).

Fixed :)

> * I'm receiving duplicate `RemoteJumpForward` and `RemoteJumpBackward` events (hit the button on the notification player once, and I get 2 events).

image

Can't reproduce it on my side, I only receive one event. Maybe it got fixed along with other things.


> * On the notification player, the Fast Forward & Rewind buttons (double arrows right and left) seem to be triggering the `RemoteJump` events, rather than `RemoteSeek` which I would expect

These are actually the right events, since we used them before removing the fork as well. The only issue was the RN side did not handle them. The new commits fix that.

mpivchev:
PR is updated with the fixes, new KA version is 1.2.1

jspizziri:
@mpivchev thanks for the work on this. I'll retest today and let you know if I have more feeback.

jspizziri:
@mpivchev I just finished running through all the same tests. I can confirm everything is now working as I expect.

Let's merge and release 🎊

jspizziri (approved):

PR #1670, opened 2022-08-23
build(package.json): add release & changelog generation automation

jspizziri:


jspizziri:
@dcvz @mpivchev 

This automates the change log generation when the `yarn version` command is run. One thing of note, is that it will only generate an incremental changelog for the last release and append it to the CHANGELOG.md file. This means that it will note overwrite anything that's currently in there.

There's also a husky hook in there that, upon commit, will prompt people to walk through the proper steps to format their commit appropriately.

jspizziri:
Alright, please take a look at a test generation in my repo. This is the result of running `yarn version --minor` (it should also build and release to npm):

https://github.com/5-stones/react-native-track-player/commit/8827c01ec7c72d978da31a79efd49dd50c7c0415

It might not be a perfect changelog, but we can at the very least use it as the basis for making manual changes.

If things look good enough all we need to do is merge this PR and run `yarn version --minor`.

PR #1672, opened 2022-08-24
fix(hooks): fix useTrackPlayerEvents dependencies

puckey:
The effect was being executed every render because we were passing `[[Event, Event]]` instead of `[Event, Event]` as dependency.


puckey:
@dcvz thanks – disabled the eslint rule because this is a special case

PR #1674, opened 2022-08-24
chore(hooks): use progressUpdateEventInterval

puckey:
This pull request uses the new `progressUpdateEventInterval` functionality for the `useProgress` hook.

Something to note is that calling the hook also calls TrackPlayer.setOptions to set the `progressUpdateEventInterval`.

@jspizziri I would be interested to hear your thoughts on this implementation!

### DIFF
--- a/example/src/components/Progress.tsx
+++ b/example/src/components/Progress.tsx
@@ -1,10 +1,10 @@
-import React from 'react';
-import { StyleSheet, Text, View } from 'react-native';
 import Slider from '@react-native-community/slider';
-import TrackPlayer, { useProgress } from 'react-native-track-player';
+import React from 'react';
+import {StyleSheet, Text, View} from 'react-native';
+import TrackPlayer, {useProgress} from 'react-native-track-player';
 
 export const Progress: React.FC = () => {
-  const progress = useProgress();
+  const progress = useProgress(2000);
   return (
     <>
       <Slider
--- a/example/src/services/SetupService.ts
+++ b/example/src/services/SetupService.ts
@@ -24,7 +24,6 @@ export const SetupService = async (): Promise<boolean> => {
         Capability.SkipToNext,
         Capability.SeekTo,
       ],
-      progressUpdateEventInterval: 2,
     });
 
     isSetup = true;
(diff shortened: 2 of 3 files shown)


jspizziri:
@puckey my first reaction:

I'm not sure about having `useProgress` update the progress interval. It's an interesting idea, but the point of the progress interval at it's core is for headless contexts. I could see someone wanting frequent progress updates when the app is in the foreground (every 1s), but in the background reducing the frequency of those events.

Doing it this way creates some "magic" where users might not be aware that they're actually modifying the value in both contexts. Perhaps a middle ground would be to have another hook called `useProgressUpdatedEvents` or something that does this, but then we're adding more code that we need to maintain.

I'd be interested in what core problem you need to solve as that might inform my opinion. Or is it just that doing it the way we're currently doing it makes you feel dirty inside? 😄

I'd be interested in @dcvz 's thoughts as well.

puckey:
I found it confusing that on the one hand we have this `useProgress` hook and we also have this `progressUpdateInterval` thing – but they are not related. The `useProgress` hook is pretty heavy – three async calls every n seconds through js => native just for progress.

We could also incorporate the lowering (or disabling) of the interval when the app goes to background in the hook.

Anyway, I am aware of this being a bit too magic. Another approach would be if we had a way also retrieve the options that have been set (`getOptions`), then we could throw an error on `__DEV__` if `options.progressUpdateInterval` was not set.

What is the headless context that this was designed for? Is it meant for sending analytics in the background or something?

jspizziri:
@puckey there are a few things that this solves in headless contexts:

1. [Tracking playback position during headless playback](https://react-native-track-player.js.org/docs/guides/saving-progress) (this is for on-demand contexts)
2. [Implementing Sleeptimers](https://react-native-track-player.js.org/docs/guides/sleeptimers)

puckey:
@jspizziri smart – makes sense! I think this implementation does indeed cause too many complications. Let's close it for now – and perhaps I come up with a better way of integrating the two at some point.

jspizziri:
@puckey thanks for working on it! I too don't feel super great about it, I just don't know of a more optimal solution right now.

PR #1683, opened 2022-08-30
chore(example): format example

puckey:
Ran everything in example/src through prettier, to avoid having these whitespace changes showing up in other pull requests. Maybe we need to add something to the ci to also check for this?

I think this comes from the fact that the .prettierrc.js in the example directory has different settings to that in the rntp directory itself. When you edit the example app through the rntp directory in vscode, it will use rntp/.prettierrc.js instead of rntp/example/.prettierrc.js

Perhaps we might unify the two to avoid running into this in the future?


puckey:
Yep lets do it that way

PR #1447, opened 2022-03-29
Fix #1377: app process not stopping

mpivchev:
This PR is aimed to fix the bug where the app process would not die even when the app + service were killed.
The cause of this bug is not concrete, and a lot of the issues here are very strange, but I do have some theories:
- We are creating and controlling an Android service in a RN module. This is outside of its intended container - an Android Activity. Activities have a special lifecycle tied to the Android OS, and a botched lifecycle might be one of the reasons why the service and process are misbehaving.
- The main component of this fix was to introduce the Activity lifecycles to our RN module, and using them to control our service.

The good news is it did cause the process to die properly, but it introduced other bugs related to notifications (services are heavily tied to Android notifications). More info [in this separate PR](https://github.com/doublesymmetry/KotlinAudio/pull/24).

---------------
Here is how the current implementation of this PR:

1. There is an Android interface that you can bind to an activity that can send lifecycle events:

```kotlin
interface ActivityLifecycleCallbacksAdapter: Application.ActivityLifecycleCallbacks {
    override fun onActivityCreated(activity: Activity, bundle: Bundle?) = Unit

    override fun onActivityStarted(activity: Activity) = Unit

    override fun onActivityResumed(activity: Activity) = Unit

    override fun onActivityPaused(activity: Activity) = Unit

    override fun onActivityStopped(activity: Activity) = Unit

    override fun onActivitySaveInstanceState(activity: Activity, p1: Bundle) = Unit

    override fun onActivityDestroyed(activity: Activity) = Unit
}
```

2. I bind it through our module by getting the currentActivity. The currentActivity will always be a ReactActivity, and this way we don’t need to require people to add code to said activity, since we just grab it internally in our module :
```kotlin
 override fun initialize() {
        currentActivity?.application?.registerActivityLifecycleCallbacks(this)
    }
```

3. The first service bind happens during setupPlayer of the module, since we can’t listen to the first activity onStart (it happens before the RN module is initialized)

```kotlin
Intent(context, MusicService::class.java).also { intent ->
    context.startService(intent)
    // This bind only happens the first time. Subsequent binds happen in onActivityStarted
    context.bindService(intent, this, Context.BIND_AUTO_CREATE)
}
```

4. We use the listener events to control the service.
- As mentioned, we bind for the first time in step 3. UI starts to update.
- When app goes to recents, we unbind in onActivityStopped . UI stops updating but the service keeps running and waits for a rebind.
[...]

PR #1687, opened 2022-08-30
Dismiss notification when app is removed from recent apps list

filippobusi:
With version 3.1.0 the multimedia notification is not dismissed when the app has been terminated by the user. The only way to dismiss the notification is to force quit the app from the settings. I have made these changes to prevent this behavior, now when the user kills the app the notification also closes.

### DIFF
--- a/android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
+++ b/android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
@@ -456,7 +456,7 @@ class MusicService : HeadlessJsTaskService() {
         super.onTaskRemoved(rootIntent)
 
         if (stoppingAppPausesPlayback) {
-            player.pause()
+            player.stop()
         }
     }
 


afkcodes:
This seems a legible solution, thus needs testing

afkcodes:
@mpivchev @jspizziri  is this seems viable ? Seems like there is no functional changes

jspizziri:
This seems to violate [the principle that guided the decision](https://react-native-track-player.js.org/docs/v3-migration) in the first place to have an always one player (to avoid several other complications).

I'm not sure what the ramifications of this are. @mpivchev will need to do the review/approval on this.

afkcodes:
yeah lets wait for @mpivchev to have a look

nicomontanari:
@jspizziri ok but why should the user always see the player notification even if he closed the app? In the end the user usually closes the app in order not to use it..
In fact we have received several reports from users who have complained about this problem. 

This is my opinion following our experiences, but I would like to understand your choices as well

afkcodes:
dont know about others device but for my device when i remove the app from recents the notification goes for spotify and apple music.

device tested: One Plus nord 2 and Xiaomi Poco F4

filippobusi:
> The backing issue was that we can't control when the music service is destroyed. So even if you dismiss/destroy the notification its not indicative that the service itself has stopped (stopping that is controlled by the OS).
> 
> As part of that we decided to do what most other big audio players do (Spotify, Apple Music, etc) which is have an always on notification.

When `player.stop` is called inside `onTaskRemoved` the notification associated with the player is removed so the `NotificationState.CANCELLED` callback is triggered. 
Inside this callback there is `stopForeground(true)` which will stop the service.

The user expects to terminate everything that your app is doing when manually close the application from recent apps list.
If the app is reopened the service will restart normally.

filippobusi:
> With this solution, if you open the app again afterwards does everything still function properly?

Yes, it's seems to work well, but this [problem](https://github.com/doublesymmetry/react-native-track-player/pull/1447#pullrequestreview-1051139407) is still present

mmmoussa:
The new "feature" of the notification persisting after the user terminates the app is a bug in my opinion as well. It's actually blocking us from upgrading to 3.x. If it was opt-in or opt-out that would be fine, but in our use-case the track player is a secondary functionality of the app rather than primary and persisting the notification is guaranteed to bring us bug reports.

nicomontanari:
https://github.com/doublesymmetry/react-native-track-player/issues/1715

jspizziri:
I'm closing this PR in favor of this one. Thanks for the work on this @filippobusi !

https://github.com/doublesymmetry/react-native-track-player/pull/1730

PR #1706, opened 2022-09-07
ref(ios, events): improve disabling of playback-progress-updated

puckey:
- When 0 is passed to `options["progressUpdateEventInterval”]` in order to disable ‘playback-progress-updated’ event, set `timeEventFrequency` back to the default of every second.
- rename (private) shouldEmitUpdateEventInterval to shouldEmitProgressEvent

### DIFF
--- a/ios/RNTrackPlayer/RNTrackPlayer.swift
+++ b/ios/RNTrackPlayer/RNTrackPlayer.swift
@@ -18,7 +18,7 @@ public class RNTrackPlayer: RCTEventEmitter, AudioSessionControllerDelegate {
     private var hasInitialized = false
     private let player = QueuedAudioPlayer()
     private let audioSessionController = AudioSessionController.shared
-    private var shouldEmitUpdateEventInterval: Bool = false
+    private var shouldEmitProgressEvent: Bool = false
 
     // MARK: - Lifecycle Methods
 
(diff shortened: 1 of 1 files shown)


jspizziri:
@puckey was this already the default behavior on the Kotlin side?

puckey:
> @puckey was this already the default behavior on the Kotlin side?

Good point – I haven't checked that yet. Let's keep this open until I have done so.

puckey:
@jspizziri I am unable to find the same issue on the kotlin side that ios has where you are unable to 'cancel' this interval  (see your comments here: https://github.com/doublesymmetry/react-native-track-player/blob/a243210ca357c0c5969480683bed55f969e044af/ios/RNTrackPlayer/RNTrackPlayer.swift#L817-L821)

Here is the kotlin side which looks like it cancels the job correctly when set to `0`: https://github.com/doublesymmetry/react-native-track-player/blob/e94e436fd36de665933b58554bbe15a1b2686ca9/android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt#L152-L159

ios/RNTrackPlayer/RNTrackPlayer.swift
```diff
@@ -307,19 +307,18 @@ public class RNTrackPlayer: RCTEventEmitter, AudioSessionControllerDelegate {
...
         resolve(NSNull())
     }
 
     private func configureProgressUpdateEvent(interval: Double) {
-        let time = CMTime(seconds: interval, preferredTimescale: 1)
-        self.player.timeEventFrequency = .custom(time: time)
+        shouldEmitProgressEvent = interval > 0
+        self.player.timeEventFrequency = shouldEmitProgressEvent
+            ? .custom(time: CMTime(seconds: interval, preferredTimescale: 1))
+            : .everySecond
```
jspizziri, line 321: The big difference I see here is that we're setting the interval to 1s when we cancel, which we weren't doing before. Are we sure this is desired?

jspizziri:
Left a question, but lgtm apart from that.

ios/RNTrackPlayer/RNTrackPlayer.swift
puckey: the reason I am doing this is that the user might have set the interval to something very small temporarily, expecting that setting it back to `0` would return stuff to normal.. (I was going off the comment that says it is impossible to turn it off..)

jspizziri (approved):
LGTM!

Issue #1691, opened 2022-08-31
In certain cases audio can be paused due to queue methods


puckey:
While browsing through the source code, I have seen what I perceive to be a mix-up between only doing something when the audio **is** playing and only doing something when the play **will** play.

The android version of skip will only play the item when ready if the player was already playing when the function was called :
https://github.com/doublesymmetry/react-native-track-player/blob/9ed308c2a8f5bbd4ab69df01a7cfb86047960538/android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt#L250-L253

Instead of passing `player.isPlaying` here – I guess we would want to pass the underlying value of `exoPlayer.playWhenReady`

Same goes in the iOS version – the call to `player.jumpToItem` passes `playWhenReady: player.playerState == .playing`:
https://github.com/doublesymmetry/react-native-track-player/blob/27b85b72d85c5347ce9205d8b0bd1b5b433f34f3/ios/RNTrackPlayer/RNTrackPlayer.swift#L385-L411

In the add function, the current `playWhenReady` value will be overridden to `false` on first call to an empty queue, even if it was `true` by a previous call of `play()`

https://github.com/doublesymmetry/react-native-track-player/blob/27b85b72d85c5347ce9205d8b0bd1b5b433f34f3/ios/RNTrackPlayer/RNTrackPlayer.swift#L347-L349

puckey:
The Android implementation of `add(tracks)` looks like it will always pause playback when called:

https://github.com/doublesymmetry/react-native-track-player/blob/9ed308c2a8f5bbd4ab69df01a7cfb86047960538/android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt#L204-L207

causing `exoPlayer.playWhenReady = false` in QueuedAudioPlayer#add(items):

https://github.com/doublesymmetry/KotlinAudio/blob/5887c86181d63f525b91c9b028d3052e858b087b/kotlin-audio/src/main/java/com/doublesymmetry/kotlinaudio/players/QueuedAudioPlayer.kt#L112-L118

puckey:
Wouldn't the best route be to discard all setting of `playWhenReady` and just have the last call to `play`, `pause` or `stop` define whether tracks will play after loading?

puckey:
The work needed to fix this comes down to creating function overloads for the following KotlinAudio and SwiftAudio – leaving out the `playWhenReady` argument:

KotlinAudio:
[...]

PR #1713, opened 2022-09-08
Queue improvements

puckey:
A few improvements to functionalities relying on updates of the underlying SwiftAudioEx and KotlinAudio libraries. 

SwiftAudio updates pr: https://github.com/doublesymmetry/SwiftAudioEx/pull/28.  
KotlinAudio updates pr: https://github.com/doublesymmetry/KotlinAudio/pull/50

### Feature 
- Implement `TrackPlayer.load(track)`: replaces the current track with the supplied track or creates a track when the queue is empty. This feature would mainly be used by simple single track apps – for example a radio station app
- Implement `TrackPlayer.move(fromIndex, toIndex)`
- Implement `TrackPlayer.getProgress()`, and deprecate `TrackPlayer.getBufferedPosition()`, `TrackPlayer.getDuration()`, `TrackPlayer.getPosition()`.
- Implement `TrackPlayer.getPlaybackState()` and deprecate `TrackPlayer.getState()`
- Implement `State.Error` which denotes that the player experienced a playback error causing the audio to stop playing (or not start playing). In this state, the return value of `TrackPlayer.getPlaybackState()` and the payload of the playback state event emitter will include an `error` property with the error that occurred with `message` and `code` properties. The `code` values are strings prefixed with `android_` on Android and `ios_` on iOS. When in `State.Error`, calling `play()` reloads the current track and seeks to its last known time.
- Implement `TrackPlayer.retry()` which retries the current track when it stopped playing due to a playback error.
- Implement `State.Loading` which denotes the initial loading phase of a track and deprecate `State.Connecting`
- `State.Buffering` is emitted whenever the player is buffering data, no matter whether playback is paused or not
- Implement `TrackPlayer.getPlayWhenReady()` and `TrackPlayer.setPlayWhenReady(playWhenReady: boolean)` (where `TrackPlayer.setPlayWhenReady(false)` is the equivalent of `TrackPlayer.pause()` and `TrackPlayer.setPlayWhenReady(true)` is the equivalent of `TrackPlayer.play()`)
- Implement `usePlayWhenReady()` hook which returns the up to date state of  `TrackPlayer.getPlayWhenReady()`
- Implement TrackPlayer.setQueue(tracks) which clears the current queue and add the supplied tracks to the now empty queue
- Deprecate the confusingly names `TrackPlayer.getCurrentTrack()` which returns the current track index and replace it with `TrackPlayer.getActiveTrackIndex()`
- Implement `TrackPlayer.getActiveTrack()` which returns the active track object
- Deprecate `Event.PlaybackTrackChanged` which contained confusingly named parameters and add `Event.PlaybackActiveTrackChanged` to replace it. The new event also includes the full track objects for the newly active and last tracks.
- Add `useActiveTrack` hook which keeps track of the currently active track using `TrackPlayer.getActiveTrack()` and `Event.PlaybackActiveTrackChanged`
- Implement `TrackPlayer.seekBy(offset)`
- (breaking) Have the `usePlaybackState()` hook return undefined initially before it has finished retrieving the current state. Before it was incorrectly returning `State.None` which means no track is loaded.
- Add back `TrackPlayer.stop()` but change the behavior so the queue is not cleared when called. Behavior is the same as `TrackPlayer.pause()` where `playWhenReady` becomes `false`, but instead of just pausing playback, the item is unloaded – causing any further loading / buffering to stop. See https://github.com/doublesymmetry/react-native-track-player/issues/1786
- Change faulty implementation of `State.Stopped` to now mean the player is in a stopped state due to `TrackPlayer.stop()` being called
- Implement `State.Ended` which indicates playback stopped due to the end of the queue being reached (still need to implement the iOS side of this)
- add `autoHandleInterruptions` option to `TrackPlayer.setupPlayer` which automatically handles toggling playback / volume for audio interruptions
- add `androidAudioContentType` to `TrackPlayer.setupPlayer`
- rename MetadataOptions to UpdateOptions & deprecate MetadataOptions
- move `alwaysPauseOnInterruption` `TrackPlayer.updateOptions` option into `android` and deprecate previous usage

### Fix
- Queue methods that modify the queue no longer mutate `playWhenReady`, so when TrackPlayer is in playing mode it stays so and opposite (https://github.com/doublesymmetry/react-native-track-player/issues/1691)
- Calling `TrackPlayer.remove(index)` on the current track no longer throws, instead it activates the next track (or the first track if there is no next track).
- Calling `TrackPlayer.jump(index)` where `index` is the index of the active track no longer throws, instead it causes the track to be restarted.
- Fix clearing of now playing info when there is no longer an active track

### Example App
- (Android) When closing and reopening the app, have the displayed track sync to that of the music service

### TODO
- test seeking beyond possible time
- Can we avoid throwing in `TrackPlayer.setupPlayer()`?


bradleyflood:
Amazing work mate. Well done!

jspizziri:
@puckey I won't look a gift-horse in the mouth, but is it possible to deal with playback-track-changed events on an entirely separate PR? This one has become a behemoth. If not I'm sure we can handle it, it'll just be trickier.

These playback events specifically are ones that I'd like to set up tests for, as they're an absolute nightmare to test manually.

puckey:
> @puckey I won't look a gift-horse in the mouth, but is it possible to deal with playback-track-changed events on an entirely separate PR? This one has become a behemoth. If not I'm sure we can handle it, it'll just be trickier.
> 
> These playback events specifically are ones that I'd like to set up tests for, as they're an absolute nightmare to test manually.

Sorry, I completely missed your message. For my workflow the easiest is to put everything in this pr... If it does turn out to be impossible to deal with, we can see if we might want to extract separate prs from here – although that can easily lead to bugs.. (If only Github supported doing multiple prs that depend on each other.)

One thing we might want to consider is to use this opportunity to introduce tests on the `RNTP` side, for checking the 
api on iOS and Android. Sadly, I don't think I can spend another month working on the library full time as I have other tasks that need to be picked up. But when a test environment is added, I am happy to help add tests for everything.

android/src/main/java/com/doublesymmetry/trackplayer/module/MusicModule.kt
```diff
@@ -110,7 +148,7 @@ class MusicModule(reactContext: ReactApplicationContext) : ReactContextBaseJavaM
         constants["STATE_PAUSED"] = State.Paused.state
         constants["STATE_STOPPED"] = State.Stopped.state
         constants["STATE_BUFFERING"] = State.Buffering.state
-        constants["STATE_CONNECTING"] = State.Connecting.state
+        constants["STATE_LOADING"] = State.Loading.state
```
bradleyflood, line 151: Should the Error state be here too?

docs/docs/api/functions/queue.md
```diff
@@ -9,10 +9,10 @@ added track index. If no tracks were added it returns `void`.
 | Param          | Type     | Description   |
 | -------------- | -------- | ------------- |
 | tracks         | `array` of [Track Object](../objects/track.md) or a single one | The tracks that will be added |
-| insertBeforeIndex | `number` | The index of the track that will be located immediately after the inserted tracks. Set it to `null` to add it at the end of the queue |
+| insertBeforeIndex | `number` | (Optional) The index of the track that will be located immediately after the inserted tracks. By default the track is added to the end of the queue |
 
 ## `remove(tracks)`
-Removes one or more tracks from the queue.
+Removes one or more tracks from the queue. If the current track is removed, the next track will activated. If the current track was the last track in the queue, the first track will be activated.
```
bradleyflood, line 15: I'm wondering, what happens when there are no tracks left in the queue? Does the player return to "Ready" state? or what?

example/src/App.tsx
```diff
@@ -6,28 +6,30 @@ import {
...
 
 const App: React.FC = () => {
-  const track = useCurrentTrack();
+  const track = useActiveTrack();
   const [isPlayerReady, setIsPlayerReady] = useState<boolean>(false);
 
   useEffect(() => {
-    async function run() {
+    let unmounted = false;
+    (async () => {
```
bradleyflood, line 20: This IIFE is nice. But I reckon it will be confusing to those less experienced or knowledgable in JS. I can imagine people saying to themselves "What the heck is happening her??"

I reckon demos/examples should always be super easy and cater to less level developers where possible.

example/src/App.tsx
```diff
@@ -6,28 +6,30 @@ import {
...
 import { QueueInitialTracksService, SetupService } from './services';
 
 const App: React.FC = () => {
-  const track = useCurrentTrack();
+  const track = useActiveTrack();
   const [isPlayerReady, setIsPlayerReady] = useState<boolean>(false);
 
   useEffect(() => {
-    async function run() {
+    let unmounted = false;
```
bradleyflood, line 19: Why do we need this? If the useEffect dependency array is empty then this should only fire once anyway right?

And the unmount callback will destroy the `unmounted` variable state anyway, won't it?

example/src/services/PlaybackService.ts
```diff
@@ -25,14 +25,12 @@ export async function PlaybackService() {
 
   TrackPlayer.addEventListener(Event.RemoteJumpForward, async (event) => {
     console.log('Event.RemoteJumpForward', event);
-    const position = (await TrackPlayer.getPosition()) + event.interval;
-    TrackPlayer.seekTo(position);
+    TrackPlayer.seekBy(event.interval);
```
bradleyflood, line 28: 😍😍

src/hooks.ts
```diff
@@ -31,13 +39,43 @@ const usePlaybackStateWithoutInitialValue = () => {
...
-export const usePlaybackState = () => {
-  const state = usePlaybackStateWithoutInitialValue();
-  return state ?? State.None;
+/**
+ * Get current playback state and subsequent updates.
+ *
+ * Note: While it is fetching the initial state from the native module, the
+ * returned state property will be `undefined`.
+ * */
+export const usePlaybackState = (): PlaybackState | { state: undefined } => {
```
bradleyflood, line 51: Why the merge? Having the hooks in seperate files does feel tidier. And cleaner once we start adding unit tests. I'm all for smaller & simpler files.

bradleyflood (commented):
Epic work mate. Very nice 👍

example/src/App.tsx
puckey: For the case where the component was unmounted in the middle of two async calls.

example/src/services/PlaybackService.ts
puckey: The small things : )

(Would be nice to go even further and have an option where we could just have the native side handle these cases completely – so we wouldn't have to add event listeners for `Event. RemoteJumpForward` etc.)

android/src/main/java/com/doublesymmetry/trackplayer/module/MusicModule.kt
puckey: Yes it would need to be added – but actually we aren't using these constants anywhere anymore.. Perhaps they can be removed? @dcvz do you have thoughts?

docs/docs/api/functions/queue.md
puckey: Good question. Currently it probably depends on the platform. On Android I am guessing `State.Stopped`, because it is linked to exoplayer's `Player.STATE_ENDED` which means `The player finished playing all media.`. On iOs it probably would be [`State.None`](https://react-native-track-player.js.org/docs/api/constants/state#:~:text=Description-,None,-State%20indicating%20that): 

> State indicating that no media is currently loaded

I created this issue to deprecate `State.Stopped` and replace it with `State.Ended` https://github.com/doublesymmetry/react-native-track-player/issues/1791

But I am also working on adding back `TrackPlayer.stop()` as per https://github.com/doublesymmetry/react-native-track-player/issues/1786 which might mean a breaking change for `State.Stopped` to mean the loading of the current active track has been stopped.

All in all it also comes down to us needing cross platform tests in place to make sure event order can be relied upon to be equal across platforms.

puckey:
I added sleep timer functionality in the following commit: https://github.com/puckey/react-native-track-player/commit/172a6080aa0c1f787024430bf3a0520ba290c9c6

I would be interested to hear if the android implementation is up to scratch. I have seen other implementations of sleep timers which use an alarm manager to complete the timer, but I couldn't figure out why this would be necessary and just used a kotlinx flow for it instead. If there is a good reason to opt for alarm manager (with `<uses-permission android:name="android.permission.SCHEDULE_EXACT_ALARM" />` permission), I am happy to take another look at the implementation. If not, perhaps the SleepTimer class could go into KotlinAudio..

bradleyflood:
@puckey This is some nice work mate, but can you split it out into multiple pull requests? It's massive right now and I reckon that because of its size it will take much longer to get it reviewed and merged by the maintainers.

bradleyflood:
If you could break it into bite size chunks then we can get it merged.

Is there something I can do to help?

bradleyflood:
@dcvz This PR fixes a lot of bugs and outstanding issues. What is your recommendation here to move this forward? How can I help?

Thank you 🙏

kesha-antonov:
Looks very interesting. I'd like to have Ended event
But I agree with the @bradleyflood that it would be better to split this PR into multiple to be able to check it

jspizziri:
@puckey I found the following bug upon testing/review:

The `Pause` button never reverts back to a `Play` button upon the completion of the final track. This seems to be caused by how the new `playWhenReady` property is being used to determine the state of playback. `playWhenReady === true` however there is nothing left to play.

Are we certain using `playWhenReady` like we are now is appropriate? If so, perhaps setting `playWhenReady` back to false when the final track finishes is the solution? Although it seems strange to mutate it internally like that if the user is expecting it to stay set.

src/interfaces.ts
```diff
@@ -514,9 +613,29 @@ export interface PlaybackProgressUpdatedEvent extends Progress {
...
+| { time: number }
+| { sleepWhenPlayedToEnd: boolean }
+| null
+
+export type SleepTimerChangedEvent = SleepTimer;
+
+export type PlaybackState =
+  | {
+      state:
+        | State.None
```
jspizziri, line 626: @puckey is there a reason that `State.Ended` is not included in this list? My assumption is that you would either get an error state, or a state with any of the other possible values.

If it's an oversight I can add it to my open PR.

puckey:
> @puckey I found the following bugs upon testing/review:

Thanks for looking over the pr!

> ## `playWhenReady` behavior (Android & iOS)
> The `Pause` button never reverts back to a `Play` button upon the completion of the final track. This seems to be caused by how the new `playWhenReady` property is being used to determine the state of playback. `playWhenReady === true` however there is nothing left to play.
> 
> Are we certain using `playWhenReady` like we are now is appropriate? If so, perhaps setting `playWhenReady` back to false when the final track finishes is the solution? Although it seems strange to mutate it internally like that if the user is expecting it to stay set.
> 
> **EDIT:** I think this actually is a non-issue. I updated the `example` projects `PlayPauseButton.tsx` to check for `Error, Paused, Ended, & Stopped` states and that appropriately updates the UI. I'll send a PR for that one.
> 
> Here's [the PR](https://github.com/puckey/react-native-track-player/pull/2) and here's [the specific commit](https://github.com/puckey/react-native-track-player/pull/2/commits/7109c76d7244c852fb00fc7385c163a71d88147a)

Hmm I think `playWhenReady` should indeed become false when the last item of the queue has stopped playing. I will look into this.

> ## `skipToNext` on final item
> Calling `skipToNext` when on the final item in the playlist resets the playback position to the beginning of the track, rather than doing nothing which is what I would expect. `skipToPrev` at the beginning of the playlist does this too, but that is to be expected.

Good point. Will look into this also.

> ## Duplicate `State.Playing`
> I seem to consistently be seeing duplicate `State.Playing` events, which I wouldn't entirely expect. See an example below:
> 
> ```
> ...
> LOG  Event.PlaybackActiveTrackChanged ...
> LOG  Event.PlaybackState {"state": "buffering"}
> LOG  Event.PlaybackState {"state": "playing"}
> LOG  Event.PlaybackState {"state": "playing"}
> LOG  Event.PlaybackState {"state": "loading"}
> LOG  Event.PlaybackActiveTrackChanged ...
> LOG  Event.PlaybackState {"state": "buffering"}
> LOG  Event.PlaybackState {"state": "playing"}
> LOG  Event.PlaybackState {"state": "playing"}
> LOG  Event.PlaybackState {"state": "loading"}
> LOG  Event.PlaybackActiveTrackChanged ...
> LOG  Event.PlaybackState {"state": "buffering"}
> LOG  Event.PlaybackState {"state": "playing"}
> LOG  Event.PlaybackState {"state": "playing"}
> LOG  Event.PlaybackState {"state": "loading"}
> LOG  Event.PlaybackActiveTrackChanged ...
> ...
> ```

Are you seeing these on both iOS and Android?

src/interfaces.ts
puckey: See https://github.com/doublesymmetry/react-native-track-player/pull/1713/commits/9e9c2caeb88e60ad202224d86533da470391387b

jspizziri:
@bradleyflood if the offer to help still stands here are some areas that we could use help:

1. Reviewing/updating the docs (including the migration docs). [I've created a PR starting that process here](https://github.com/doublesymmetry/react-native-track-player/pull/1887)
2. Installing the `next` release (which should go out tonight) and seeing what breaks in your project so we can figure out what needs to be refined.

Issue #1709, opened 2022-09-08
Duration value is always '0' in useProgress hooks 


kyimoemin:
**Describe the Bug**
the value of duration that returned from useProgress hooks is always '0' in iOS simulator. Android simulator don't have that problem, the duration is showing correctly in android.

**Steps To Reproduce**
start the iOS simulator and run the following code.

**Code To Reproduce**

```js
const {position,buffered,duration} = useProgress();

console.log(position,buffered,duration);
/**
 the duration value is always 0,
e.g.,
 position:12
 duration:0
*/
```

**Environment Info:**

System:
    OS: macOS 12.5.1
    CPU: (8) x64 Intel(R) Core(TM) i5-8257U CPU @ 1.40GHz
    Memory: 47.27 MB / 8.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 14.17.0 - /usr/local/bin/node
    Yarn: Not Found
    npm: 6.14.13 - /usr/local/bin/npm
    Watchman: 2021.06.07.00 - /usr/local/bin/watchman
  Managers:
    CocoaPods: 1.11.3 - /usr/local/bin/pod
  SDKs:
    iOS SDK:
      Platforms: DriverKit 21.4, iOS 15.5, macOS 12.3, tvOS 15.4, watchOS 8.5
    Android SDK:
      API Levels: 27, 28, 29, 30, 31
      Build Tools: 29.0.2, 30.0.3, 31.0.0
      System Images: android-29 | Intel x86 Atom, android-29 | Intel x86 Atom_64, android-30 | Google APIs Intel x86 Atom
      Android NDK: Not Found
  IDEs:
    Android Studio: 2020.3 AI-203.7717.56.2031.7784292
    Xcode: 13.4.1/13F100 - /usr/bin/xcodebuild
  Languages:
    Java: 1.8.0_282 - /usr/bin/javac
  npmPackages:
    @react-native-community/cli: Not Found
    react: 17.0.1 => 17.0.1 
    react-native: 0.64.2 => 0.64.2 
    react-native-macos: Not Found
  npmGlobalPackages:
    *react-native*: Not Found

`react-native-track-player`:`3.1.0`
device: simulator
running OS: iOS

puckey:
@kyimoemin please provide an example audio file which gives this issue

puckey:
See https://github.com/doublesymmetry/react-native-track-player/pull/1723 for a fix candidate

[...]

PR #1723, opened 2022-09-13
fix(hooks): useProgress & usePlayback hooks

puckey:
- (breaking) fix: usePlaybackState was incorrectly returning `State.None` by default. Now it returns undefined while it is waiting on the async call to TrackPlayer to resolve. This also fixes the following issue where duration was incorrectly being reported as `0`: https://github.com/doublesymmetry/react-native-track-player/issues/1709
- optimization: avoid unnecessarily reevaluating the useEffect on every player state change by moving the `playerState === State.None` outside of the effect and including it as a dependency instead

### DIFF
--- a/src/hooks.ts
+++ b/src/hooks.ts
@@ -3,8 +3,7 @@ import { useEffect, useRef, useState } from 'react';
 import { Event, EventsPayloadByEvent, Progress, State } from './interfaces';
 import TrackPlayer from './trackPlayer';
 
-/** Get current playback state and subsequent updates  */
-export const usePlaybackState = () => {
+const usePlaybackStateWithoutInitialValue = () => {
   const [state, setState] = useState<State | undefined>(undefined);
   useEffect(() => {
     let mounted = true;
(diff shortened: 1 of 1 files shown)


jspizziri:
@dcvz is this a significant enough breaking change to hold back the release?

@puckey is there a way of accomplishing this that doesn't introduce a breaking change?

src/hooks.ts
```diff
@@ -131,7 +131,7 @@ export function useProgress(updateInterval = 1000) {
     return () => {
       mounted = false;
     };
-  }, [playerState, updateInterval]);
+  }, [isNone, updateInterval]);
 
   return state;
```
jspizziri, line 136: @puckey would doing a:

```
return state || State.None;
```

Allow us to avoid the breaking change here?

jspizziri (changes requested):
@puckey please see the comment I left. I'd like to find a way to do this without a breaking change. As we'd need to hold off until the next major version if not.

gavrichards:
I'm keen to see this fixed asap, so I'll share that @puckey and I were discussing this in the discord yesterday, and the initial fix proposed which worked without any breaking changes is mentioned here:
https://discord.com/channels/567636850513018880/600714011318681610/1019166051461779467

src/hooks.ts
puckey: Not in this way, but there is slightly more convoluted way to fix this.. Will take a look at it later

puckey:
How about something like this @jspizziri: https://github.com/doublesymmetry/react-native-track-player/pull/1723/commits/b895b9c9a183af7b25b8cc13c242501052316ed0

PR #1729, opened 2022-09-14
docs(ts, ios): add inline documentation and docs for iOS constants

jspizziri:


jspizziri:
@dcvz I can't merge because of the tests that don't run (I haven't seen those before).

Issue #1598, opened 2022-07-14
Queue End Reached firing too early on iOS


rpadilla6:
**Describe the Bug**
Hi there, I'm writing some functionality to track when a queue has ended in order to reset to the beginning of the first track. I ran into an issue where if the queue is longer than one track, the `PlaybackQueueEnded` is fired too early, which breaks the experience for me on iOS. This is not the case on Android.

**Steps To Reproduce**

1. Add tracks to a queue, (minimum of two)
2. Play first track, you can seek to the end of it to go faster.
3. When that song finishes, you can see that the track player event `PlaybackQueueEnded` gets fired on iOS.

**Code To Reproduce**
```
useTrackPlayerEvents([Event.PlaybackQueueEnded], async (queueEndedPayload) => {
    console.log({ queueEndedPayload });
  });
```

**Environment Info:**
Paste the results of `npx react-native info`:
```
System:
    OS: macOS 12.4
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 261.01 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 14.18.0 - ~/.nvm/versions/node/v14.18.0/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 6.14.15 - ~/.nvm/versions/node/v14.18.0/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  Managers:
    CocoaPods: 1.11.3 - /usr/local/bin/pod
  SDKs:
    iOS SDK:
      Platforms: DriverKit 21.4, iOS 15.5, macOS 12.3, tvOS 15.4, watchOS 8.5
    Android SDK:
      API Levels: 23, 25, 26, 28, 29, 30, 31
      Build Tools: 26.0.3, 27.0.3, 28.0.3, 29.0.2, 29.0.3, 30.0.2, 31.0.0, 32.0.0
      System Images: android-26 | Google APIs Intel x86 Atom, android-28 | Intel x86 Atom_64, android-28 | Google APIs Intel x86 Atom, android-29 | Google Play Intel x86 Atom
      Android NDK: Not Found
  IDEs:
    Android Studio: 2021.1 AI-211.7628.21.2111.8193401
    Xcode: 13.4.1/13F100 - /usr/bin/xcodebuild
  Languages:
    Java: 11.0.12 - /usr/bin/javac
[...]

PR #1750, opened 2022-09-23
fix(ios, events): fix an issue with PlaybackQueueEnded resulting from…

jspizziri:
… a race condition

https://github.com/doublesymmetry/react-native-track-player/issues/1598

### DIFF
--- a/ios/RNTrackPlayer/RNTrackPlayer.swift
+++ b/ios/RNTrackPlayer/RNTrackPlayer.swift
@@ -26,7 +26,6 @@ public class RNTrackPlayer: RCTEventEmitter, AudioSessionControllerDelegate {
         super.init()
 
         audioSessionController.delegate = self
-        player.event.playbackEnd.addListener(self, handleAudioPlayerPlaybackEnded)
         player.event.receiveMetadata.addListener(self, handleAudioPlayerMetadataReceived)
         player.event.stateChange.addListener(self, handleAudioPlayerStateChange)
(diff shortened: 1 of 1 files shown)


jspizziri:
This fixes a race condition affecting the `PlaybackQueueEnded` event on iOS. Previously we were firing this event based off of the `playbackEnd` event. When that event was fired we were checking to see if the queue had more items left to play by evaluating `player.nextItems.count`. However, this property was evaluating incorrectly due to a race condition. By the time the handler gets called, the `nextItem.count` has already been updated by the `player` so it can be off by a value of `1`.

This was causing the `PlaybackQueueEnded` event to be fired on both the `n` **AND** the `n - 1` track transitions.

I fixed it by moving the event firing to the `PlaybackTrackChanged` handler, which is more consistent.

jspizziri:
@dcvz running one more test on this.

jspizziri:
@dcvz so everything was working fine, but I decided to change `player.currentIndex` to `index` when firing the event. I wasn't seeing it but I realized this could be a race condition as well. It's safer to do it this way as that's the value we're using in the `track-changed` event too.

Sorry for triggering another 20min iOS build :/. Should've thought of that on initial implementation.

jspizziri:
@dcvz ready for another approval

trm217:
@dcvz Could you please consider releasing patch versions when ever such rather substancial bugs are fixed?

PR #1751, opened 2022-09-23
fix(android): fix state constants

puckey:


jspizziri (approved):
I independently made and verified these changes.

PR #1753, opened 2022-09-23
refactor(src): refactor the file structure for better organization

jspizziri:


jspizziri:
@dcvz @mpivchev @puckey I'd appreciate all of your feedback on this one.

puckey:
Yes, this looks good! I wonder however if replacing the default object export with individual exports is not a breaking change..

jspizziri:
@puckey fair question. We are instructing people to import directly from the root, and not use deep imports from the `./trackPlayer` & `./interfaces` files. It would break things if people were doing deep exports like that for sure, however, everything is still exposed from the root in the exact same way it was before.

It also could just be something that lands in v4 too.

jspizziri:
I've created a new PR based on @puckey 's work here https://github.com/puckey/react-native-track-player/pull/1

This one will likely get closed in favor of that one.

jspizziri:
This ended up being done as part of https://github.com/doublesymmetry/react-native-track-player/pull/1713

PR #1755, opened 2022-09-27
Allow void return value from PlaybackService ServiceHandler

bradleyflood:
I don't need my playback-service to be async, as I've outsourced all the work to another service file.

Though we should keep the documentation and examples demonstrating an async service, it doesn't need to be enforced right? The callback is fired fine regardless of it's return type.

See my playback-service screenshot...

Thoughts?

Screen Shot 2022-09-27 at 11 31 06


jspizziri:
@bradleyflood Did you test the example when using a sync callback?

Issue #1740, opened 2022-09-20
[Feature Request] Rename `reset` to `resetPlaylist` to more accurately describe what the function does


jspizziri:
Created from https://github.com/doublesymmetry/react-native-track-player/issues/1637#issue-1333490271

> Rename reset to resetPlaylist to more accurately describe what the function does. Rename TrackPlayer.stop to TrackPlayer.resetPlaylist as well.
> Note: More descriptive.

bradleyflood:
Hmm. And it also stops playback, which isn't explicitly clear in `resetPlaylist`.

Also, this module doesn't use the word `playlist` much does it? Usually Queue and Track , ie `getQueue()` returns `tracks`, so perhaps it should be renamed to `resetQueue()` or `resetQueueAndStop()` hah

Or, just leave it as `reset()` :P

anti-duhring:
> Hmm. And it also stops playback, which isn't explicitly clear in `resetPlaylist`.
> 
> Also, this module doesn't use the word `playlist` much does it? Usually Queue and Track , ie `getQueue()` returns `tracks`, so perhaps it should be renamed to `resetQueue()` or `resetQueueAndStop()` hah
> 
> Or, just leave it as `reset()` :P
> 
> @anti-duhring 

I like `resetQueue()`, it makes more sense. Hahaha
 
So `resetQueue()` has to reset the playlist but keep playing and `resetQueueAndStop()` has to reset and stop, right? 

I will fix it

jspizziri:
@bradleyflood @anti-duhring ,

I think @mpivchev is proposing that we change the word queue to playlist in all contexts. There's a feature request somewhere relating to that.

anti-duhring:
> @bradleyflood @anti-duhring ,
> 
> I think @mpivchev is proposing that we change the word queue to playlist in all contexts. There are a few other features relating to that.

Oh, I got it. So I keep `resetPlaylist()`? 

It will be something like that:
- `resetPlaylist()` reset the queue but not stops playback
- `resetPlaylistAndStop()` reset the queue and stops playback

@jspizziri

puckey:
we have `TrackPlayer.getQueue()` (and soon we will have `TrackPlayer.setQueue()`)

so how about `TrackPlayer.clearQueue()` which does nothing but remove all tracks from the queue

puckey:
(I guess the same could be done using `TrackPlayer.setQueue([])` soon though)

puckey:
[...]

PR #1756, opened 2022-09-27
feat(ts): rename `reset` to `resetPlaylist` to more accurately describe what the function does

anti-duhring:
#1740 

BREAKING CHANGE:

The `reset` method has been renamed `resetPlaylist`.


mpivchev:
Thanks for your PR :) This is a breaking change, so it will be merged in 4.0

jspizziri:
@anti-duhring thanks for the PR, we'll probably do some review with you on this once we're a bit closer to v4 (which hopefully isn't too far off).

PR #1758, opened 2022-09-27
fix: example app to RN v0.69.6 and appKilledPlaybackBehavior to StopPlaybackAndRemoveNotification

mmmoussa:
This is not meant to be merged necessarily (the update to the RN version could be but the appKilledPlaybackBehavior doesn't need to be). The goal of this PR is to demonstrate that the `appKilledPlaybackBehavior` option of `StopPlaybackAndRemoveNotification` does not work correctly.

Repro steps:
1. Install the app on your device and start playing music
2. Close the app and note that the music stopped playing and the notification disappeared
3. Open the app and start playing music
4. Close the app and note that the music continued playing and the notification did not disappear

RN version update followed https://react-native-community.github.io/upgrade-helper/?from=0.68.1&to=0.69.6


jspizziri:
@mmmoussa I'd prefer to see an example that has only the isolated repro, rather than also containing a RN upgrade. Adding many things into the mix makes it more complicated, unless you're saying the issue only happens in RN `0.69.6`

jspizziri:
It looks like the only thing you changed was configuring the behavior. I'll test that against the existing app.

mmmoussa:
I couldn't get the release build working without upgrading React Native and so that's why I've got it here, but if the release build is working for you as is then all you need to do is change `appKilledPlaybackBehaviour`. You can even test with a debug build and you should experience the same issue.

jspizziri:
Closing this in favor of https://github.com/doublesymmetry/react-native-track-player/issues/1759

PR #1762, opened 2022-09-28
fix: notification not removed a second time

### LINKED ISSUE #1759: StopPlaybackAndRemoveNotification fails on second app exit
Origninally reported by @mmmoussa here https://github.com/doublesymmetry/react-native-track-player/pull/1758

Repro steps:

1. Set `appKilledPlaybackBehavior` to `StopPlaybackAndRemoveNotification`
2. Start playing a track
3. Close the app and note that the music stopped playing and the notification disappeared
4. Open the app and start playing a track
5. Close the app and note that the track continued playing and the notification did not disappear

mpivchev:
This fixes #1759


android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
```diff
@@ -400,6 +401,7 @@ class MusicService : HeadlessJsTaskService() {
                         }
 
                         stopSelf()
+                        exitProcess(0)
```
mpivchev, line 404: `onTaskRemoved` is only called once, which means that the notification would not disappear next time the app is removed from recents. This is probably related to the process remaining running.

`exitProcess` ensures the process stops. Not the most elegant solution, but it works.

mmmoussa:
Patched this into my project and confirmed it fixed the issue as well.

android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
mmmoussa: I discovered a serious flaw with this approach. Calling `TrackPlayer.reset()` causes this to be triggered and kills the app, which means there is no longer any way to stop the player during regular use.

jspizziri: @mmmoussa https://github.com/doublesymmetry/react-native-track-player/issues/1768

dimbslmh: @mpivchev I noticed you are calling both startService() and bindService(). 
https://github.com/doublesymmetry/react-native-track-player/blob/e742959cc1d697d69daeade39830155bce2ccde7/android/src/main/java/com/doublesymmetry/trackplayer/module/MusicModule.kt#L181-L184

Maybe this explanation on SO helps on how to stop the service? https://stackoverflow.com/a/17178363/3229423

PR #1770, opened 2022-10-04
docs(api): clarify type of `.remove()` and `.add()` methods

bradleyflood:
Just to make this clearer.

I would rather use TS typing and just have `number[]` but I think that won't be clear enough to everyone. Thoughts?


bradleyflood:
Agreed. I'll create a new PR with all of the changes required.

PR #1789, opened 2022-10-10
internal: Ensure to set AudioSession configuration on iOS

Spice-Z:
PR for https://github.com/doublesymmetry/react-native-track-player/issues/1790

Call `AVAudioSession.sharedInstance().setCategory` every time `play` method is called.

### DIFF
--- a/ios/RNTrackPlayer/RNTrackPlayer.swift
+++ b/ios/RNTrackPlayer/RNTrackPlayer.swift
@@ -19,6 +19,10 @@ public class RNTrackPlayer: RCTEventEmitter, AudioSessionControllerDelegate {
     private let player = QueuedAudioPlayer()
     private let audioSessionController = AudioSessionController.shared
     private var shouldEmitProgressEvent: Bool = false
+    private var sessionCategory: AVAudioSession.Category = .playback
+    private var sessionCategoryMode: AVAudioSession.Mode = .default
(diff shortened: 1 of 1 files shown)


bradleyflood:
Hmm. I assume most RNTP users don't have this problem or need. 

Can't you just call 'setup()' again when your app is returning to this module?

bradleyflood:
I say that because my app doesn't use any other audio library so I don't have this problem. 

Yet my users do call the pause and play methods hundreds of times per session. So this PR change seems like an intrusive change that would be impacting the majority of RNTP only users to solve a specific use-case problem. 

I feel like there has to be a better way?

Spice-Z:
> Can't you just call 'setup()' again when your app is returning to this module?

I tried, but app will crash when setup is called multiple time. (I have to destroy player, but RNTP doesn't have `destory` method now)


And in trial PR, @puckey  suggested [better idea](https://github.com/doublesymmetry/react-native-track-player/issues/1790#issuecomment-1274238673)!

> How about adding iosCategory, iosCategoryMode, iosCategoryOptions as options to TrackPlayer.updateOptions() instead?


@bradleyflood 
How do you think about this?
I think extending updateOptions is very good.

bradleyflood:
@Spice-Z Good idea!

Spice-Z:
Ok, I will make new PR for extending updateOptions!

PR #1835, opened 2022-11-28
fix: don't overwrite previous progress interval when it's unspecified in updateOptions

fivecar:
Previous code would clear/overwrite your progress interval if you called `updateOptions` without re-specifying the same interval. This change conforms the code to the documentation by preserving your previous progress interval if a new call to `updateOptions` doesn't specify a new interval.


bradleyflood (commented):
LGTM

fivecar:
@dcvz : Just a gentle ping on whether this looks good to merge...

fivecar:
@dcvz / @bradleyflood  : The platform validation is failing on not finding `rexml/document`, but it's really quite unlikely that this PR caused it. Could you LMK if this is a known issue that's fixable?

fivecar:
Fix for break submitted in #1869. Once that lands, this should pass.

PR #1855, opened 2022-12-18
fix(rntrackplayer.swift): prevents overwriting of forward/backward secs

### LINKED ISSUE #1853: updateOptions overwrites previously set forwardJumpInterval
**Describe the Bug**
Although the docs say that `updateOptions` won't overwrite values that you omit, it actually does. You can see this in how [RNTrackPlayer.swift](https://github.com/doublesymmetry/react-native-track-player/blob/3a28d48f50c4b5ad54ef96ca2c871bf3c7f51b3a/ios/RNTrackPlayer/RNTrackPlayer.swift#L302) calls [Capabilities.swift](https://github.com/doublesymmetry/react-native-track-player/blob/3a28d48f50c4b5ad54ef96ca2c871bf3c7f51b3a/ios/RNTrackPlayer/Models/Capabilities.swift#L36). Basically, if you want to update capabilities and retain the ability to jump forward, you can't do so without also re-specifying `forwardJumpInterval` (because the code otherwise overwrites your previous interval with 15).

**Steps To Reproduce**
The following sequence is the simplest repro:
```Typescript
player.updateOptions({
  capabilities: [
    Capabilities.Play,
    Capability.Pause,
    Capability.SeekTo,
    Capability.JumpForward,
  ],
  forwardJumpInterval: 60,
});
...
player.updateOptions({
  capabilities: [
    Capabilities.Play,
    Capability.Pause,
    Capability.SeekTo,
    Capability.JumpForward,
  ],
});  
```
After this sequence, the `forwardJumpInterval` becomes 15, even though the docs claim that you only need to set the `MetadataOptions` that you wish to change.

**Code To Reproduce**
[see above]

**Replicable on Example App?**
Yes. Simply modify its code as suggested.

**Environment Info:**
System:
    OS: macOS 13.1
    CPU: (16) x64 Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
    Memory: 2.88 GB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.16.0 - ~/.nvm/versions/node/v16.16.0/bin/node
    Yarn: 2.4.3 - ~/.nvm/versions/node/v16.16.0/bin/yarn
    npm: 8.11.0 - ~/.nvm/versions/node/v16.16.0/bin/npm
    Watchman: Not Found
  Managers:
    CocoaPods: 1.11.3 - /Users/philipsu/.rvm/gems/ruby-2.7.5/bin/pod
  SDKs:
    iOS SDK:
      Platforms: DriverKit 22.1, iOS 16.1, macOS 13.0, tvOS 16.1, watchOS 9.1
    Android SDK: Not Found
  IDEs:
    Android Studio: 2020.3 AI-203.7717.56.2031.7784292
    Xcode: 14.1/14B47b - /usr/bin/xcodebuild
  Languages:
    Java: 17.0.4.1 - /usr/bin/javac
  npmPackages:
    @react-native-community/cli: Not Found
    react: 18.0.0 => 18.0.0 
    react-native: 0.69.3 => 0.69.3 
    react-native-macos: Not Found
  npmGlobalPackages:
    *react-native*: Not Found

`react-native-track-player`: **"^3.2.0"**

Real device? Or simulator? **Both**
What OS are you running? **iOS 12.4**

**How I can Help**
What can you do to help resolve this?
I don't know Swift, but I can attempt a fix and send it along

Have you investigated the underlying JS or Swift/Android code causing this bug?
Yes. It's pretty clear where the error is.

Can you create a Pull Request with a fix?
I might be able to, but will need help with Swiftage.

fivecar:
The previous code overwrote forward/backward skip seconds when you subsequently called updateOptions without re-specifying the same secs. It's because the previous code, if you ever specified JumpForward or JumpBackward, would require you to also specify jumpForwardSecs and jumpBackwardSecs even if you had set them previously; otherwise, it'd overwrite each with 15s.

This fix prevents that problem by caching previously set values. Ultimately, this allows the library to follow its documentation, which claims that any subset of `MetadataOptions` can be specified without altering previously-set values.

fix #1853


ios/RNTrackPlayer/RNTrackPlayer.swift
```diff
@@ -298,12 +301,24 @@ public class RNTrackPlayer: RCTEventEmitter, AudioSessionControllerDelegate {
         }
         let capabilities = capabilitiesStr.compactMap { Capability(rawValue: $0) }
 
+        var forwardJumpInterval: NSNumber = lastForwardJumpInterval;
+        if let explicitForwardJumpInterval = options["forwardJumpInterval"] as? NSNumber {
+            forwardJumpInterval = explicitForwardJumpInterval;
+            lastForwardJumpInterval = forwardJumpInterval;
+        }
```
fivecar, line 308: Would love to learn if there's a better way to accomplish this (e.g. to read `player.remoteComments[.skipForward[interval]]` directly, but I'm no Swift expert.

fivecar:
Fix for break submitted in #1869 1869. Once that lands, this should pass.

ios/RNTrackPlayer/RNTrackPlayer.swift
```diff
@@ -298,12 +301,24 @@ public class RNTrackPlayer: RCTEventEmitter, AudioSessionControllerDelegate {
...
+
         player.remoteCommands = capabilities.map { capability in
-            capability.mapToPlayerCommand(forwardJumpInterval: options["forwardJumpInterval"] as? NSNumber,
-                                          backwardJumpInterval: options["backwardJumpInterval"] as? NSNumber,
-                                          likeOptions: options["likeOptions"] as? [String: Any],
-                                          dislikeOptions: options["dislikeOptions"] as? [String: Any],
-                                          bookmarkOptions: options["bookmarkOptions"] as? [String: Any])
+            capability.mapToPlayerCommand(forwardJumpInterval: forwardJumpInterval,
+                                            backwardJumpInterval: backwardJumpInterval,
```
jspizziri, line 318: @fivecar here I am being nitpicky, but I'd really prefer not to co-mingle simple stylistic changes (like the additional whitespace here) with fixes.

ios/RNTrackPlayer/RNTrackPlayer.swift
```diff
@@ -20,6 +20,9 @@ public class RNTrackPlayer: RCTEventEmitter, AudioSessionControllerDelegate {
     private let audioSessionController = AudioSessionController.shared
     private var shouldEmitProgressEvent: Bool = false
 
+    private var lastForwardJumpInterval: NSNumber = 0;
```
jspizziri, line 23: @fivecar my only big question here is, are we sure the default of `0` is appropriate here before anything is passed in? It looks like before it would've been set to `nil`? Can you verify that defaulting to `0` is functionally equivalent to how it was behaving before?

jspizziri (changes requested):

ios/RNTrackPlayer/RNTrackPlayer.swift
jspizziri: @fivecar (not a swift expert either) it might be cleaner to simply have a `forwardJumpInterval` and use a "Nil Coalescing" operator to set it every time. Something like this:

```swift
private var forwardJumpInterval: NSNumber = 0;

...

public func update(....) {
    forwardJumpInterval = options["forwardJumpInterval"] as? NSNumber ?? forwardJumpInterval
    backwardJumpInterval = options["backwardJumpInterval"] as? NSNumber ?? backwardJumpInterval
    ...
}


```

fivecar: Awesome suggestion, and so much cleaner! Thank you, @jspizziri !

ios/RNTrackPlayer/RNTrackPlayer.swift
fivecar: Agreed - I shouldn't have done this. Reverted!

ios/RNTrackPlayer/RNTrackPlayer.swift
fivecar: Great point -- I've now followed your previous suggestion, which is to preserve nil-hood. Thank you!

jspizziri (approved):

fivecar:
Apologies, @jspizziri : I pulled the latest Gemfile fix into the history of this diff so that it should pass all platform validation. That overwrote your approval... sorry about that.

jspizziri (approved):

jspizziri:
@dcvz this LGTM. The fix perserves jump interval settings between calls to `updateOptions` for iOS. I vote merge.

puckey:
(Looks good to me too)

fivecar:
@dcvz : ok - merge conflicts fixed. Would you mind re-approving and merging? Thanks!

ios/RNTrackPlayer/RNTrackPlayer.swift
jspizziri: @fivecar it looks like the nil coalescing operator got dropped in the cleanup. Was that intentional?

jspizziri:
@fivecar just left a question for you.

ios/RNTrackPlayer/RNTrackPlayer.swift
fivecar: Ack, no, not intentional. Thanks for catching, @jspizziri . Will re-up.

fivecar:
> @fivecar just left a question for you.

Thanks again, @jspizziri. I believe I've now restored everything back to correctness.

jspizziri (approved):

PR #1869, opened 2023-01-02
chore(example): update example Gemfile to fix platform-ios automation

### LINKED ISSUE #1868: Build automation breaking because of missing rexml
**Describe the Bug**
When PRs are submitted to this codebase, platform-ios verification fails with `LoadError - cannot load such file -- rexml/document`

**Steps To Reproduce**
Submit any PR to this repo

**Replicable on Example App?**
Yes

**How I can Help**
I'm going to submit a PR in attempt to fix this

fivecar:
We need to include `rexml` in the Gemfile in order to unbreak automation on the server. Otherwise, builds fail with things [like this](https://github.com/doublesymmetry/react-native-track-player/actions/runs/3723356292/jobs/6314820349).

This also removes a duplicate Gemfile. The original, under `example/ios`, was introduced by @dcvz . The more recent one, under `example/`, was introduced by @jspizziri . I've deleted the old one and updated the new one.

fix #1868


fivecar:
Awesome. @dcvz : This will unblock the other PRs that I've got active -- would you mind reviewing and merging so that those others can be unblocked? Thanks!

fivecar:
Updated as requested, @jspizziri!

I spent some time digging around, but I couldn't find the exact toolchain commit that changed this. But I suspect it's within our workflow yaml, where we run with `macos-latest`. Since we don't use `actions/setup-ruby@v1` to lock the CI to Ruby 2.x, we probably inherited the `macos-latest` Ruby bump to 3.x some time between last Oct and today.

This, BTW, suggests an alternative fix, which is that we lock Ruby to 2.x in the workflow via `setup-ruby`. LMK if you'd prefer that instead.

jspizziri:
@fivecar thanks for the explanation. LGTM.

jspizziri (approved):

PR #1886, opened 2023-01-24
feat(web): add a web implementation

jspizziri:
@dcvz here's the web implementation I've been working on. You can test it out via the following:

1. Pull the branch, install deps, & build
2. `yarn example web:start`


hirbod:
FYI: This has been unreviewed for about 8 months, but I decided to give it a try. I manually integrated all the web implementation parts into my app and made a few adjustments, such as preventing multiple Shaka instances. I also added support for `.pause` (which was previously missing) and ensured the emission of `PlaybackProgressUpdated` (also previously missing).

Aside from that, the web component is quite solid. I'd love to see this integrated into the core (with some modifications) as it's very helpful and the API is almost identical.

In case someone is interested how I did that (it's ugly, since I had to copy some source files because of some resolving issues). https://github.com/showtime-xyz/showtime-frontend/compare/staging...feat/track-player

I really like RNTP (beside of some hot reload issues) and its a solid library. I would love to see this merged, otherwise I'd be forced to maintain another fork, which is not helpful for the ecosystem imho.

jspizziri:
@hirbod please submit PR's for all the work that you've done to improve things to my fork and I'll review and merge them in there.

I've released my fork under [@5stones/react-native-track-player](https://www.npmjs.com/package/@5stones/react-native-track-player) and that's what I'm currently using (see the `4.0.0-web.x` tags. I'll be maintaining my fork until it gets merged into the mainline repo, so I'd recommend working with me there so you don't have to do it solo). I periodically rebase it on the upstream and publish a new version in addition to releasing new versions when I have bug fixes etc. As @dcvz said, the hope is that we'll ultimately get this reviewed and merged once 4.0.0 is out of RC, at which time I'd abandon my fork and we'd all switch over to mainline.

I'd greatly appreciate PR's for the issues you've identified to [`feat/web-3`](https://github.com/5-stones/react-native-track-player/tree/feat/web-3).

hirbod:
Thanks, @dcvz and @jspizziri, for the heads up. I'm glad to hear you guys have this on your radar.
@jspizziri, I regret not checking if your fork was actually published. Are you publishing from `main`? My codebase currently uses v3 for native, and, well, I suppose your implementation is from v4 :D

I can submit my minor bugfixes to your branch; they're straightforward.

jspizziri:
@hirbod 

> Are you publishing from main?

It's... complicated 😄. But no, I'm maintaining `feat/web-3` and publishing from `publish`. PR's are best sent to `feat/web-3`.

> My codebase currently uses v3 for native, and, well, I suppose your implementation is from v4 :D

Yep, I'm using v4 (which I'd recommend for you as well). There aren't that many BC changes between 3 & 4 and we've published a number of RC's. It's pretty close to being a formal v4 release and many folks, use it in prod.

Regardless, any PR's you can send will make this a smoother process to get merged and released in the mainline in the coming months. Thanks!

hirbod:
@jspizziri we started working on this feature again and I'd like to push my PRs to your repo. Would you mind to rebase to upstream and prepare a new version so I can incorporate some of the fixes?

hirbod:
@jspizziri, after pulling in your branch, it appears that most of the issues I had are now resolved. The pause function works correctly, and it seems that PlaybackProgressUpdates is also emitting correctly upon pausing.

However, I did encounter a couple of issues. Firstly, re-exporting was not possible (I've created a PR for this). Secondly, it seems that `setupPlayer` isn't preventing multiple instances. I'll create another PR to address this.

jspizziri:
@hirbod , I just did a rebase. It was a little messy due to all the upgrades done recently in the `example` app. It seems to be working however, please let me know if you find issues.

@dcvz just an FYI. Some of the improvements in the example app recently have caused some issues with web. Notably some things with the `@gorhom/bottom-sheet` `BottomSheetScrollViews` and some of the newer package version of `react-native-reanimated` (they're going through [a pretty big transition with their typescript support](https://github.com/software-mansion/react-native-reanimated/issues/4645)). I might've glossed over some other things.

jspizziri:
@dcvz we discussed this on our call, but to leave a paper trail.

The issues seemed to be specific with some of the new minor versions of the `react-native-reanimated` and `react-native-gesture-handler` packages. There are some internal structural issues and typescript issues that I believe are likely going to be resolved by them in the near future, so I pinned them to earlier minor versions so we don't have to worry about working around those issues for the time being. There's nothing fundamentally incompatible with them and web though, just existing bugs that will be resolved soon.

sajorahasan:
Hi @jspizziri thank you for your work on adding support for the web platform.

However I'm facing error "You must call `setupPlayer` prior to interacting with the player." which I have already done so its working fine in android ios too.

I run your example from your fork and it worked on web only difference I could find was you are using webpack bundler and I am using metro bundler.

Is there any way to make it work with metro bundler on web platform?

jspizziri:
@sajorahasan you're using the metro bundler on web? I didn't know people did that. You'll have to create a reproduction via a github repo that you share for me to be able to help. Although I must admit, it's probably not going to be a top priority of mine as it sounds like a niche thing.

sajorahasan:
Hi @jspizziri ohh I don't have much knowledge about react native web project. I am using ignite boilerplate and its have web platform support and every other packages used so far worked without much trouble with metro bundler.

I have uploaded demo project here: https://github.com/sajorahasan/WebTrackPlayerDemo 
If you have time please check that.

sajorahasan:
Hi @jspizziri I was able to fix the issue by replacing below import statement in Player.js

`const shaka = await import('shaka-player/dist/shaka-player.ui');`

This import statement was not resolving so I switched from import to require and moved it up top with other import statments like this.

`const shaka = require('shaka-player/dist/shaka-player.ui');`

I am not sure why import promise was not resolving with metro bundler.

jspizziri:
@sajorahasan That dynamic import is required for SSR contexts. I'd recommend looking at something like nextjs for web.

jspizziri:
@dcvz I just finished my audit of the `web`'s API and here's what I found:

### Currently Unimplemented

- `setQueue`

### Stubbed But Not Implemented

All the below stubbed methods [can be seen here](https://github.com/doublesymmetry/react-native-track-player/pull/1886/files#diff-d3e02c950065453cc885600bf38ebe8fdf1d7f8f9eff867c1c3c114bdb6dca6aR200):

- `updateMetadataForTrack`
- `updateNowPlayingMetadata`
- `updateOptions`

### Deprecations

There are several deprecated methods that I didn't implement. I'm personally not in favor of implementing currently deprecated APIs.

### Incorrectly Exposure

Currently, all of these getters/setters are public in the underlying Player class
but maybe shouldn't be, we could make them protected easily enough to lock them
down. However, they aren't exposed in the RNTP interface so getting at them would
be a chore if someone wanted to.

- `current`
- `playWhenReady`
- `state`



Once you've taken a look at this LMK how you'd like to proceed and I can raise any issues in the issue tracker that we need to and fix whatever you want.

web/TrackPlayer/Player.ts
```diff
@@ -0,0 +1,208 @@
...
+    this.element = document.createElement('audio');
+    this.element.setAttribute('id', 'react-native-track-player');
+    this.player = new shaka.Player(this.element);
+
+    // Listen for relevant events events.
+    this.player!.addEventListener('error', (error: any) => {
+      // Extract the shaka.util.Error object from the event.
+      this.onError(error.detail);
+    });
+    this.element.addEventListener('ended', () => this.onTrackEnded());
```
dcvz, line 72: nitpick: can't we just do this? Were you having binding issues? 

```suggestion
    this.element.addEventListener('ended',  this.onTrackEnded);
```

We probably can even generalize this, either updating the state directly in the closure here or having a general 

```ts
onStateUpdate(state: string) {
   switch (state) { ... }
}

this.element.addEventListener('playing', () => this.onStateUpdate('playing'))
```

dcvz (commented):
Great work @jspizziri! Overall clean implementation, have mostly only minor comments. I did however notice that Whip (HLS) fails to load the duration / progress. More urgently though skipping past Whip throws an error.  Is it perhaps the media type - next track is a live stream.

<details>
  <summary>Expand Error</summary>
  
Screenshot 2024-02-04 at 01 47 01

</details>



Also we probably (doesn't have to be part of this PR) want to leverage [navigator.mediaSession](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/mediaSession) to communicate with the OS and take advantage of media keys.

jspizziri:
> I did however notice that Whip (HLS) fails to load the duration / progress.

Interesting, I looked into this and (at least on my device) the HLS wasn't even playing. It looks like it has to do with [a limitation of the transmuxer used by shaka-player](https://github.com/shaka-project/shaka-player/issues/4392). I re-encoded the example HLS stream to be `MPEG2-TS` and it's working. Not sure what to make of this limitation itself or if things need to be reevaluated or if we just need to document it.

> More urgently though skipping past Whip throws an error. Is it perhaps the media type - next track is a live stream.

web/TrackPlayer/Player.ts
jspizziri: @dcvz yes, actually for some reason there are issues with binding. I'm not exactly sure why and I didn't dig into it a ton.

Regarding the modification to the `onStateUpdate` pattern. Which direction are you looking to go with this?

### Option 1

```ts
onStateUpdate(state: string) {
   switch (state) { 
     ...
     case 'ended':
        this.onTrackEnded();
        break;
     ...
   }
}
```

### Option 2

```ts
onStateUpdate(state: string) {
   switch (state) { 
     ...
     case 'ended':
        // directly add logic for the end state
        break;
     ...
   }
}
```

My sense is that you're going for Option 2? However, this will make extending state change behavior a little trickier potentially, as if you want to completely override a behavior you'll need to be aware of all the other events logic.

Let me know which direction you'd like to go and I'll send a commit.

dcvz: I was going more for option 2 -- should someone be able to extend this functionality? we expect most users to subscribe to state changes events right?

dcvz: regarding the binding, you may have to do `.bind(this, 'param')` or make these `const funName = () => {}`

dcvz:
> It looks like it has to do with [a limitation of the transmuxer used by shaka-player](https://github.com/shaka-project/shaka-player/issues/4392).

Interesting. I guess for a first alpha version this doesn't have to be fixed.. but I'm guessing its gonna be an issue for people with HLS.. we may have to resort to `hls.js` here. However, I think the bigger issue is that it breaks the rest of the queue. That we should probably think about a bit further. Whether the existing API for recovering from errors is enough here.

dcvz:
@jspizziri perhaps a useful reference: uses shaka, dash.js and hls.js -- https://github.com/Eyevinn/html-player

web/TrackPlayer/Player.ts
jspizziri: We can use `.bind` but AFAIK bind does basically the same thing right? From the docs:

> The bind() method of [Function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function) instances creates a new function that, when called, calls this function with its this keyword set to the provided value, and a given sequence of arguments preceding any provided when the new function is called.

If I'm understanding that correctly I feel stylistically lambdas are more syntactically clear, but I might be missing some nuance. I'll update to bind.

dcvz: @jspizziri inso t have strong feelings then. Let’s leave it as is :)

jspizziri: Also, regarding the extension issue. Users shouldn't be able to extend things. But we extend them internally in `TrackPlayerModule` and `PlaylistPlayer`. I'll update to the generic function so you can see the difference.

web/TrackPlayerModule.ts
```diff
@@ -0,0 +1,172 @@
...
+            const progress = await this.getProgress()
+            this.emitter.emit(Event.PlaybackProgressUpdated, {
+              ...progress,
+              track: this.currentIndex,
+            });
+          }
+        },
+        options.progressUpdateEventInterval * 1000,
+      )
+    }
```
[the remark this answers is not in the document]
jspizziri: Updated here https://github.com/doublesymmetry/react-native-track-player/pull/1886/commits/5949f2908ba239d7c1bc24c280c8b0ddfc37c0dc

web/TrackPlayer/Player.ts
jspizziri: As a note, the `onError` one isn't in there since it has a few more parameters that are dependent upon the error received. LMK your thoughts on that, feel like I could've done a better job.

web/TrackPlayer/PlaylistPlayer.ts
```diff
@@ -10,8 +11,15 @@ export class PlaylistPlayer extends Player {
   protected _currentIndex?: number;
   protected repeatMode: RepeatMode = RepeatMode.Off;
 
+  protected async onStateUpdate(state: Exclude<State, State.Error>) {
+    super.onStateUpdate(state);
```
dcvz, line 15: This will still emit a state: ended, right? even though we might not be at the end of the queue.

dcvz: I guess we want to delay this call, and only send it out in `onPlaylistEnded`

dcvz:
> Not sure what to make of this limitation itself or if things need to be reevaluated or if we just need to document it.

Let's just document it for now and mark this as a limitation of the web player.

example/src/index.ts
```diff
@@ -0,0 +1,44 @@
...
+  style.appendChild(document.createTextNode(iconFontStyles));
+}
+
+// Inject stylesheet
+document.head.appendChild(style);
+/******************************************
+ * END: react-native-vector-icons import
+ ******************************************/
+
+const appName = 'Your app name';
```
dcvz, line 36:
```suggestion
const appName = 'RNTP Example';
```

package.json
```diff
@@ -72,6 +73,9 @@
   "bugs": {
     "url": "https://github.com/doublesymmetry/react-native-track-player/issues"
   },
+  "dependencies": {
+    "shaka-player": "^4.3.2"
```
dcvz, line 77: not an expert on deps, but does this meen it gets added a dep for any RNN project as well? shoudl we then perhaps place it in devDeps to mark it as "optional"?

SpadarShut: BTW I noticed some deprecation warnings from shaka-player in the console. Would be great to fix those imo.

jspizziri: @SpadarShut I already fixed the deprecation warning I believe you're referring to [here](https://github.com/doublesymmetry/react-native-track-player/pull/1886/commits/5949f2908ba239d7c1bc24c280c8b0ddfc37c0dc#diff-7380cb609d1f2921eb2360101dbfe66dd422b75ca699917df2d79501dc90d3afR66)

jspizziri: @dcvz correct, it'll get installed everywhere. We can move it to an optional dependency, it just means that getting set up with `web` will require a separate installation step. I'll go ahead and do that.

jspizziri: @dcvz [fixed here](https://github.com/doublesymmetry/react-native-track-player/pull/1886/commits/ba0fb044b7680b4345908964f2463a537d7f793c)

jspizziri: Hmmm... I just realized that removing the package as a dependency causes some typescript errors as we can't access the types at build time. Looking into the best way to solve this.

jspizziri: Doh. scratch that. I'll just add it as a `devDependency` as well as an optional `peerDependency`.

web/TrackPlayer/PlaylistPlayer.ts
jspizziri: @dcvz I believe this is right, isn't it? Here's the sequence:

1. The track will end.
2. An `Event.PlaybackState` with a value of `State.Ended` will be emitted signaling the track has ended.
3. We'll run the `onTrackEdned/onPlaylistEnded` methods which will emit an `Event.PlaybackQueueEnded` event signaling the queue has ended.

So we get an event saying that the track has ended, and another event that says the queue has ended. Are we only supposed to emit one or the other?

jspizziri:
> @jspizziri perhaps a useful reference: uses shaka, dash.js and hls.js -- https://github.com/Eyevinn/html-player

I took a deeper look here and it's pretty interesting. It looks like they're using `shaka-player` for `dash`,  `dash.js` for `mss`, and `hls.js` for `hls`. An interesting combination, and I'm guessing that the reason is that each of the libraries has limitations with a subset of the available streaming formats.

In our case, if we did something like this we'd effectively need to build and teardown the player for each track based on its type. Or instantiate multiple players and switch between them for each track. It's definitely possible, however it's quite a lot more complicated.

It might be more simple to see if we could submit a PR to `mux.js` to enhance its transmuxing capabilities.

Shaka definitely isn't perfect, however in terms of a single solution that supports all the various formats, DRM, etc, it definitely seems to be the best _single_ solution. Open to being wrong about this, let me know what your thoughts are.

jspizziri:
> In my project where I control the files played, I wouldn't want to pull all those extra dependencies when I just need to play mp3s. So I'd suggest to let developers configure support for extra formats as needed.

@SpadarShut , while allowing developers to configure support for extra formats sounds like an appealing option. The implementation of something like that would be quite tricky.

SpadarShut:
> @SpadarShut , while allowing developers to configure support for extra formats sounds like an appealing option, the implementation of something like that would be quite tricky.

Sure, I get it. Just wanted to share my context, that it would be nice to avoid those extra deps if possible.

web/TrackPlayer/PlaylistPlayer.ts
dcvz: This is what iOS does: https://github.com/doublesymmetry/react-native-track-player/blob/ef9fb0061ce5fc6c2892bbc480482a832ac6af92/ios/RNTrackPlayer/RNTrackPlayer.swift#L806

This is what Android does: https://github.com/doublesymmetry/react-native-track-player/blob/ef9fb0061ce5fc6c2892bbc480482a832ac6af92/android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt#L606


If I'm following the logic correctly here, it seems like this should have the same outcome! 👍

dcvz (dismissed):
@jspizziri this looks good to go on my end! Looks like CI typescript check is failing, will be happy to merge once that's green

jspizziri:
@dcvz I fixed the ci [in this commit](https://github.com/doublesymmetry/react-native-track-player/pull/1886/commits/a3dd6de3facea567e37e3017ef76e9b9a26299c1) and I rebased on `main`. Apart from that I didn't touch anything (there was [a merge conflict as a result of this commit](https://github.com/doublesymmetry/react-native-track-player/commit/6e2f30c934fb126dc7b89fd929316b4f2019121c)).

I think we should be good to go once the iOS job finishes taking its sweet time.

dcvz:
Awesome work here @jspizziri!

jspizziri:
@dcvz huzzah! What are your thoughts on a minor release?

dcvz:
@jspizziri i think I'll be making a minor release tomorrow which will include web -- if there are any things you'd like to add to the docs, let's do it. I can only think of a few places where we say we support iOS, Android and Windows.

jspizziri:
@dcvz [docs PR here](https://github.com/doublesymmetry/react-native-track-player/pull/2265)

Issue #1676, opened 2022-08-25
Android Crash: ForegroundServiceDidNotStartInTimeException


CDBridger:
**Describe the Bug**
android.app.ForegroundServiceDidNotStartInTimeException: Context.startForegroundService() did not then call Service.startForeground(): ServiceRecord{334a779 u0 nz.co.nbr/com.doublesymmetry.trackplayer.service.MusicService}

using RNTP 3.0.0

I haven't been able to reproduce this crash in my dev environment, I can just see it happening to users via crashalytics

Looks like its only happening for android 12 / API level 32

**Environment Info:**
System:
    OS: macOS 12.5.1
    CPU: (8) arm64 Apple M1 Pro
    Memory: 85.55 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 18.7.0 - /opt/homebrew/bin/node
    Yarn: 1.22.19 - /opt/homebrew/bin/yarn
    npm: 8.17.0 - /opt/homebrew/bin/npm
    Watchman: 2022.08.15.00 - /opt/homebrew/bin/watchman
  Managers:
    CocoaPods: 1.11.3 - /opt/homebrew/bin/pod
  SDKs:
    iOS SDK:
      Platforms: DriverKit 21.4, iOS 15.5, macOS 12.3, tvOS 15.4, watchOS 8.5
    Android SDK:
      API Levels: 31, 33
      Build Tools: 30.0.3, 31.0.0, 33.0.0
      System Images: android-33 | Google APIs ARM 64 v8a
      Android NDK: Not Found
  IDEs:
    Android Studio: 2021.2 AI-212.5712.43.2112.8815526
    Xcode: 13.4.1/13F100 - /usr/bin/xcodebuild
  Languages:
    Java: 18.0.2 - /usr/bin/javac
  npmPackages:
    @react-native-community/cli: Not Found
    react: 18.0.0 => 18.0.0 
    react-native: ^0.69.4 => 0.69.4 
    react-native-macos: Not Found
  npmGlobalPackages:
    *react-native*: Not Found

**How I can Help**

it looks like the google docs says startForeground needs to be called within 5 seconds of context.startService
but I can't see where NotificationState.POSTED gets called to trigger the event which calls startForeground.

my three current theories as to why this crash is occurring is
[...]

Issue #1874, opened 2023-01-12
[Android] Error: Context.startForegroundService() did not then call Service.startForeground()


ernestbies:
**Describe the Bug**
This problem has been present in the library for a very long time but is still not fixed and all topics related to it are closed. So I decided to create a new one.

Topics related to this issue: [#473](https://github.com/DoubleSymmetry/react-native-track-player/issues/473), [#524](https://github.com/doublesymmetry/react-native-track-player/issues/524), [#1676](https://github.com/doublesymmetry/react-native-track-player/issues/1676).

The problem is serious, has been present for several years and affects the user experience of users using the **react-native-track-player** library.

**Steps To Reproduce**
Lock the screen when building the app. In addition, the problem occurs after listening to a given track for a long period of time.
The problem also occurs when users use the application (data from Crashlytics, reported by users).

1. Run the application via **react-native run-android** and make a screen lock.
2. An error will occur.

```
Fatal Exception: android.app.RemoteServiceException$ForegroundServiceDidNotStartInTimeException: Context.startForegroundService() did not then call Service.startForeground(): ServiceRecord{1fac5c6 u0 com.exampleapp.app/com.doublesymmetry.trackplayer.service.MusicService}
       at android.app.ActivityThread.generateForegroundServiceDidNotStartInTimeException(ActivityThread.java:2242)
       at android.app.ActivityThread.throwRemoteServiceException(ActivityThread.java:2213)
       at android.app.ActivityThread.-$$Nest$mthrowRemoteServiceException()
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2505)
       at android.os.Handler.dispatchMessage(Handler.java:106)
       at android.os.Looper.loopOnce(Looper.java:226)
       at android.os.Looper.loop(Looper.java:313)
       at android.app.ActivityThread.main(ActivityThread.java:8741)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:571)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1067)```
```

Another way to trigger the problem is to replace **startService()** with **startForegroundService()** in the file **MusicModule.kt**. Wait about 10 seconds and after this change application will crash with the same error.

**Replicable on Example App?**
Yes.

**Environment Info:**
The problem occurs on physical devices and emulators. Versions:
"react-native": "^0.70.6"
[...]

Issue #1812, opened 2022-11-07
[Android] Crash Fatal Exception: android.app.BackgroundServiceStartNotAllowedException: Not allowed to start service Intent 


RamProg:
**Describe the Bug**
I have reports of consistent crashes on production. I haven't been able to reproduce it.

**Steps To Reproduce**
```
Fatal Exception: android.app.BackgroundServiceStartNotAllowedException: Not allowed to start service Intent { cmp=com.charmander.pikachu/com.doublesymmetry.trackplayer.service.MusicService }: app is in background uid UidRecord{46d20f0 u0a331 TPSL idle procs:0 seq(0,0,0)}
       at android.app.ContextImpl.startServiceCommon(ContextImpl.java:1941)
       at android.app.ContextImpl.startService(ContextImpl.java:1887)
       at android.content.ContextWrapper.startService(ContextWrapper.java:793)
       at android.content.ContextWrapper.startService(ContextWrapper.java:793)
       at com.doublesymmetry.trackplayer.module.MusicModule.setupPlayer(MusicModule.java:184)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.facebook.react.bridge.JavaMethodWrapper.invoke(JavaMethodWrapper.java:372)
       at com.facebook.react.bridge.JavaModuleWrapper.invoke(JavaModuleWrapper.java:188)
       at com.facebook.react.bridge.queue.NativeRunnable.run(NativeRunnable.java)
       at android.os.Handler.handleCallback(Handler.java:938)
       at android.os.Handler.dispatchMessage(Handler.java:99)
       at com.facebook.react.bridge.queue.MessageQueueThreadHandler.dispatchMessage(MessageQueueThreadHandler.java:27)
       at android.os.Looper.loopOnce(Looper.java:226)
       at android.os.Looper.loop(Looper.java:313)
       at com.facebook.react.bridge.queue.MessageQueueThreadImpl$4.run(MessageQueueThreadImpl.java:226)
       at java.lang.Thread.run(Thread.java:920)
```

**Code To Reproduce**
It started happening when updating to 3.0.0.
In 100% of the cases this happens the app is in the background.

**Replicable on Example App?**
Not able to do this at the moment.

**Environment Info:**
Not happening on local, only production in several different devices, specially HUAWEI.
`react-native-track-player 3.2.0-d23c3fb990aaf9d6783881fabe21e059b62b0782`

**How I can Help**
Here to help however I can.

[...]

Issue #1666, opened 2022-08-23
Service.startForeground() not allowed due to mAllowStartForeground false


RamProg:
**Describe the Bug**
Production crashes. 99% on Android 12. 100% with the app in the background.

```
Fatal Exception: java.lang.RuntimeException: Unable to create service com.guichaguri.trackplayer.service.MusicService: android.app.ForegroundServiceStartNotAllowedException: Service.startForeground() not allowed due to mAllowStartForeground false: service com.economist.darwin/com.guichaguri.trackplayer.service.MusicService
       at android.app.ActivityThread.handleCreateService(ActivityThread.java:4567)
       at android.app.ActivityThread.access$1700(ActivityThread.java:256)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2110)
       at android.os.Handler.dispatchMessage(Handler.java:106)
       at android.os.Looper.loopOnce(Looper.java:201)
       at android.os.Looper.loop(Looper.java:288)
       at android.app.ActivityThread.main(ActivityThread.java:7870)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:548)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1003)
```
```
Caused by android.app.ForegroundServiceStartNotAllowedException: Service.startForeground() not allowed due to mAllowStartForeground false: service com.economist.darwin/com.guichaguri.trackplayer.service.MusicService
       at android.app.ForegroundServiceStartNotAllowedException$1.createFromParcel(ForegroundServiceStartNotAllowedException.java:54)
       at android.app.ForegroundServiceStartNotAllowedException$1.createFromParcel(ForegroundServiceStartNotAllowedException.java:50)
       at android.os.Parcel.readParcelable(Parcel.java:3334)
       at android.os.Parcel.createExceptionOrNull(Parcel.java:2421)
       at android.os.Parcel.createException(Parcel.java:2410)
       at android.os.Parcel.readException(Parcel.java:2393)
       at android.os.Parcel.readException(Parcel.java:2335)
       at android.app.IActivityManager$Stub$Proxy.setServiceForeground(IActivityManager.java:7158)
       at android.app.Service.startForeground(Service.java:736)
       at com.guichaguri.trackplayer.service.MusicService.onCreate(MusicService.java:125)
[...]

PR #1888, opened 2023-01-25
V3.2

martinmidtsund:
The first commit removes the static MediaButtonReceiver which is causing a lot of RemoteServiceExceptions, and a group of other related exceptions, most of them mentioned in [this issue](https://github.com/doublesymmetry/react-native-track-player/issues/1676). This change made all of the related errors go away in our production environment, and has been running there for almost a month now. @elliotdickison has also reported the same.

As seen [here](https://developer.android.com/reference/androidx/media/session/MediaButtonReceiver) a service can also receive media button actions: `A service can receive a key event by including an intent filter that handles [ACTION_MEDIA_BUTTON](https://developer.android.com/reference/android/content/Intent.html#ACTION_MEDIA_BUTTON())` which is already registered in the manifest, and it seems to work just fine. We have verified with a connected bluetooth headset that skip, play and pause is still working when originating from the headset. It would be great if you could test the PR and verify it with other devices as well.

The last commit exporting the MusicService and setting foregroundServiceType is not tested in production yet, but should be set as I gather from [this page](https://developer.android.com/guide/topics/manifest/service-element)

Sorry for the wall of text. Please advise if anything should be changed.

### DIFF
--- a/android/src/main/AndroidManifest.xml
+++ b/android/src/main/AndroidManifest.xml
@@ -13,14 +13,6 @@
                 <action android:name="android.intent.action.MEDIA_BUTTON" />
             </intent-filter>
         </service>
-
-        <!-- A receiver that will receive media buttons. Required on pre-lollipop devices -->
-        <receiver android:name="androidx.media.session.MediaButtonReceiver" android:exported="true">
-            <intent-filter>
-                <action android:name="android.intent.action.MEDIA_BUTTON" />
-            </intent-filter>
-        </receiver>
-
     </application>
 
 </manifest>
--- a/android/src/main/AndroidManifest.xml
+++ b/android/src/main/AndroidManifest.xml
@@ -8,7 +8,7 @@
     <application>
 
         <!-- The main service, handles playback, playlists and media buttons -->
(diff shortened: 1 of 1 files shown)


RamProg (approved):

mmmoussa (commented):
We have been using this change, but with `android:exported` left unchanged (ie. `false`), in production with a large number of users. I can confirm that it has significantly reduced our crash rate by resolving #1676 and #1874. However, we are still encountering #1666 and #1812.

martinmidtsund:
@mmmoussa Great to hear! My guess, without having been able to test it yet, is that the last commit hopefully alleviates some of the notAllowed errors.

mmmoussa:
We have been running with all of the changes in this PR (including `exported` as `true`) and I can confirm that #1666 and #1812 are not fixed, though this is still worth merging as it fixes other issues as mentioned previously.

bradleyflood:
Yep same here. Deployed this change to production, still getting crash reports on new versions for #1666 and #1812 :(

martinmidtsund:
@mmmoussa @bradleyflood Thanks for testing!

Perhaps @jspizziri or @dcvz could take a look? I really think this should go into the main-branch as well as this 3.2 branch, but it's a good place to start. :)

PR #1907, opened 2023-02-06
feat(ios): Set user-agent on iOS 16+

kaioduarte:
Following up on https://github.com/doublesymmetry/react-native-track-player/pull/1816 as the author is busy and can't update the PR.


ios/RNTrackPlayer/Models/Track.swift
```diff
@@ -35,6 +36,8 @@ class Track: AudioItem, TimePitching, AssetOptionsProviding {
         guard let url = MediaURL(object: dictionary["url"]) else { return nil }
         self.url = url
         self.headers = dictionary["headers"] as? [String: Any]
+        self.userAgent = dictionary["userAgent"] as? String
+        self.artworkURL = MediaURL(object: dictionary["artwork"])
```
puckey, line 40: Why is this added? Seems unrelated to the user-agent..

ios/RNTrackPlayer/Models/Track.swift
```diff
@@ -124,11 +127,18 @@ class Track: AudioItem, TimePitching, AssetOptionsProviding {
...
 
     func getAssetOptions() -> [String: Any] {
+        var options: [String: Any] = [:]
         if let headers = headers {
-            return ["AVURLAssetHTTPHeaderFieldsKey": headers]
+            options["AVURLAssetHTTPHeaderFieldsKey"] = headers
         }
-
-        return [:]
+        if #available(iOS 16, *) {
```
puckey, line 134: Is there any way to also add the user agent when the ios version is lower than 16?

puckey (changes requested):
Thanks for your pull request – added a few comments / questions

example/Gemfile.lock
```diff
@@ -92,9 +92,10 @@ PLATFORMS
...
   cocoapods (~> 1.11, >= 1.11.2)
+  rexml
 
 RUBY VERSION
-   ruby 2.7.4p191
+   ruby 3.0.5p211
 
 BUNDLED WITH
-   2.2.27
+   2.2.33
```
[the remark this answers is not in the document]
kaioduarte: Actually, I had to make those changes to run the same steps the CI is running locally.

I got this error while running `bundle install` with 2.6:
```sh
Your Ruby version is 2.6.10, but your Gemfile specified 3.0.5
```

Should I revert those changes and see if the CI works?

ios/RNTrackPlayer/Models/Track.swift
kaioduarte: Oops, I missed this one when rebasing with main 🙃

example/Gemfile.lock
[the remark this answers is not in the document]
kaioduarte: Sounds good

PR #1908, opened 2023-02-06
Avoid exiting process when notification closes

puckey:
On android calling `TrackPlayer.reset()` causes the app to exit. I tracked it down to  `exitProcess(0)` being called when the notification closes: https://github.com/doublesymmetry/react-native-track-player/blob/64d2fb1858a56290d0b4f2339dc6b3e8415b0e5e/android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt#L488

This was introduces in the following commit https://github.com/puckey/react-native-track-player/commit/e742959cc1d697d69daeade39830155bce2ccde7 and while I have asked the author @mpivchev for some clarification for the reason for this being added, I have yet to hear back.

For this reason, this pull request will need a bunch of testing as I am unsure on the side effects it might produce. Some cursory testing seems to point to things working though..

### DIFF
--- a/android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
+++ b/android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
@@ -485,7 +485,6 @@ class MusicService : HeadlessJsTaskService() {
                         }
 
                         stopSelf()
-                        exitProcess(0)
                     }
                 }
             }


puckey:
I encorporated one of the commits from the following pull request: https://github.com/doublesymmetry/react-native-track-player/pull/1879/commits/4dd9d2f8724dbaba29a1daa9d9847de5567ee472 – it seems to fully fix the issues that were being seen as per this comment from @HyopeR: https://github.com/doublesymmetry/react-native-track-player/pull/1879/commits/6273fa66a7f205a11216d050f926f37b7f220309#r1085368928

Thanks @HyopeR for the solution!

Issue #1653, opened 2022-08-17
clearNowPlayingMetadata clears metadata, but never comes back


DennisdeWitNL:
**Describe the Bug**
After clearing the now playing metadata by clearNowPlayingMetadata, every new track added, will not show the metadata anymore, unless you fully restart the app after forcing the app to stop. This is an Android issue.

**Steps To Reproduce**
1) Download my example here: https://github.com/DennisdeWitNL/example-clearmetadata
2) Play the song. Check the metadata in the notification control center. It is there.
3) Now pause the song and check the metadata. It is still there.
4) Now click Clear Metadata. The metadata has been cleared.
5) Now click Queue and then click Play. There is still no metadata, unless you fully restart the app and stop the service.

**Code To Reproduce**
Please download my edited example here: https://github.com/DennisdeWitNL/example-clearmetadata

**Environment Info:**
Paste the results of `npx react-native info`

```
info Fetching system and libraries information...
System:
    OS: macOS 13.0
    CPU: (10) arm64 Apple M1 Max
    Memory: 3.93 GB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 18.7.0 - /opt/homebrew/bin/node
    Yarn: 1.22.19 - ~/Documents/Apps/node_modules/.bin/yarn
    npm: 8.15.0 - /opt/homebrew/bin/npm
    Watchman: Not Found
  Managers:
    CocoaPods: 1.11.3 - /opt/homebrew/bin/pod
  SDKs:
    iOS SDK:
      Platforms: DriverKit 22.0, iOS 16.0, macOS 13.0, tvOS 16.0, watchOS 9.0
    Android SDK: Not Found
  IDEs:
    Android Studio: 2021.2 AI-212.5712.43.2112.8609683
    Xcode: 14.0/14A5270f - /usr/bin/xcodebuild
  Languages:
    Java: 17.0.3.1 - /usr/bin/javac
  npmPackages:
    @react-native-community/cli: Not Found
    react: 17.0.2 => 17.0.2 
    react-native: 0.68.1 => 0.68.1 
    react-native-macos: Not Found
  npmGlobalPackages:
    *react-native*: Not Found


```
Paste the exact `react-native-track-player` version you are using

``` react-native-track-player@3.0.0 ```
Real device? Or simulator?
Both

What OS are you running?
[...]

Issue #1834, opened 2022-11-28
[Feature Request]  Close track player and remove from notifications


arthurgeron-work:
**What is the need and use case of this feature?**
When a user is playing a content through other means (e.g. a ultra low latency stream using aws's react-native player) it's misleading to show media controls in the notification center.

**Describe the ideal solution**
Having a clear method to close and remove track player from notification center, which would then open again if "setup" is called. Better still would be if I could pass an argument during it's initial setup so that it'd automatically close if nothing is being streamed/played.

jspizziri:
@arthurgeron-work  please see this option https://react-native-track-player.js.org/docs/api/objects/android-options

arthurgeron-work:
> @arthurgeron-work please see this option https://react-native-track-player.js.org/docs/api/objects/android-options

That API refers to a Android specific event based behaviour which depends on the app being killed, not a on-demand solution as I described.

jspizziri:
In that case please see this feature request https://github.com/doublesymmetry/react-native-track-player/issues/1786

But I don't think I know what you're trying to achieve. Are you trying to use this library as a way to headlessly control expoplayer being initialized in another app? That's not really what this project is designed to do. I'm not sure what the reference to AWS's native player means. I think I'm missing a link here.

arthurgeron-work:
> In that case please see this feature request #1786 its also possible `reset` is something you're looking for.
> 
> But I don't think I know what you're trying to achieve. Are you trying to use this library as a way to headlessly control expoplayer being initialized in another app? That's not really what this project is designed to do. I'm not sure what the reference to AWS's native player means. I think I'm missing a link here.

Yes, headless play of content was the intended use, the confusion here is that the lib we previously used, react-native-music-control, had that feature but was incompatible with android sdk 30+ (forcing us to migrate to another solution), that lib's readme tells devs to "use this lib instead" and neither lib's readme files mentions the change in how headless play is supported. Maybe it'd be interesting to add that to the read me file.

Regarding the play/stop and forgetting about mixing other concepts (like ivs), the issue is that customers find deeply annoying to have a empty lockscreen player that can't be closed, specially for background apps that might stay running, which ultimately leads to the user being induced to always kill the app after watching a stream.  Summing up, it'd be really nice if I could programatically remove the player from the notification center, or have it do that automatically when the track queue is empty.

[...]

PR #1909, opened 2023-02-06
Deprecate clearNowPlayingMetadata

puckey:
See https://github.com/doublesymmetry/react-native-track-player/issues/1653

### DIFF
--- a/src/trackPlayer.ts
+++ b/src/trackPlayer.ts
@@ -259,6 +259,11 @@ export async function updateMetadataForTrack(
   return TrackPlayer.updateMetadataForTrack(trackIndex, metadata);
 }
 
+/**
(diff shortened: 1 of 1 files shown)


jspizziri (approved):
LGTM.

@gavrichards pinging you here as I believe you may use these methods.

gavrichards:
> @gavrichards pinging you here as I believe you may use these methods.

Just checked, we don't use this method. Thanks for the ping!

watadarkstar:
@puckey We are on version 3.2 and use `clearNowPlayingMetadata` method to remove the track player from the notifications tray. See here for more context:
https://github.com/doublesymmetry/react-native-track-player/issues/1834#issuecomment-1472722508

Are you sure we want to deprecate this method? If so, how will we clear the track player from the notification tray?

PR #1914, opened 2023-02-08
add missing ios getPlayWhenReady & setPlayWhenReady methods, simplify example

puckey:


puckey:
The docs for these were already in – just the ios implementation was missing

PR #1920, opened 2023-02-10
Fix updateOptions

puckey:
__iOS__

not done yet

__Android__

(yet to be tested)

MusicService#updateOptions:
- merge options in order to avoid reverting previous options
- check whether updates are necessary before updating notification / progress update job etc
- extract updateProgressUpdate and updateNotification methods because they are quite complex
- inline getIcon helper and merge with getIconOrNull using default value for defaultIcon parameter

chores:
- rename `MusicService#getPlayerStateBundle` to `MusicService#getPlaybackStateBundle` and  remove unnecessary parameter
- extract const vals into enums grouped by topic
- inline const vals that are only used in one place, since there were already a bunch of inlined string keys for bundles.. and the code becomes more legible this way
- use Bundle().apply where possible
- inline MusicService.isCompact
- move DEFAULT_JUMP_INTERVAL value closer to where it is actually being used


gavrichards:
Could this be reopened? It addresses a legitimate issue and was a work in progress.

PR #1928, opened 2023-02-15
fix(ios): emit state passed to handleAudioPlayerStateChange

puckey:
I noticed double `State.Playing` events and tracked it down to this issue in ios.

### DIFF
--- a/ios/RNTrackPlayer/RNTrackPlayer.swift
+++ b/ios/RNTrackPlayer/RNTrackPlayer.swift
@@ -703,8 +703,7 @@ public class RNTrackPlayer: RCTEventEmitter, AudioSessionControllerDelegate {
     @objc(getPlaybackState:rejecter:)
     public func getPlaybackState(resolve: RCTPromiseResolveBlock, reject: RCTPromiseRejectBlock) {
         if (rejectWhenNotInitialized(reject: reject)) { return }
-
-        resolve(getPlaybackStateBodyKeyValues())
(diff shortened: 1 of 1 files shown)


bradleyflood (commented):
🤘

ios/RNTrackPlayer/RNTrackPlayer.swift
```diff
@@ -768,9 +767,9 @@ public class RNTrackPlayer: RCTEventEmitter, AudioSessionControllerDelegate {
         }
     }
 
-    private func getPlaybackStateBodyKeyValues() -> Dictionary<String, Any> {
-        var body: Dictionary<String, Any> = ["state": State.fromPlayerState(state: player.playerState).rawValue]
-        if (player.playerState == AudioPlayerState.failed) {
+    private func getPlaybackStateBodyKeyValues(state: AudioPlayerState) -> Dictionary<String, Any> {
+        var body: Dictionary<String, Any> = ["state": State.fromPlayerState(state: state).rawValue]
+        if (state == AudioPlayerState.failed) {
             body["error"] = getPlaybackStateErrorKeyValues()
```
martinmidtsund, line 773: Just curious and want to learn more about the codebase; shouldn't the state be passed to the `getPlaybackStateErrorKeyValues` as well, if changing this? Now The `state === AudioPlayerState.failed` check will not be the same state as the error-values function are checking on in line 741?

Issue #1931, opened 2023-02-18
usePlaybackState only updates when playback state is changed manually


tylermwatts:
**Describe the Bug**
The `usePlaybackState` hook will return `{ state: undefined }` until playback state is changed manually.

**Steps To Reproduce**
1. Start playing a track
2. Render a different component that consumes `usePlaybackState`
3. Notice that playback state will be `{ state: undefined }` until playback state is changed manually.

**How I can Help**
The culprit may be this line using the nullish coalescing operator
https://github.com/doublesymmetry/react-native-track-player/blob/979f02c85469cfc742926cba0aa78988bcc84930/src/hooks/usePlaybackState.ts#L24

Since `state` in that instance will always be an object with a property of `state: undefined` and therefore never be `null` or `undefined`. It may be as simple as updating the code to
```js
setState((state) => state.state ? state : initialState);
```

PR #1932, opened 2023-02-18
fix(useplaybackstate): updates setting initial playback state in usePlaybackState hook

tylermwatts:
Fixes an issue with the initial state being set in the `usePlaybackState` hook.

Currently, the `??` nullish coalescing operator always returns the current state since it evaluates the entire state object `{ state: undefined }` rather than the internal value. This change evaluates the inner `state` value of the current state (`currentState.state`) to determine whether it should use the fetched state from calling `getPlaybackState()`. 

re #1931


puckey:
LGTM - thanks Tyler

puckey (approved):

PR #1939, opened 2023-02-22
chore(ios): allow higher time precision for progressUpdateEventInterval

### LINKED ISSUE #1916: [Feature Request] Accept float values for progressUpdateEventInterval
It would be nice to be able to have progress updates faster than one time per second. Obviously depending on the load they may not be emitted as accurately as wanted, but I would like to have them twice per second, which should be fine.

puckey:
Fixes #1916


jspizziri (approved):
LGTM

Issue #1823, opened 2022-11-14
3.2.0 breaks detox tests


Johan-dutoit:
**Describe the Bug**
Pre-3.2.0 (aka <=3.1.0) detox e2e tests seem to work just fine, however after upgrading to 3.2.0 detox started failing about (JS timers synchronisation issues).  

I had a look at the change log and ruled out SwiftAudioEx, updating that solely does not cause the issue.  Which leaves a whole host of changes that could be causing this.

FWIW I've only tested this on iOS, not Android.

The app still starts, etc.  Just no luck on progressing due to the timers "blocking" the thread 

**Steps To Reproduce**
- New Project,
- Install RNTP@3.2.0,
- Install Detox
- Added the example test
- Run tests
- Experience bug

**Environment Info:**
System:
    OS: macOS 12.6.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 1.62 GB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.18.0 - /usr/local/bin/node
    Yarn: 1.22.18 - /opt/homebrew/bin/yarn
    npm: 8.19.2 - /opt/homebrew/bin/npm
    Watchman: 2022.10.10.00 - /opt/homebrew/bin/watchman
  Managers:
    CocoaPods: 1.11.3 - /opt/homebrew/bin/pod
  SDKs:
    iOS SDK:
      Platforms: DriverKit 22.1, iOS 16.1, macOS 13.0, tvOS 16.1, watchOS 9.1
    Android SDK: Not Found
  IDEs:
    Android Studio: 2021.2 AI-212.5712.43.2112.8815526
    Xcode: 14.1/14B47b - /usr/bin/xcodebuild
  Languages:
    Java: 11.0.11 - /usr/bin/javac
  npmPackages:
    @react-native-community/cli: 6.3.1 => 6.3.1 
    react: 17.0.2 => 17.0.2 
    react-native: 0.68.2 => 0.68.2 
    react-native-macos: Not Found
  npmGlobalPackages:
    *react-native*: Not Found

Simulator & macOS

**How I can Help**
I can debug and provide any info needed.

Johan-dutoit:
The actual detox issue

```
 There are 62 work items pending on the dispatch queue: "Main Queue (<OS_dispatch_queue_main: com.apple.main-thread>)".
• 2 enqueued JavaScript timers:
  - Timer #1:
[...]

PR #1944, opened 2023-02-28
feat(hooks): add option to use setInterval to useProgress in order to fix tests

sam1463:
add option to use setInterval instead of setTimeout in useProgress

This proof of concept addresses issues with detox tests if you pass `true` for optional `useSetInterval` arg

#1823


puckey:
How about we add an option to use the [progress update event](https://react-native-track-player.js.org/docs/next/api/events#playbackprogressupdated) instead?

Screenshot 2023-03-25 at 16 50 25

sam1463:
> How about we add an option to use the [progress update event](https://react-native-track-player.js.org/docs/next/api/events#playbackprogressupdated) instead?
> 
> Screenshot 2023-03-25 at 16 51 16

Thanks for the input! This is my first time making any changes to this repo, so I'm not that familiar with it. Could you elaborate a bit on what you were thinking it would look like to use the `PlaybackProgressUpdate` event? Would it still call `setTimeout` rather than `setInterval` under the hood?

The reason I changed it back to `setInterval` is due to the nature of how detox waits on js timers, where it will wait on a setTimeout but not a setInterval, and so with endless `setTimeout` calls it will never proceed to run tests.

bradleyflood:
Why not just use `setInterval` for the whole hook instead? Get rid of setTimeout.. should work right?

sam1463:
> Why not just use `setInterval` for the whole hook instead? Get rid of setTimeout.. should work right?

Yeah I think so, but that's how it used to be and it was changed to be `setTimeout` [here](https://github.com/doublesymmetry/react-native-track-player/pull/1686) and I was hesitant to revert that, since it seems like an intentional bugfix. Doing it this way also enables people to use setTimeout in detox builds and keep the existing functionality in prod builds (by passing in `useSetInterval = true` in detox builds but `false` in prod builds)

sam1463:
Closing this since the same result can be achieved by just increasing the interval to >1500ms. Per the Detox docs, it only waits 1500ms for setTimeout. So no change to this library is needed. You just need to pass different progressUpdateInterval values in detox tests than in your app, so that it's >1500 in detox tests

PR #1959, opened 2023-03-15
docs(example): update the documentation on building and installing a …

jspizziri:
…local KotlinAudio package


jspizziri:
1. `mavenLocal()` is now added directly in the `example/build.gradle`
2. We improved the tooling so you don't have to deal with `yarn add file:..` anymore.

PR #1964, opened 2023-03-18
Add iosInitialTime to enable beginning track playback from a specified time

mitchdowney:
In Podverse we use this iosInitialTime patch provided by @gilbertl and add it every time using patch-package. Any reason not to merge this into RNTP?

https://github.com/doublesymmetry/react-native-track-player/issues/387#issuecomment-709544925

### DIFF
--- a/ios/RNTrackPlayer/Models/Track.swift
+++ b/ios/RNTrackPlayer/Models/Track.swift
@@ -11,7 +11,7 @@ import MediaPlayer
 import AVFoundation
 import SwiftAudioEx
 
-class Track: AudioItem, TimePitching, AssetOptionsProviding {
+class Track: AudioItem, TimePitching, AssetOptionsProviding, InitialTiming {
     let url: MediaURL
 
     @objc var title: String?
@@ -26,9 +26,16 @@ class Track: AudioItem, TimePitching, AssetOptionsProviding {
     var userAgent: String?
     let pitchAlgorithm: String?
     var isLiveStream: Bool?
+    
+    var initialTime: TimeInterval = 0.0
 
     var album: String?
     var artwork: MPMediaItemArtwork?
+
+    let initialTime = dictionary["iosInitialTime"] as? Double
+    if let x = initialTime {
+        self.initialTime = x
+    }
 
     private var originalObject: [String: Any] = [:]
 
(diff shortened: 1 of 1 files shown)


jspizziri:
@mitchdowney please see the conversations associated with this PR: https://github.com/doublesymmetry/react-native-track-player/pull/1413 and the associated comment thread here https://github.com/DoubleSymmetry/react-native-track-player/pull/713#issuecomment-1036286082

In short I'll summarize why we wouldn't want to merge this:

1. There are some conceptual problems with an `initialTime` like this PR implements as outlined by David here https://github.com/DoubleSymmetry/react-native-track-player/pull/713#issuecomment-1025872309
2. If we were going to do it like this, the proper implementation IMO would be platform agnostic and not specific to something like iOS or Android.

mitchdowney:
@jspizziri thanks for the info.

> EDIT: FWIW, I'm using the skip(initialTime) method in my app and it works well.

I tried doing that for a long time, but kept running into a problem where a track would start playing from the beginning of the track before the `skip` got handled and the player jumped to the correct playback position. I thought maybe that issue is unavoidable when relying on track events to detect track change.

I can imagine putting the "start from X time" handling inside the track does not work for all project's use cases, but in the case of our podcast app, it seems to work well.

Without this PR merged in, it does cause problems for us, because it means we have to recreate our track-player patch every time we upgrade RNTP versions. Anyway, if it doesn't fit though no problem. We'll either need to figure out how to use `skip` with queue events seamlessly, or fork the repo.

jspizziri:
@mitchdowney The old way (which was to call skip and then immediately call seek) definitely behaved as you described. The modified skip that natively seeks to an initial time shouldn't be behaving as you describe, and I've never seen it do that in my app. 

Can you create a repro for me so I can look at it?

mitchdowney:
@jspizziri ah ok, sorry I forgot about calling `skip` with the `initialPosition` param.

I can work on a repro, but I think there's just one key thing I'm not understanding...if there are 2 items in a queue, and the second item is supposed to begin playing at its 30 minute mark, and the 1st item is currently playing then reaches its end, and the queue automatically proceeds to play the 2nd item...are we supposed to call `skip` with the `initialPosition` somewhere in that process?

jspizziri:
@mitchdowney that's the only way to do it yes. In my use-case I don't actually use the Queue/Playlist, I'm adding a single track at a time. I don't disagree that it's not ideal, and if you can make a compelling enough case to change `dcvz`'s mind on the subject then it's possible to get implemented. Regardless, I don't think the approach implemented here would be the way we'd want to go.

mitchdowney:
@jspizziri ok interesting. I have seen other people mention using a "one track at a time" approach, then handling the next item in the queue within the queue ended event. It's a little counter-intuitive because the docs kind of steer us to using the queue differently than that, but that approach can work fine. Maybe I'll propose a note in the documentation. Thanks again for your help.

Issue #1117, opened 2021-02-12
iOS: operations do not guarantee order of execution in async JS functions


vjeranc:
I've noticed that some sequence of calls will not work.
```javascript
await TrackPlayer.reset();
await TrackPlayer.add(track);
```

The code above will sometimes clear the added track. The reset happens after the track is added to the queue.

Also if I track the player position:
```javascript
await TrackPlayer.getPosition();
await TrackPlayer.getCurrentTrack();
await TrackPlayer.getDuration();
```
I'll sometimes get the state that current track switched to the next one but the position and the duration are from the previous track.

From looking at the code it looks like `reject` `resolve` functions in the RN bridge are instantly resolved in some cases. Swift does not seem to use them when async code is run.
Android on the other hand wraps the promise into runnable and I guess resolves it correctly.

Also, if some of these state values will update inbetween `await` calls maybe a unified call for the whole player state is a better interface.

marcinkieruzel:
Sounds like me case as well

vjeranc:
To clarify,
```
await TrackPlayer.reset();
await TrackPlayer.add(track);
```
If player is playing and `reset` is called on track end, then if another track is added the position and duration lags and the player (on native side) thinks that it's the new track that is also ending, so it clears it too.

There seems to be a lot of issues with internal state.

TheFrostage:
According to logs in XCode, the reset function triggers multiple times. Here is the code:

```
await TrackPlayer.reset();
await TrackPlayer.add(queue);
await TrackPlayer.skip(track.id);
```

![Screen Shot 2021-05-05 at 13 14 18](https://user-images.githubusercontent.com/46031765/117126695-cf0de180-ada3-11eb-9300-558a7d46146d.png)

Is there a workaround to reset playback, clear the queue, fill it with new tracks & skip to a specific track?

fivecar:
@dcvz : is this actually fixed? I ask because I continue to get these issues -- particularly what @vjeranc raised, where you get multiple tracks prematurely ended because a track added after reset is sometimes mistakenly PlaybackQueueEnded.

You can also get inconsistencies in these two other conditions:
```typescript
await TrackPlayer.reset();
const queue = await TrackPlayer.getQueue();
// queue.length > 0 (!)
```
and
```typescript
await TrackPlayer.add(track);  // where track.duration > 0
const duration = await TrackPlayer.getDuration();
// duration === 0 (!)
```

[...]

PR #1965, opened 2023-03-19
fix: #1117 - safeguard reset() and add() against races

fivecar:
The bug describes this further, there are race conditions when you call reset() or add() and expect the native side (AVPlayer) to actually have been reset or added to correctly by the time these methods resolve.

The issue comes in SwiftAudioEx's design. You could ask why we're not fixing these problems there. Two things: a) I'm already trying to with https://github.com/doublesymmetry/SwiftAudioEx/pull/43, and b) there are structural design problems in SwiftAudioEx that don't make the ultimate resolution of this issue easy/obvious. It comes down to the way that `add` and `reset` are decoupled from the actual AVPlayer updates through onCurrentItemChanged in QueuedAudioPlayer; because of this decoupling, it doesn't even allow you to hold on to a promise until the AVPlayer has actually been updated prior to resolving it.

So in the meantime, this fallback. We basically poll the native side until it's in the state we expect it to be in. Far from ideal, and I'm super-open to other suggestions. But at the moment, if you don't patch this, the following scenarios won't work correctly:

1. Calling `reset()` followed by `add()`. If the newly added track is shorter than the previously active track in the queue, you'll immediately get an erroneous PlaybackQueueEnded.
2. Calling `reset()` followed by `getQueue()`. This will yield a queue of nonzero length because reset resolves even before the native side of the queue in SwiftAudioEx has fully updated.
2. Calling `add()` (at the top position) followed by `getDuration()`, `seek()`, or any other function that ultimately tunnels through to AVPlayer. These operations will instead act on the previous top track.

This PR fixes all those issues by polling `duration`, which is one of the most direct ways to tunnel all the way through to AVPlayer's perception of the current track. This has two weaknesses:

* Polling. Le sigh.
* It won't be able to help you in cases where you add a track of the exact same duration as the previous track you're replacing.

Note also that this PR does not fix several scenarios:
* `remove()`. This can't be fixed until https://github.com/doublesymmetry/SwiftAudioEx/pull/43 is merged and taken up by this project. In the meantime, if you remove the currently-active item from the queue, operations will still act on the removed item. See the linked PR and bug for more details.
* A bunch of new operations that SwiftAudioEx's current codebase exposes, like `moveItem`, `replaceItem`, and other operations that suffer the exact same problem. When those are taken up into this codebase, someone will need to come back in to patch those as well.

### DIFF
--- a/src/trackPlayer.ts
+++ b/src/trackPlayer.ts
@@ -34,6 +34,12 @@ function resolveImportedPath(path?: number | string) {
   return resolveAssetSource(path) || path;
 }
 
+const POLL_SLEEP_MILLIS = 50;
+const MAX_POLL_ATTEMPTS = 60; // Give up after 3 secs you're still not done
+async function pollSleep() {
+  return new Promise((resolve) => setTimeout(resolve, POLL_SLEEP_MILLIS));
+}
+
 // MARK: - General API
 
 /**
@@ -121,7 +127,51 @@ export async function add(
     tracks[i].artwork = resolveImportedPath(tracks[i].artwork);
   }
 
-  return TrackPlayer.add(tracks, insertBeforeIndex);
+  // If you're inserting beyond the first element, you definitely won't hit
+  // timing issues outlined in #1117.
+  if (insertBeforeIndex > 0) {
+    return TrackPlayer.add(tracks, insertBeforeIndex);
+  }
+
+  // #1117: Now we need to figure out whether you might be replacing an existing
+  // first track -- because if you are, we need to poll/block until we're
+  // sure that the native AVPlayer tied with SwiftAudioEx's AVPlayerWrapper
+  // actually has the new track loaded before we return. Otherwise, future
+  // calls to things like getDuration() will return the old track's
+  // duration.
+  const queue = await TrackPlayer.getQueue();
+
+  // This is another case where you're safe to return immediately: there's
+  // already a top track and you're asking to insert behind it.
+  if (queue.length > 0 && insertBeforeIndex === -1) {
+    return TrackPlayer.add(tracks, insertBeforeIndex);
+  }
+
+  // Now we're in a case where there's either no top track, or there is one
+  // but you've asked to replace it. In both of those cases, we must poll
+  // until we're sure your track has been inserted through to AVPlayer. It
+  // turns out that duration is one of the facets that, when asked for,
+  // reflects what AVPlayer's current track says it is. So we use it as
+  // a proxy for figuring out whether your track is loaded. This of course
+  // won't work when the track you're adding coincidentally has the same
+  // duration as the current top track, but it's the best we currently know
+  // how to do.
+  const curTopTrackDuration = await TrackPlayer.getDuration();
+  const newTopTrackDuration = tracks[0].duration;
+  const index = TrackPlayer.add(tracks, insertBeforeIndex);
+  let attempts = 0;
+
+  if (curTopTrackDuration !== newTopTrackDuration) {
+    while (
+      attempts < MAX_POLL_ATTEMPTS &&
+      (await TrackPlayer.getDuration()) === curTopTrackDuration
+    ) {
+      attempts++;
+      await pollSleep();
+    }
+  }
+
+  return index;
 }
 
 /**
@@ -286,7 +336,19 @@ export function updateNowPlayingMetadata(
  * Resets the player stopping the current track and clearing the queue.
  */
 export async function reset(): Promise<void> {
-  return TrackPlayer.reset();
+  const duration = await TrackPlayer.getDuration();
+
+  await TrackPlayer.reset();
+  if (duration > 0) {
+    let attempts = 0;
+    while (
+      attempts < MAX_POLL_ATTEMPTS &&
+      (await TrackPlayer.getDuration()) > 0
+    ) {
+      attempts++;
+      await pollSleep();
+    }
+  }
 }
 
 /**


fivecar:
... and for posterity, once https://github.com/doublesymmetry/SwiftAudioEx/pull/43 is fixed, here's an implementation of `remove()` that will also patch the issue.

```typescript
async function remove(tracks) {
    if (!Array.isArray(tracks)) {
        tracks = [tracks];
    }

    // If we're going to remove the top track, we need to make sure we block
    // until the changes make it all the way through to AVPlayer on the native
    // side. More details in #1117.
    if (tracks.includes(0)) {
        const curTopTrackDuration = await TrackPlayer.getDuration();
        const res = TrackPlayer.remove(tracks);
        let attempts = 0;
    
        while (
            attempts < MAX_POLL_ATTEMPTS &&
            await TrackPlayer.getDuration() === curTopTrackDuration
        ) {
            attempts++;
            if (attempts % 5 === 0) {
                console.log(`Polling on remove ${curTopTrackDuration} (${attempts})`);
            }
            await pollSleep();
        }
        return res;
    } else {
        return TrackPlayer.remove(tracks);
    }
}
```

jspizziri:
@fivecar thanks so much for the thoughtful PR(s). I need a bit of brain-space in order to think through this. My high-level thought is that I'd rather focus my initial round of thoughts on `SwiftAudioEx` to see if there's a good solution there (as you mentioned, it may not be clear but perhaps with enough eyeballs something will reveal itself).

CC @puckey @dcvz

puckey:
Thanks for bringing this to our attention. I have tried replicating a few of the scenarios you describe, but haven't seen the same behavior..

> Calling reset() followed by add(). If the newly added track is shorter than the previously active track in the queue, you'll immediately get an erroneous PlaybackQueueEnded.

I wasn't able to replicate this. Are you awaiting the add call?

> Calling reset() followed by getQueue(). This will yield a queue of nonzero length because reset resolves even before the native side of the queue in SwiftAudioEx has fully updated.

In this case `getQueue()` gives me `[]`

> Calling add() (at the top position) followed by getDuration(), seek(), or any other function that ultimately tunnels through to AVPlayer. These operations will instead act on the previous top track.

What do you mean with 'at the top position'?

I would very much like to be able to (consistently if possible) reproduce these issues. Any instructions that might help with this are welcome.

puckey:
> Calling add() (at the top position) followed by getDuration(), seek(), or any other function that ultimately tunnels through to AVPlayer. These operations will instead act on the previous top track.

Could it be that you were expecting `add()` to activate the added track? Because you would need to jump to it first..

fivecar:
@puckey : I await all the calls, so it's definitely not an await issue.

> What do you mean with 'at the top position'?

What I mean is, for instance, `reset()` followed by `add()` (meaning you add a track at the to-be-played position). I do understand that you need to jump to the right track in order to activate it; in my case, that's easy to be right about, because I only ever use one track (i.e. my queue is either empty or length 1).

I've been able to reproduce all of these issues, including the final one in the example project. But the first two issues are intermittent, and depend on timing. I've unfortunately been unable to find the means by which to reproduce these consistently in the example project (except for the last one):

```typescript
@@ -18,4 +18,22 @@ export const QueueInitialTracksService = async (): Promise<void> => {
       duration: 28,
     },
   ]);
+  const [dur, prog, queue] = await Promise.all([
+    TrackPlayer.getDuration(),
+    TrackPlayer.getProgress(),
+    TrackPlayer.getQueue(),
+  ]);
+
+  console.log(
+    '* * * QueueInitialTracksService dur',
+    dur,
+    'prog',
+    prog,
+    'queueCount',
+    queue.length
+  );
+
+  await TrackPlayer.remove(0);
+  const newDur = await TrackPlayer.getDuration();
+  console.log('* * * QueueInitialTracksService newDur', newDur);
 };
```
When running this in the example project, you'll see that `newDur` is non-zero due to the bug that this PR fixes. This issue repro's 100%. The other two issues repro often, but not cleanly enough for me to provide code. In the project I'm actually coding on, it happens often enough that I've already applied this patch (via `yarn patch`) to my version of react-native-track-player.

LMK if you have questions! Thanks for taking a look. And to @jspizziri 's point, I hope there's a cleaner solution.

Do you at least see the core problem I'm talking about (i.e. the problem that's clear even from code inspection?).

puckey:
Thanks for getting back to me and providing some example code.

Running it I am getting the following console output:

```
 Event.PlaybackState {"state": "loading"}
 LOG  Event.PlaybackActiveTrackChanged {"index": 0, "lastPosition": 0, "track": {"artist": "David Chavez", "artwork": "https://react-native-track-player.js.org/example/Longing.jpeg", "duration": 143, "title": "Longing", "url": "https://react-native-track-player.js.org/example/Longing.mp3"}}
 LOG  * * * QueueInitialTracksService dur 0 prog {"buffered": 0, "duration": 0, "position": 0} queueCount 7
 LOG  Event.PlaybackState {"state": "paused"}
 LOG  * * * QueueInitialTracksService newDur 0
 LOG  Event.PlaybackState {"state": "loading"}
 LOG  Event.PlaybackActiveTrackChanged {"index": 0, "lastIndex": 0, "lastPosition": 0, "lastTrack": {"artist": "David Chavez", "artwork": "https://react-native-track-player.js.org/example/Longing.jpeg", "duration": 143, "title": "Longing", "url": "https://react-native-track-player.js.org/example/Longing.mp3"}, "track": {"artist": "David Chavez", "artwork": "https://react-native-track-player.js.org/example/Soul%20Searching.jpeg", "duration": 77, "title": "Soul Searching (Demo)", "url": "https://react-native-track-player.js.org/example/Soul%20Searching.mp3"}}
 LOG  Event.PlaybackState {"state": "paused"}
 LOG  Event.PlaybackState {"state": "ready"}
```

I also tried with waiting for a second during the initial load to make sure the initial duration is non-zero:

```ts
export const QueueInitialTracksService = async (): Promise<void> => {
  await TrackPlayer.add([
    ...(playlistData as Track[]),
    {
      url: localTrack,
      title: 'Pure (Demo)',
      artist: 'David Chavez',
      artwork: localArtwork,
      duration: 28,
    },
  ]);
  await new Promise<void>((resolve) => setTimeout(resolve, 1000))
  const [dur, prog, queue] = await Promise.all([
    TrackPlayer.getDuration(),
    TrackPlayer.getProgress(),
    TrackPlayer.getQueue(),
  ]);

  console.log(
    '* * * QueueInitialTracksService dur',
    dur,
    'prog',
    prog,
    'queueCount',
    queue.length
  );

  await TrackPlayer.remove(0);
  const newDur = await TrackPlayer.getDuration();
  console.log('* * * QueueInitialTracksService newDur', newDur);

};
```

which gives me:

```
 LOG  Event.PlaybackState {"state": "loading"}
 LOG  Event.PlaybackActiveTrackChanged {"index": 0, "lastPosition": 0, "track": {"artist": "David Chavez", "artwork": "https://react-native-track-player.js.org/example/Longing.jpeg", "duration": 143, "title": "Longing", "url": "https://react-native-track-player.js.org/example/Longing.mp3"}}
 WARN  Could not find image file:///private/var/containers/Bundle/Application/24243F54-6D6C-4F95-8B25-98C352003593/example.app/undefined.png
 LOG  * * * QueueInitialTracksService dur 0 prog {"buffered": 0, "duration": 0, "position": 0} queueCount 7
 LOG  * * * QueueInitialTracksService newDur 0
 LOG  Event.PlaybackState {"state": "paused"}
 LOG  Event.PlaybackState {"state": "loading"}
 LOG  Event.PlaybackActiveTrackChanged {"index": 0, "lastIndex": 0, "lastPosition": 0, "lastTrack": {"artist": "David Chavez", "artwork": "https://react-native-track-player.js.org/example/Longing.jpeg", "duration": 143, "title": "Longing", "url": "https://react-native-track-player.js.org/example/Longing.mp3"}, "track": {"artist": "David Chavez", "artwork": "https://react-native-track-player.js.org/example/Soul%20Searching.jpeg", "duration": 77, "title": "Soul Searching (Demo)", "url": "https://react-native-track-player.js.org/example/Soul%20Searching.mp3"}}
 LOG  Event.PlaybackState {"state": "paused"}
 LOG  Event.PlaybackState {"state": "ready"}
```

Neither of these are giving me non-zero durations for `newDur`.. I have tried both with a simulator and with an iPhone 14 pro.. What kind of device are you testing on?

(To reiterate, `0` as an initial duration is currently what is expected and also what the android side of the library does..)

puckey:
(edited above comment to correct second console output)

fivecar:
> Is this related? My sentry log show that it happens regularly #1981

@s123121 : I don't see how it's related (in any way) -- could you please elaborate a bit more? #1981 seems to be about a crash due to unsafe access to memory in the native iOS implementation. This PR is about timing issues in how the AVPlayer is managed relative to the RN bridge with the native code. This PR also doesn't address crashes at all -- it merely addresses an inconsistency between RN's understood state of track reset/update vs. what TrackPlayer actually returns (for calls like `getDuration`).

PR #1966, opened 2023-03-19
Allows the granularity to be less than one second. Player updates wou…

vinnybad:
…ld be limited to a minimum value of one second on iOS if this is not done.


vinnybad:
Closed it - saw I was on an older version where the value was 1. it's 1000 on master.

Issue #1963, opened 2023-03-18
[Expo] iOS Notification Controls Missing


jcksncllwy:
**Describe the Bug**
Audio plays in the background just fine, but playback controls do not appear on the lock screen.

**Steps To Reproduce**
I only know how to reproduce this with my setup.

**Code To Reproduce**
I don't know if there is a simple way to reproduce this.

**Replicable on Example App?**
Can you replicate this bug in the [React Native Track Player Example App](https://github.com/doublesymmetry/react-native-track-player/tree/main/example)?
The example app doesn't use Expo, which I think might be part of the problem.

**Environment Info:**
System:
    OS: macOS 13.2.1
    CPU: (8) arm64 Apple M1 Pro
    Memory: 3.31 GB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 19.5.0 - ~/.nvm/versions/node/v19.5.0/bin/node
    Yarn: Not Found
    npm: 9.3.1 - ~/.nvm/versions/node/v19.5.0/bin/npm
    Watchman: Not Found
  Managers:
    CocoaPods: 1.11.3 - /usr/local/bin/pod
  SDKs:
    iOS SDK:
      Platforms: DriverKit 22.2, iOS 16.2, macOS 13.1, tvOS 16.1, watchOS 9.1
    Android SDK: Not Found
  IDEs:
    Android Studio: Not Found
    Xcode: 14.2/14C18 - /usr/bin/xcodebuild
  Languages:
    Java: Not Found
  npmPackages:
    @react-native-community/cli: Not Found
    react: 18.1.0 => 18.1.0 
    react-native: 0.70.5 => 0.70.5 
    react-native-macos: Not Found
  npmGlobalPackages:
    *react-native*: Not Found

Paste the exact `react-native-track-player` version you are using
`react-native-track-player: ^3.2.0`

Real device? Or simulator?
Real device

What OS are you running?
iOS

**How I can Help**
What can you do to help resolve this?
Have you investigated the underlying JS or Swift/Android code causing this bug?
Can you create a Pull Request with a fix?

jspizziri:
@jcksncllwy as I've mentioned [here](https://github.com/doublesymmetry/react-native-track-player/issues/1675#issuecomment-1229154827) and a few other places (linked below for additional context). None of the maintainers use Expo and we don't really officially support it given the lack of expertise in it on the team.

[...]

PR #1969, opened 2023-03-20
docs(expo): Remove misleading documentation about Expo support 

jcksncllwy:
It seems that "none of the maintainers use Expo and we don't really officially support it given the lack of expertise in it on the team." https://github.com/doublesymmetry/react-native-track-player/issues/1963#issuecomment-1475198626

I would have saved a lot of time and effort if the docs had not clearly stated that Expo is supported when it isn't. Hopefully this change will save others from the same wasted effort.

### DIFF
--- a/docs/docs/basics/installation.mdx
+++ b/docs/docs/basics/installation.mdx
@@ -62,4 +62,13 @@ cd ios && pod install
 
 ## Expo
 
-Unfortunately, we do not support Expo at this time.
+You can now use React Native Track Player with Expo. 
+
+Please be aware that while many people are using React Native Track Player with Expo successfully, the current maintainers of this project do not use Expo and their ability to resolve issues involving Expo is limited.
+
+To get started, create a [custom development client](https://docs.expo.dev/clients/getting-started/) for your Expo app and then install React Native Track Player.
+
+Here is the configuration required for audio playback in background:
+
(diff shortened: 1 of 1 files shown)


docs/docs/basics/installation.mdx
```diff
@@ -62,11 +62,4 @@ cd ios && pod install
...
 
-You can now use React Native Track Player with Expo.
-
-Start by creating a [custom development client](https://docs.expo.dev/clients/getting-started/) for your Expo app and then install React Native Track Player.
-
-Here is the configuration required for audio playback in background:
-
-- [iOS: Enable audio playback in background via your app.json](https://docs.expo.dev/versions/latest/sdk/audio/#playing-or-recording-audio-in-background-ios)
-- [Android: Stop playback when the app is closed](./background-mode.md/#android)
+Unfortunately, we do not support Expo at this time.
```
jspizziri, line 65: @jcksncllwy this isn't entirely true, although neither is the earlier message. I know that there are a lot of users who are using Expo successfully, so I don't want to remove this documentation. However, it's probably good to add a note/warning that while it does work you can't expect much support from the maintainers (as they currently exist).

jcksncllwy: Gotcha. I've updated the wording to better reflect that. Let me know what you think

PR #1973, opened 2023-03-22
feat: sleep timer

puckey:
- ios implementation depends on https://github.com/doublesymmetry/SwiftAudioEx/pull/47
- android implementation
- `TrackPlayer.setSleepTimer(seconds)`
- `TrackPlayer.sleepWhenActiveTrackReachesEnd()`
- `TrackPlayer.getSleepTimer()`
- `TrackPlayer.clearSleepTimer()`
- `sleep-timer-changed` event
- `sleep-timer-complete` event
- `useSleepTimer()` hook

### DIFF
--- a/android/src/main/java/com/doublesymmetry/trackplayer/addons/SleepTimer.kt
+++ b/android/src/main/java/com/doublesymmetry/trackplayer/addons/SleepTimer.kt
@@ -1,0 +1,46 @@
+package com.doublesymmetry.trackplayer.addons
+
+import android.os.Handler
+import android.os.Looper
+import kotlinx.coroutines.*
+
+open class SleepTimer {
+    var time: Double? = null
+    private var runnable: Runnable? = null
+    private val handler = Handler(Looper.getMainLooper())
+
+    val isRunning: Boolean
+        get() = time != null
+
+    fun sleepAfter(seconds: Double) {
+        stopTimer()
+        val runnable = Runnable {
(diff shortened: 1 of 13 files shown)


pke:
I am not sure if such should be part of the library. 
Couldn't this easily be solved outside this library with a 3rd party package named like `react-native-track-player-sleep`?

bradleyflood:
It's nice work but my concern is this brings more code to maintain, more Issues that might be raised, more surface area for bugs, and more documentation.

I vote we leave this for individuals to integrate in their own way. Or publish it as a Recipe or Addon.

puckey:
Since there doesn't seem to be much animo for integrating this feature, I will maintain it in our fork instead

martinmidtsund:
😢 We are also in need of this, and thought this was going to be merged so we have awaited the implementation. We do have a pure JS version, but we get reports that sometimes it doesn't stop after the current track.

Is your fork open-source @puckey ?

puckey:
> 😢 We are also in need of this, and thought this was going to be merged so we have awaited the implementation. We do have a pure JS version, but we get reports that sometimes it doesn't stop after the current track.
> 
> Is your fork open-source @puckey ?

yes it is open-source.. I haven't done a production release with it yet though – I will be doing extensive testing in the next few weeks

julianCast:
This was going to be a nice addition!!

PR #1974, opened 2023-03-23
fix(android): avoid service crashes

puckey:
Based on @kyunkakata’s work at https://github.com/doublesymmetry/react-native-track-player/issues/1666#issuecomment-1481534212


android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
```diff
@@ -87,11 +93,45 @@ class MusicService : HeadlessJsTaskService() {
 
     override fun onStartCommand(intent: Intent?, flags: Int, startId: Int): Int {
         startTask(getTaskConfig(intent))
+
+        // Fix crash "Context.startForegroundService() did not then call Service.startForeground()
```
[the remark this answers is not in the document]
puckey: @martinmidtsund Ah interesting, I will remove it unless @kyunkakata has something to add about this issue..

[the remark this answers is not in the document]
kyunkakata: It's great to remove it because #1888 has fixed it. It's just a tricky solution. I am happy with that.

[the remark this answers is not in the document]
martinmidtsund: @kyunkakata Yes, I think so. I haven't seen any reports saying otherwise at least 😄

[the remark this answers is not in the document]
puckey: So first I implemented `TrackPlayer.setupPlayer(options)` to reject if the app was in the background. But then I realized we could also just delay executing the `setupPlayer` call to when the app comes back to the foreground: https://github.com/doublesymmetry/react-native-track-player/pull/1974/commits/ef8f162b98b5a2852a90e02cfa3b34226d6f60cc

Should work, no?

android/src/main/java/com/doublesymmetry/trackplayer/module/MusicModule.kt
```diff
@@ -228,7 +239,16 @@ class MusicModule(reactContext: ReactApplicationContext) : ReactContextBaseJavaM
         manager.registerReceiver(eventHandler!!, IntentFilter(EVENT_INTENT))
 
         Intent(context, MusicService::class.java).also { intent ->
-            context.startService(intent)
+            try {
```
puckey, line 242: @kyunkakata do we need the try / catch here?

puckey:
@kyunkakata did you figure out a way to consistently cause these crashes to happen? I would like to test to make sure everything is fixed now..

kyunkakata:
The crash "Context.startForegroundService() did not then call Service.startForeground() within 5s" is easy to reproduce. Just calling setupPlayer without invoking add function. An ANR or  a crash should happen. 
The second one - "https://github.com/doublesymmetry/react-native-track-player/issues/1666" is diffcult to reproduce. Base on our crash logs, our users that received this type of crash is when the app was in the background and received a push notification, then the crash happens. We tried to reproduce but got no luck. So that we need to check is app in the foreground to start MusicService and is that a MusicService notification or not to ensure that the crash will not happen.

android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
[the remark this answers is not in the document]
puckey: I think I tend to agree with you that this solution is too magical. As always it is a question of where to put the complexity : )

`TrackPlayer.setupPlayer(options)` currently already rejects when the player is already set up. The other option would be to remove that functionality and replace it with rejecting when `TrackPlayer.setupPlayer(options)` fails. Or we check the contents of the error message, which I am a little hesitant about somehow..

puckey:
> The crash "Context.startForegroundService() did not then call Service.startForeground() within 5s" is easy to reproduce. Just calling setupPlayer without invoking add function. An ANR or a crash should happen.

Just to make sure: did the `fake_audio_channel` notification code in your diff fix this issue? Would it stop the anr from happening when you don't call `add()` quickly enough?

puckey:
> > The crash "Context.startForegroundService() did not then call Service.startForeground() within 5s" is easy to reproduce. Just calling setupPlayer without invoking add function. An ANR or a crash should happen.
> 
> Just to make sure: did the `fake_audio_channel` notification code in your diff fix this issue? Would it stop the anr from happening when you don't call `add()` quickly enough?

So I added back the ANR fix, since it is necessary when you don't want to start the exoplayer notification within 5 seconds of starting MusicService. I also reject the `setupPlayer` promise with error code `android_cannot_setup_player_in_background` and have adjusted the example app implementation to retry if it failed in this way.

@kyunkakata any feedback is welcome!

android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
```diff
@@ -87,11 +90,41 @@ class MusicService : HeadlessJsTaskService() {
 
     override fun onStartCommand(intent: Intent?, flags: Int, startId: Int): Int {
         startTask(getTaskConfig(intent))
+        startAndStopEmptyNotificationToAvoidANR()
         return START_STICKY
     }
 
+    /**
+     * Workaround for the "Context.startForegroundService() did not then call Service.startForeground()"
+     * within 5s" ANR and crash by reating an empty notification and stopping it right after. For more
```
jspizziri, line 99: typo `reating` should be `creating`

example/src/App.tsx
```diff
@@ -101,4 +84,28 @@ const styles = StyleSheet.create({
   },
 });
 
+function useSetupPlayer() {
```
jspizziri, line 87: @puckey would it be worth wrapping som version of this logic in our JS lib?

puckey: Originally I implemented retrying on the Android side… but it is probably best left up to the developer on how they want to handle the rejection instead… for example they might decide to want to not retry if a certain amount of time has gone by, or they might want to show a retry button.. or send a notification to the user (although not sure that would be possible from the js side if the app is in the background)

puckey:
Fixed the ts issues, so should be good to go

MortadhaFadhlaoui:
Nice Work! Thank you for the fix, but we still experience the issue, due this line

android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
```diff
@@ -486,12 +486,13 @@ class MusicService : HeadlessJsTaskService() {
             event.notificationStateChange.collect {
                 when (it) {
                     is NotificationState.POSTED -> {
-                        if (AppForegroundTracker.foregrounded && it.notificationId == MUSIC_NOTIFICATION_ID) {
-                            startForeground(it.notificationId, it.notification)
-                        }
+                        startForeground(it.notificationId, it.notification)
```
MortadhaFadhlaoui: This line causing a crash as I mentioned [here](https://github.com/doublesymmetry/react-native-track-player/pull/1974#issuecomment-1487575919), @puckey wdyt about this quick fix 
```
if (Build.VERSION.SDK_INT < Build.VERSION_CODES.S) {
                            startForeground(it.notificationId, it.notification)
    }
```

martinmidtsund:
I'm not entirely sure the service should be foregrounded immediately after the notification is posted. Could this, together with START_STICKY which restarts the service if killed be the reason for the mAllowForeground error? In the Pocket Casts repo they only use setForeground when they start playing sound. https://github.com/Automattic/pocket-casts-android/blob/ee8da0c095560ef64a82d3a31464491b8d713104/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackService.kt#L250

jspizziri:
@martinmidtsund care to submit a PR or test your theory on a fork?

martinmidtsund:
@jspizziri Yes, I'll be testing this and can submit a PR. Can't set a timeline though, don't know when I will get the time.

puckey:
> I'm not entirely sure the service should be foregrounded immediately after the notification is posted. Could this, together with START_STICKY which restarts the service if killed be the reason for the mAllowForeground error? In the Pocket Casts repo they only use setForeground when they start playing sound. https://github.com/Automattic/pocket-casts-android/blob/ee8da0c095560ef64a82d3a31464491b8d713104/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackService.kt#L250

I wonder though – isn't there a big chance the app will be in the background already by the time the sound starts playing? i.e. the users changes the track and moves to another track directly

puckey:
@martinmidtsund shall we regroup on this with a fresh issue containing stack traces as experienced using the latest rc release? Would you mind opening one?

android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
puckey: @MortadhaFadhlaoui Could you open a new pr for this with some explanation on your changes? Thanks!

martinmidtsund:
@puckey Sorry, we are not using v4 yet, so we don't have any stack traces as of now. I could create a new issue with the stack traces we do have, but they are from v3.2.0. Not sure if that helps, but I would guess it's still the same problem in the new RCs?

PR #1976, opened 2023-03-28
deps(android): Move to new lifecycle version

dcvz:


jeroenptrs:
Can confirm that this did the trick for me, I was able to build 4.0.0-rc2 with this change 🙂

PR #1982, opened 2023-04-07
docs: update expo url

Spice-Z:
It seems that expo document changed page's DOM id, so RNTP's link does not work for correct position.

I just fixed page hash.


travisvn:
Nice fix 👍 LGTM

jspizziri (approved):

PR #1990, opened 2023-04-16
docs: rename androidContentType to androidAudioContentType

mitchdowney:
I think this `setupPlayer` options key is incorrectly named in docs.


jspizziri (approved):
Thanks!

PR #2000, opened 2023-04-25
Update getting-started.md to include necessary code for implementing the trackplayer

heaversm:
The way the documentation is set up currently seems to imply you should comment out the `import {AppRegistry} from 'react-native';` line. It also doesn't include the necessary dependencies to implement the shown `TrackPlayer.registerPlaybackService(() => require('./service'));` which are:

```
import TrackPlayer from 'react-native-track-player';
import {PlaybackService} from './src/services';
```

### DIFF
--- a/docs/docs/basics/getting-started.md
+++ b/docs/docs/basics/getting-started.md
@@ -19,6 +19,8 @@ TrackPlayer.registerPlaybackService(() => require('./service'));
 
 ```
 
+Then create the file `service.js`
+
 ```ts
 // service.js
 module.exports = async function() {
@@ -31,7 +33,14 @@ Then, you need to set up the player. This usually takes less than a second:
 ```ts
 import TrackPlayer from 'react-native-track-player';
 
-await TrackPlayer.setupPlayer()
+async function setupPlayer() {
+  await TrackPlayer.setupPlayer();
+  // The player is ready to be used
+}
+
+function App(): JSX.Element {
+  await setupPlayer();
+}
 // The player is ready to be used
 ```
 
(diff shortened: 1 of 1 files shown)


docs/docs/basics/getting-started.md
```diff
@@ -23,7 +33,14 @@ Then, you need to set up the player. This usually takes less than a second:
 ```ts
 import TrackPlayer from 'react-native-track-player';
 
-await TrackPlayer.setupPlayer()
+async function setupPlayer() {
+  await TrackPlayer.setupPlayer();
+  // The player is ready to be used
+}
+
+function App(): JSX.Element {
```
bradleyflood, line 41: I can see less experienced developers getting caught up on `: JSX.Element`. Perhaps they aren't using TS, just plain old JS, and they won't know what to do. I remember this happening to me in my early days.

bradleyflood (approved):
Agreed it does look like to comment out the app.

docs/docs/basics/getting-started.md
```diff
@@ -23,7 +33,14 @@ Then, you need to set up the player. This usually takes less than a second:
 ```ts
 import TrackPlayer from 'react-native-track-player';
 
-await TrackPlayer.setupPlayer()
+async function setupPlayer() {
```
jspizziri, line 36: This additional function wrapping `setupPlayer` doesn't really do anything.

docs/docs/basics/getting-started.md
```diff
@@ -23,7 +33,14 @@ Then, you need to set up the player. This usually takes less than a second:
...
+  await TrackPlayer.setupPlayer();
+  // The player is ready to be used
+}
+
+function App(): JSX.Element {
+  await setupPlayer();
```
jspizziri, line 42: This isn't the proper way to call `setupPlayer` as you'll get an exception if `App` ever re-renders and `setupPlayer` has already been called. We should probably have a minimal example of what's happening in the `example` app and then link to it for more detailed information as there are a lot of ways this can be done.

jspizziri (changes requested):

jspizziri:
I'm closing this PR due to no response to my feedback.

PR #2007, opened 2023-05-03
Add deprecation notice for `PlaybackStateEvent` to V4 migration

bradleyflood:


docs/docs/v4-migration.md
```diff
@@ -46,7 +46,8 @@ instead of `null`
 
 ### Typescript Deep Imports
 
-If you were using deep imports from RNTP, the `src` has been completely
+1. If you were using deep imports from RNTP, the `src` has been completely
 reorganized, and so you may need to adjust your imports accordingly. If you've
 been importing everything directly (ex. `import ... from 'react-native-track-player';`)
 then you don't need to do anything.
+1. The `PlaybackStateEvent` interface has been renamed to `PlaybackState`
```
jspizziri, line 53: Can we mention that this is TS only?

jspizziri (changes requested):

docs/docs/v4-migration.md
bradleyflood: This point and the one above it are under the `### Typescript Deep Imports` header. Which maybe could just be renamed to Typescript Imports. Thoughts?

bradleyflood: Updated heading to `Typescript Imports`

PR #2012, opened 2023-05-11
docs(player): fix references to PlayerOptions and UpdateOptions

curthipster:
1. The documentation for the `updateOptions` function was pointing to the deprecated type `MetadataOptions`. This fixes the reference to point to the `UpdateOptions` type.
2. The documentation for the `setupPlayer` function had the correct name, but the link was pointing to the wrong place. This fixes the link to point to `PlayerOptions`.


jspizziri (approved):
Thanks for the PR!

PR #2023, opened 2023-06-06
fix(android): Allow precise seeking on Android

aurotones:
Converting to Long eliminates the decimal values.

For example, i made an app that replicates the SoundCloud waveform and whenever i seek the track, the waveform would sometimes go backwards from the position because there's no decimal values to seek precisely.

This PR will allow precise seeking.

i checked on iOS and it seem already able to seek precisely out of the box.


puckey (approved):
LGTM

Issue #1986, opened 2023-04-10
[android][rc-03] Crash: Unable to start service com.doublesymmetry.trackplayer.service.MusicService@78647c9 with null


ZeroPie:
**Describe the Bug**`

Firebase Analytics is reporting the following crash
```
Fatal Exception: java.lang.RuntimeException: Unable to start service com.doublesymmetry.trackplayer.service.MusicService@78647c9 with null: android.app.ForegroundServiceStartNotAllowedException: Service.startForeground() not allowed due to mAllowStartForeground false: service com.dierotenbullen.app/com.doublesymmetry.trackplayer.service.MusicService
       at android.app.ActivityThread.handleServiceArgs(ActivityThread.java:5192)
       at android.app.ActivityThread.access$2200(ActivityThread.java:315)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2324)
       at android.os.Handler.dispatchMessage(Handler.java:106)
       at android.os.Looper.loopOnce(Looper.java:226)
       at android.os.Looper.loop(Looper.java:313)
       at android.app.ActivityThread.main(ActivityThread.java:8751)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:571)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1135)
       
```
      
       
```
       Caused by android.app.ForegroundServiceStartNotAllowedException: Service.startForeground() not allowed due to mAllowStartForeground false: service com.rnapp.app/com.doublesymmetry.trackplayer.service.MusicService
       at android.app.ForegroundServiceStartNotAllowedException$1.createFromParcel(ForegroundServiceStartNotAllowedException.java:54)
       at android.app.ForegroundServiceStartNotAllowedException$1.createFromParcel(ForegroundServiceStartNotAllowedException.java:50)
       at android.os.Parcel.readParcelable(Parcel.java:3346)
       at android.os.Parcel.createExceptionOrNull(Parcel.java:2433)
       at android.os.Parcel.createException(Parcel.java:2422)
       at android.os.Parcel.readException(Parcel.java:2405)
       at android.os.Parcel.readException(Parcel.java:2347)
       at android.app.IActivityManager$Stub$Proxy.setServiceForeground(IActivityManager.java:8095)
       at android.app.Service.startForeground(Service.java:733)
[...]

PR #2028, opened 2023-06-07
fix(Android): improved foreground service handling

johsoe:
Additional improvements on foreground service handling. Fixes https://github.com/doublesymmetry/react-native-track-player/issues/1986

Inspired by still seeing crashes on a client project and taking inspiration from other open source projects, we found that continuing to call `startForeground(...)` when not in the foreground causes the `ForegroundServiceStartNotAllowedException` to be thrown after a while.

Our findings are that once the initial first call to `startForeground()` succeeds and we are in the background, we are good to go. The documentation from Google is a bit unclear (to say the least) on this.

Inspiration from VLC: 
https://code.videolan.org/videolan/vlc-android/-/blob/master/application/vlc-android/src/org/videolan/vlc/PlaybackService.kt#L762
https://code.videolan.org/videolan/vlc-android/-/blob/master/application/vlc-android/src/org/videolan/vlc/PlaybackService.kt#L917

PocketCasts:
https://github.com/Automattic/pocket-casts-android/blob/ee8da0c095560ef64a82d3a31464491b8d713104/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackService.kt#L249-L261

### DIFF
--- a/android/src/main/AndroidManifest.xml
+++ b/android/src/main/AndroidManifest.xml
@@ -5,6 +5,9 @@
     <uses-permission android:name="android.permission.WAKE_LOCK" />
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
 
+    <uses-permission
+        android:name="android.permission.FOREGROUND_SERVICE_MEDIA_PLAYBACK" />
+
     <application>
 
         <!-- The main service, handles playback, playlists and media buttons -->
--- a/android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
+++ b/android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
@@ -3,6 +3,7 @@ package com.doublesymmetry.trackplayer.service
 import android.app.*
 import android.content.Context
 import android.content.Intent
+import android.content.pm.ServiceInfo
 import android.net.Uri
 import android.os.Binder
 import android.os.Build
@@ -25,6 +26,7 @@ import com.doublesymmetry.trackplayer.model.Track
 import com.doublesymmetry.trackplayer.model.TrackAudioItem
 import com.doublesymmetry.trackplayer.module.MusicEvents
 import com.doublesymmetry.trackplayer.module.MusicEvents.Companion.EVENT_INTENT
+import com.doublesymmetry.trackplayer.utils.AppForegroundTracker
 import com.doublesymmetry.trackplayer.utils.BundleUtils
 import com.doublesymmetry.trackplayer.utils.BundleUtils.setRating
 import com.facebook.react.HeadlessJsTaskService
@@ -108,10 +110,13 @@ class MusicService : HeadlessJsTaskService() {
             )
         }
 
-        val notification = NotificationCompat.Builder(this, name)
+        val notificationBuilder = NotificationCompat.Builder(this, name)
             .setPriority(PRIORITY_LOW)
             .setCategory(Notification.CATEGORY_SERVICE)
-            .build()
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S) {
+           notificationBuilder.setForegroundServiceBehavior(NotificationCompat.FOREGROUND_SERVICE_IMMEDIATE)
+        }
+        val notification = notificationBuilder.build()
         startForeground(EMPTY_NOTIFICATION_ID, notification)
         @Suppress("DEPRECATION")
         stopForeground(true)
@@ -507,7 +512,21 @@ class MusicService : HeadlessJsTaskService() {
             event.notificationStateChange.collect {
                 when (it) {
                     is NotificationState.POSTED -> {
-                        startForeground(it.notificationId, it.notification)
+                        try {
+                            if (AppForegroundTracker.foregrounded) {
+                                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
+                                    startForeground(it.notificationId, it.notification, ServiceInfo.FOREGROUND_SERVICE_TYPE_MEDIA_PLAYBACK)
+                                } else {
+                                    startForeground(it.notificationId, it.notification)
+                                }    
+                            }
+                        } catch (e: Exception) {
+                            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S &&
+                                e is ForegroundServiceStartNotAllowedException
+                            ) {
+                                // Potentially log - ForegroundServiceStartNotAllowedException: startForeground(it.notificationId, it.notification) called from background...
+                            }
+                        }
                     }
                     is NotificationState.CANCELLED -> {
                         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {


jspizziri:
@puckey @dcvz could both of you have a look at this one?

martinmidtsund:
@johsoe If you have done some research on this, did you consider not running the service in foreground until there actually is sound playing? This is an interesting post regarding this area, of what gets killed and not, and they recommend being careful with the foreground state, "Be a good citizen and build your app such that you are working at the appropriate priority at all times." https://medium.com/androiddevelopers/who-lives-and-who-dies-process-priorities-on-android-cb151f39044f 

Like PocketCasts is doing here on PLAY, not on POSTED, just a few lines above your previous link : https://github.com/Automattic/pocket-casts-android/blob/ee8da0c095560ef64a82d3a31464491b8d713104/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackService.kt#L247

and then they stop it being foregrounded when the sound stops here: https://github.com/Automattic/pocket-casts-android/blob/ee8da0c095560ef64a82d3a31464491b8d713104/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackService.kt#L289

Sorry for the long comment, just wanted to share my thoughts on this, as we are having the same problem, but haven't been able to find the time to test the implementation of a fix. Great that you have! 👍

johsoe:
> @johsoe If you have done some research on this, did you consider not running the service in foreground until there actually is sound playing? This is an interesting post regarding this area, of what gets killed and not, and they recommend being careful with the foreground state, "Be a good citizen and build your app such that you are working at the appropriate priority at all times." https://medium.com/androiddevelopers/who-lives-and-who-dies-process-priorities-on-android-cb151f39044f
> 
> Like PocketCasts is doing here on PLAY, not on POSTED, just a few lines above your previous link : https://github.com/Automattic/pocket-casts-android/blob/ee8da0c095560ef64a82d3a31464491b8d713104/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackService.kt#L247
> 
> and then they stop it being foregrounded when the sound stops here: https://github.com/Automattic/pocket-casts-android/blob/ee8da0c095560ef64a82d3a31464491b8d713104/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackService.kt#L289
> 
> Sorry for the long comment, just wanted to share my thoughts on this, as we are having the same problem, but haven't been able to find the time to test the implementation of a fix. Great that you have! 👍

Yeah we could potentially look into doing it less frequently. I have a slight suspicion that there are some penalty in doing it too often as well. 

With that being said we've been getting good enough results with this minor change and I think going forward the right move for this library is changing the underlying dependencies away from `kotlin-audio` and `exoplayer` directly and use the newer `Media3` suite of libraries.

martinmidtsund:
@puckey 

> what if playback starts when the app is in the background?

Then I think it is a question of how the playback is started. Google say you can start a foreground process from the background for these reasons, where one reason is "The service starts by interacting with a notification." which is the only reason a track player should start in the background? ref https://developer.android.com/guide/components/foreground-services#bg-access-restriction-exemptions

>  How about when a user paused the audio in the lock screen – wouldn't disabling foregrounding cause the app to be killed?

That might be, but ref this https://developer.android.com/guide/components/foreground-services#remove-from-foreground it says the service (and then the notification, I guess) should stay put (obviously if not under memory pressure etc). And if the notification stays, the app should start right up again, with playback, when notification is pressed and go into foreground mode again? And we are experiencing users having the app being killed even if the service is foregrounded, so I guess it's never a guarantee.

puckey:
> @puckey
> 
> > what if playback starts when the app is in the background?
> 
> Then I think it is a question of how the playback is started. Google say you can start a foreground process from the background for these reasons, where one reason is "The service starts by interacting with a notification." which is the only reason a track player should start in the background? ref https://developer.android.com/guide/components/foreground-services#bg-access-restriction-exemptions

Ah I see now that in the PocketCasts implementation foregrounding already happens when buffering starts. So the app ought to be in the foreground in that case. I am going to take a look to see how trivial it would be to replicate their approach. (Still baffles me how many hoops you need to jump through to get a stable media player on Android)

puckey:
@martinmidtsund how about something like this: https://github.com/puckey/react-native-track-player/commit/05b7b851f77b1112b5790a08861013b020921973

johsoe:
> @martinmidtsund how about something like this: [puckey@05b7b85](https://github.com/puckey/react-native-track-player/commit/05b7b851f77b1112b5790a08861013b020921973)

I like the approach, but I think a library should give logging responsibility to the consumer. 

Currently we can only broadcast errors via `playbackError` as I see it - does this count as a playback error?

kirillzyusko:
@johsoe I also agree with you that we should give an ability to user to log the error. However I don't think foreground exception can be considered as `playbackError` - maybe better to add new type of error? Maybe `playerError`?

@puckey or your assumption is that we shouldn't notify user about such errors?

martinmidtsund:
@puckey This looks promising! It would be nice if the error would bubble, as mentioned by @johsoe , but I believe you are on the right path here, great work! 👏

puckey:
I think the best approach would indeed be to emit some kind of error event in order to allow the developer to show some kind of message that the user could adjust their battery savings settings..

The annoying thing is that I still have no way to easily reproduce this error – does anyone have a list of steps to take to replicate?

I noticed that PocketCast even is able to popup a dialog to allow the user to adjust their battery settings for the app, without having to direct them to the settings app. The question is where our implementation should end..

kirillzyusko:
> The question is where our implementation should end..

@puckey I believe emitting an event is enough 😀

puckey:
Added emitting of `Event.AndroidForegroundServiceStartNotAllowed` on failure here: https://github.com/puckey/react-native-track-player/commit/3ecf7b2675d5412401c7527f8da1bd36cd7eb06d

kirillzyusko:
@puckey maybe it would be better to emit a generic error (like `PlayerError`) and add `description`/`type` field, where we could specify concrete error (`android-foreground-service-start-not-allowed`).

In this case this error could be reused several times for future needs 🤔

johsoe:
> The annoying thing is that I still have no way to easily reproduce this error – does anyone have a list of steps to take to replicate?

Try `adb shell dumpsys deviceidle step light` to force the app into light idle. Next track change should hit the notification change event and call `startForeground()`. 

Alternative is just waiting until it happens with the app in the background playing on..

puckey:
@johsoe `adb shell dumpsys deviceidle step light` doesn't seem to have any effect on my samsung test phone.. So I am still unable to see my https://github.com/puckey/react-native-track-player/tree/foregrounding branch actually fixes this issue

puckey:
I have finished off the pr at https://github.com/doublesymmetry/react-native-track-player/pull/2034

Looking forward to feedback on it. Especially since I have yet to experience a `ForegroundServiceStartNotAllowedException` while testing on a `Galaxy A13` with `Android 13`

dodgex:
am I wrong or can this PR be closed as it is superseded by https://github.com/doublesymmetry/react-native-track-player/pull/2034 ?

bradleyflood:
Completed in https://github.com/doublesymmetry/react-native-track-player/pull/2034

Great work everyone.

PR #2029, opened 2023-06-09
fixed background crash

DineshRajput21:


puckey:
duplicate - will reopen if there is a good reason

PR #2033, opened 2023-06-16
fix: resolving imported local assets

puckey:
I noticed the example app was not displaying the image of the locally imported track `Pure (demo)`. This pr fixes things.


puckey:
Should be good to go now

puckey:
Hmm looks like while it was working, there are some inconsistencies in how asset resolving was implemented.. I will need to take a second look at this and do some testing

puckey:
So I encountered and fixed a few more inconsistencies: https://github.com/doublesymmetry/react-native-track-player/commit/29ae37d3db891f0340b160cbe0cb8674a75cdc77

- use resolveImportedAsset for options expecting `number | undefined`
- use resolveImportedAssetOrPath for options expecting `number | string | undefined`
- fixed the difference between what a track is that can be passed to add vs a track that is returned from `getQueue` by leaving out `ResourceObject` from `Track.url` and `Track.artwork` and creating a second `AddTrack` type which includes them.
- in example app avoid setting a faulty uri when `track?.artwork` is undefined

puckey:
@jspizziri since you have quite some experience with the frontend code, would you mind taking a second look at this?

Issue #2036, opened 2023-06-19
Q: State.Buffering ambiguity in play state


fivecar:
I couldn't get this answered on Discord despite asking two times, so thought I'd post here...

In v4's move to `getPlaybackState()` over `getState()`, the migration docs say that `State.Buffering` will now be returned regardless of whether it's supposed to be playing or not. In v3, I could use `if (state === State.Playing || state === State.Buffering)` to determine whether the player is in the act of trying to play (e.g. it can be used to dictate whether a play button shows a Pause icon or a Play icon). But it's unclear to me now how to do this in v4, because I can no longer tell between `state === State.Buffering` because you're buffering during playback vs. because you're paused and the player is pre-caching some upcoming audio.

Do I understand this correctly? If not, what is the right way in v4 to determine whether to render a Play or Pause icon by consulting the state of TrackPlayer?

puckey:
You can use ‘TrackPlayer.getPlayWhenReady()’ to find out whether the player will start playing when it had loaded enough data to do so.

puckey:
You may also want to check the example app for implementation reference

fivecar:
Ah, ok. So IIUC, `getPlayWhenReady` is true iff the player is in a "playing" state (regardless of buffer status), and is false iff the player is in a paused or stopped state? The naming of the function had me otherwise thinking that the call is only valid when the player isn't ready to play a track yet.

fivecar:
@puckey: IIUC, this change in the v4 API makes it really tricky to implement even a basic play/pause button. According to the example, all of the following is needed before you can decide whether to render Play or Pause:

```
  const playWhenReady = usePlayWhenReady();
  const isLoading = useDebouncedValue(
    state === State.Loading || state === State.Buffering,
    250
  );

  const isErrored = state === State.Error;
  const isEnded = state === State.Ended;
  const showPause = playWhenReady && !(isErrored || isEnded);
  const showBuffering = playWhenReady && isLoading;
```

Does that sound right? Including subtleties like you can be `playWhenReady` even when the track is errored or has already ended. Seems like it'd be very easy for developers to make a mistake here. Would a hook that encapsulates all this potentially be helpful?

fivecar:
#2040 submitted to make this easier for others. Thanks for clarifying the subtleties, @puckey .

tsheaff:
[...]

PR #2040, opened 2023-06-24
feat: useIsPlaying hook and isPlaying() method

fivecar:
This addresses some of the trickiness and confusion that users may have in v4's API, where `State.Buffering` spans both playing and paused states (whereas in v3, you could just `const playing = state === State.Playing || state === State.Buffering`). #2036 shows the subtlety of doing this right.

This PR adds two things to the public API:
- `useIsPlaying`: a hook that will likely be used by anyone managing a Play button. It correctly tells you when the button should be shown as "playing" and when it should be shown as "buffering."
- `isPlaying()`: an async function that gives you the most current playing state. This is needed for cases where you can't use a hook (e.g. outside of React components) or for cases where a hook's delay (due to subscriptions to events) isn't enough for you to make the right fork in your code.

I've added Javadoc-style comments, but am unsure where to update The Real Documentation™. I've also updated the example app for this simpler way of doing things.

### DIFF
--- a/docs/docs/guides/play-button.md
+++ b/docs/docs/guides/play-button.md
@@ -1,0 +1,24 @@
+---
+sidebar_position: 4
+---
+
+# Play Buttons
+
+UI often needs to display a Play button that changes between three states:
+
+1. Play
+2. Pause
+3. Spinner (e.g. if playback is being attempted, but sound is paused due to buffering)
+
+Implementing this correctly will take a bit of care. For instance, `usePlaybackState` can return `State.Buffering` even if playback is currently paused. `usePlayWhenReady` is one way to check if the player is attempting to play, but can return true even if `PlaybackState` is `State.Error` or `State.Ended`.
+
+To determine how to render a Play button in its three states correctly, do the following:
+
+* Render the button as a spinner if `playWhenReady` and `state === State.Loading || state === State.Buffering`
+* Else render the button as being in the Playing state if `playWhenReady && !(state === State.Error || state === State.Buffering)`
+* Otherwise render the button as being in the Paused state
+
+To help with this logic, the API has two utilities:
+
+1. The `useIsPlaying()` hook. This returns `{playing: boolean | undefined, bufferingDuringPlay: boolean | undefined}`, which you can consult to render your play button correctly. You should render a spinner if `bufferingDuringPlay === true`; otherwise render according to `playing`. Values are `undefined` if the player isn't yet in a state where they can be determined.
+2. The `async isPlaying()` function, which returns whether the TrackPlayer is (attempting to) play at any point. Note that you can't easily just instead call `getPlaybackState()` to determine the same answer, unless you've accounted for the issues mentioned above.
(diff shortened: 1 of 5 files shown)


puckey:
This is a really good addition to the library - thanks!

fivecar:
Done, @dcvz. Apparently had an empty blank line at the end of index.ts.

puckey (changes requested):

fivecar:
@puckey : made both updates. PR now returns undefined for both `useIsPlaying()` as well as `isPlaying()` when their states are undefined.

Also added docs.

puckey:
`useIsPlaying()` is not returning `undefined` for `playing` & `bufferingDuringPlay` yet – it would need to be something like:

```ts
function determineIsPlaying(playWhenReady?: boolean, state?: State) {
  if (playWhenReady === undefined || state === undefined) {
    return { playing: undefined, bufferingDuringPlay: undefined };
  }

  const isLoading = state === State.Loading || state === State.Buffering;
  const isErrored = state === State.Error;
  const isEnded = state === State.Ended;

  return {
    playing: playWhenReady && !(isErrored || isEnded),
    bufferingDuringPlay: playWhenReady && isLoading,
  };
}
```

Also, as it is I think the typescript return type of `isPlaying()` will unnecessarily be `Promise<undefined | boolean>`. So it might be an idea to do the following to fix that:

```ts
export function useIsPlaying() {
  const { state } = usePlaybackState();
  const playWhenReady = usePlayWhenReady();
  return playWhenReady === undefined || state === undefined
    ? { playing: undefined, bufferingDuringPlay: undefined }
    : determineIsPlaying(playWhenReady, state);
}

function determineIsPlaying(playWhenReady: boolean, state: State) {
  const isLoading = state === State.Loading || state === State.Buffering;
  const isErrored = state === State.Error;
  const isEnded = state === State.Ended;

  return {
    playing: playWhenReady && !(isErrored || isEnded),
    bufferingDuringPlay: playWhenReady && isLoading,
  };
}

export async function isPlaying() {
  const [playbackState, playWhenReady] = await Promise.all([
    TrackPlayer.getPlaybackState(),
    TrackPlayer.getPlayWhenReady(),
  ]);
  const { playing } = determineIsPlaying(playWhenReady, playbackState.state);

  return playing;
}
```

Finally, why not include the full object returned from `determineIsPlaying()` in `isPlaying()`?

fivecar:
@puckey : Ok, I've made all changes.

1. `useIsPlaying` now returns double undefineds if either PlaybackState or PlayWhenReady aren't known.
2. `isPlaying` now returns both `playing` and `bufferingDuringPlay`, instead of just one of them.

Note that the original `Promise<boolean|undefined>` from `isPlaying` was in response to code review feedback that I should do that, IIRC. Basically return `undefined` whenever the state isn't fully known. I think I've now changed things to where everyone's feedback is accounted for. LMK.

puckey (approved):

tsheaff:
I know it's been almost a year since this came out, but just want to thank @fivecar for this excellent PR. Extremely helpful!

PR #2043, opened 2023-06-27
feat: Android Auto Support

lovegaoshi:
This PR adds android auto support as discussed in [here](https://github.com/doublesymmetry/react-native-track-player/discussions/1984).

What I did:

- I copied `HeadlessJsTaskService.java `from the react-native repo and extended `MediaBrowserServiceCompat` instead of `Service`. No further changes other than adding a few imports.
- I added an if in `onBind()` that returns the `MediaBrowserServiceCompat.onBind()` when intent is not null, and the default` binder `when intent is null (which is used by RNTP's serviceConnector)
- I added relevant declarations as outlined [here](https://developer.android.com/training/cars/media).

This uses [my kotlinAudio fork](https://github.com/lovegaoshi/KotlinAudio/releases/tag/v2.0.0-aa9) that exposes its mediaSession's token, I'm submitting a PR there as well.

testing on an android auto emulator is described [here](https://developer.android.com/training/cars/testing).

I achieved synced media playback control both in app and android auto as shown below.

Capture

relevant links:
https://developer.android.com/training/cars/media
https://developer.android.com/training/cars/testing

I also tried to put `MediaBrowserServiceCompat` inside `MusicService` [here](https://github.com/doublesymmetry/react-native-track-player/compare/main...lovegaoshi:react-native-track-player:dev-android-auto-inner-service) but I was having some errors. If this works then we don't have to mess with maintaining a copied `HeadlessJsTaskService.java `.


lovegaoshi:
new update:
I added content hierarchy support that should complete this PR functionality-wise. For users they must implement event listeners [RemotePlayId and RemotePlaySearch](https://github.com/doublesymmetry/react-native-track-player/pull/2043/files#diff-1e0f7428565ec6903d4b20e677df280e6248b707031ff02cd82687c4cee773cd) to properly handle playback calls from AA. Content hierarchy can be built as outlined [here](https://github.com/doublesymmetry/react-native-track-player/pull/2043/files#diff-a27fec4b4a6500c4a11a3a464bd6facb806a69411f6c81ba594fcf81db50878a). It will look like below:


Capture

I will try to submit my app to the appstore and see if google complains any features not implemented for AA

lovegaoshi:
i have some test ideas to fix 
> MediaBrowserServiceCompat can run without UI/without RN app actually started on mobile phone

at https://github.com/lovegaoshi/react-native-track-player/issues/4#issuecomment-1654637773 . wish someone with android experience can chime in

lovegaoshi:
@czekFREE I actually encountered the first problem you described quite often this past week, guess i was just ignoring it until now; and after too many frustrating debug prints, I came up with something probably sketchy and definitely could use some advice. 

The issue is exactly as you described, MusicService is created but MainActivity is not.  its mentioned some years ago at https://stackoverflow.com/questions/44354261/android-auto-communication-between-background-service-and-activity . However I'm not understanding that RNTP already has its activity and service separated, and it just appears that MusicService isn't even created at all - and it doesnt! I put a timber.d inside override onCreate and it doesnt print when launched from AA. I don't understand that not even extending MusicService to MediaBrowserCompat gets it printed. While UAMP works just as expected, MusicService's onCreated gets called both from UI and from AA. 

I did find out that overriding onCreate in HeadlessJsMediaService will get printed just fine. I'm then launching an intent to call up the UI via deeplinking at its onCreate, so UI will be brought up, RNTP will get properly set up and all listeners will be properly registered, rather than trying to be smart about setting up service without the activity like AIMP or UAMP does. I have this change below though I dont really understand what I'm doing with 3 weeks of android experience... Ill use it in my app for a while.
 
https://github.com/lovegaoshi/react-native-track-player/commit/597106563bb644487c28d4bdf0df43118869ae70

lovegaoshi:
@czekFREE welp, you will be surprised - my app with this capability passed google's open testing review just yesterday. 

IMO this is very much dwelling into RN's limitations - MediaBrowserService will start independently from MainActivity, without MainActivity RN will not start, RNTP will not be set up, the whole RNTP part wont work, without bringing up the UI the entire RNTP package becomes nonsense, should have started with native to begin with. And this also means there is no way to customize this behavior (bringing up the UI from HeadlessJSMediaService's onCreate via deeplinking) from the RN side. 

and imo an error message here is really bad - AA apps should never prompt the user to use their phone and open an app. Auto start MainActivity from MediaBrowserService is the necessary lesser evil.

I'll leave this change in my own branch until the devs weighs in...

PR #2062, opened 2023-07-18
refactor: don't recreate INITIAL on every re-render

kirillzyusko:
Really minor optimisation which avoids object re-creation/destructurization on every re-render.

Since `useProgress` changes its state every second, every second it'll create a new object and then move it to GC. In my opinion it could be avoided and `INITIAL` variable can be initialized only one time 🙂 

Note, that current code doesn't introduce any performance issues - this PR it's really minor optimization 🙂


kirillzyusko:
@dcvz I fixed all remarks. Please, review again 😊

PR #2076, opened 2023-07-27
chore(capabilities): Revamp Capabilities

dcvz:
This PR revamps the system of capabilities and the `updateOptions` parameters. It associates all related objects together to simplify and give finer control over options.

Capabilities now differentiate (particularly on Android) between items that:
1. should be used as media session actions
2. should be included in the notification (and whether they are allowed in compact mode)

The Android side requires to be used in conjunction with: https://github.com/doublesymmetry/KotlinAudio/pull/86


puckey:
I really like the direction this is headed! On the typescript side of things, I was wondering what you think of a format like:

```ts
  capabilities: {
    play: true,
    pause: true,
    skipToNext: true,
    skipToPrevious: {
      // optional: enabled: true,
      compact: false,
      icon: require('./previous.png')
    },
    seekTo: true
  }
})
```

- we are no longer dealing with order, so an object is more fitting than an array
- no need for any extra imports
- no difference between what you see (function calls) and the data that it produces
- nicely typeable

Issue #2090, opened 2023-08-11
reenable artwork caching on android


puckey:
In the latest version of KotlinAudio the disk / memory caching of artwork has been disabled. Since SwiftAudioEx does use caching through its use of `URLSession.shared.dataTask`, this creates an inconsistency between ios and android.

Might it be an idea that we re enable caching  for the time being and then possibly release it as configurable option with platform parity after 4.0 has been released?

I can imagine that most use cases would want imagery to be cached – and those that expect the same url to produce varying results can always use a cache busting technique such as adding something to their url parameter like `http://domain.com/url.mp3?cache=${Date.now()}`.

In any case I would expect a library like coil to respect cache headers – so you could probably force it to not use the cache by setting the correct headers. But I haven't checked this to be certain.

PR #2092, opened 2023-08-11
chore(android): undeprecate updateNowPlayingMetadata & restore image caching

### LINKED ISSUE #2089: Undeprecate updateNowPlayingMetadata and fix implementation
`updateNowPlayingMetadata()` should not be deprecated, since it has a slightly different usecase to `updateMetadataForTrack`. When you set metadata through `updateNowPlayingMetadata()`, it will be automatically removed again when you switch to another track. So when you switch back to the audio item you will see the last set metadata, since `updateMetadataForTrack` also changes the fields of the audio item.

In a radio app for example, you might want to initially see the name of the radio station and then when it starts playing and the fresh metadata flows in, you can update the now playing metadata to what is playing. In this case, if `updateNowPlayingMetadata` is removed, you will see stale metadata when you move back to a previously played track. To remedy this, you would have to perform a few calls each time from userland code to reset the audio item back to its original form again.

Also the new implementation of the deprecated `updateNowPlayingMetadata()` method now works the same as `updateMetadataForTrack()` under the hood.

dcvz:
Resolves #2089 & #2090

### DIFF
--- a/example/src/components/ActionSheet.tsx
+++ b/example/src/components/ActionSheet.tsx
@@ -38,9 +38,9 @@ export const ActionSheet: React.FC = () => {
         type={'primary'}
       />
       <Button
-          title={'Update Current Track Metadata Randomly'}
-          onPress={onUpdateCurrentTrackMetadata}
-          type={'primary'}
+        title={'Update Current Track Metadata Randomly'}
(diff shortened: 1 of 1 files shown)


puckey (approved):
lgtm

PR #2121, opened 2023-09-09
feat: volume fade in/out

lovegaoshi:
adds volume fade in/out effect.

on not android platforms this is done by using Animated.Value. I do not have iOS to test but it seems like iOS doesn't have a background react rendering problem that this should work.

on android this is done by using https://github.com/doublesymmetry/KotlinAudio/pull/91. because Animated.Value doesnt update within HeadlessJsTaskService I added a simple linear fading function for it. 

kotlin's scope.async doesn't work as I expected (Deferred.await() doesn't wait for RN's Promise.resolve?) any help is appreciated! because of this I cannot reliably await for the volume fading change to complete, so I used event emitting and handle the callback function within that event emitting, which is quite unnecessarily complicated, compared to a simple callback function in Animated.Timing.Start().


mihaibulic2:
I'd love to see this get submitted into 4.1.0 — @dcvz do you think that's possible?

lovegaoshi:
bump

lovegaoshi:
bump

PR #2152, opened 2023-10-05
fix (events): prefix all new metadata event payloads with “metadata”

### LINKED ISSUE #2147: Issues with the new metadata events in combination with useTrackPlayerEvents
Until now all emitted events have been plain javascript objects. The new `MetaData*Received` events all are emitted as arrays. This causes issues with the `useTrackPlayerEvents` hook, since it adds a `type` property on to the received event object, in order to allow distinguishing the event types from each other (since you can subscribe to multiple events in one go).

Because the `MetaData*Received` events are arrays, this causes the way they are serialized to JSON te become:

```json
{
  "type": "metadata-timed-received",
  "0": {
    "title": "Live Now - Helmets w/ Hablamos Muertos",
    "raw": [
      {
        "key": "StreamTitle",
        "commonKey": "title",
        "value": "Live Now - Helmets w/ Hablamos Muertos",
        "time": 0.6530612244897959,
        "keySpace": "icy"
      }
    ]
  }
}
```

A few ways we could solve this situation:
- modify  the `useTrackPlayerEvents` hook to call the callback with `{ type: string; payload: event-object-or-array }`
- modify the value emitted through `Event.MetadataChapterReceived`, `Event.MetadataTimedReceived`, `Event.MetadataCommonReceived` to be something like `{ metadata: array-of-metadatas }`

dcvz:
Resolves #2147

### DIFF
--- a/src/interfaces/events/EventPayloadByEvent.ts
+++ b/src/interfaces/events/EventPayloadByEvent.ts
@@ -56,6 +56,6 @@ type Simplify<T> = { [KeyType in keyof T]: T[KeyType] } & {};
 
 export type EventPayloadByEventWithType = {
   [K in keyof EventPayloadByEvent]: EventPayloadByEvent[K] extends never
-  ? { type: K }
-  : Simplify<EventPayloadByEvent[K] & { type: K }>;
+    ? { type: K }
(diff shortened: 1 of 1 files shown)


puckey (approved):
nice!

Issue #2159, opened 2023-10-11
[Android 12+] Loading new track crashes the app after being in background for ~5 mins with locked screen 


takacskalman:
**Describe the Bug**

When trying to load a new track when the previous ended (via the PlaybackService callback) the app crashes. It just happens if the screen is locked and the app is in background and it only affects Android 12+ versions. Prior Android 12 it works fine without any issues, also on IOS no issues. It works with local files, just the remote ones cause the problem.
It works if we load all the tracks by calling `TrackPlayer.add([list of tracks])` but it does not work if I we call `TrackPlayer.load` and switch the track from the PlaybackService's callback by `TrackPlayer.load`. 
Our problem is that we cant add all the tracks at once, we need to add them 1 by 1 because we have custom logic for preparing a track. 

**Steps To Reproduce**
1. Load a track via `TrackPlayer.load` method
2. Load the next track in PlaybackService's PlaybackState event's State.Ended callback (again via `TrackPlayer.load`)
3. Start the player
4. Move the app to the background
5. Lock the screen
6. Listen for ~5 minutes
7. It should crash

**Code To Reproduce**
Provided in the next section.

**Replicable on Example App?**
Can you replicate this bug in the [React Native Track Player Example App](https://github.com/doublesymmetry/react-native-track-player/tree/main/example)?

It's replicable in the demo app by changing it's playlist management from queue added to track player to loading 1 track at a time and then changing the track when the actual ended. The screenshots with the chnages:
![Screenshot 2023-10-11 at 18 06 06](https://github.com/doublesymmetry/react-native-track-player/assets/29872514/db71befb-1a80-4718-8e58-d765681fb3c0)

![Screenshot 2023-10-11 at 18 07 56](https://github.com/doublesymmetry/react-native-track-player/assets/29872514/6fff0081-0c07-48dd-aad1-cdc3c24ed3d3)

![Screenshot 2023-10-11 at 18 08 22](https://github.com/doublesymmetry/react-native-track-player/assets/29872514/168a43e7-5e3b-4110-a140-dcbe6efea32f)

**Environment Info:**
Paste the results of `npx react-native info` It was tested by different developers on different platforms
Paste the exact `react-native-track-player` version you are using 4.0.0-rc09 but it's the same with other versions too
Real device? Or simulator? Real device
What OS are you running? Android

**How I can Help**
[...]

Issue #1845, opened 2022-12-07
[Feature Request] Optimize Android resource usage


elliotdickison:
**What is the need and use case of this feature?**

When loading local files/tracks into the player on Android, the full contents of all files/tracks are loaded into memory immediately. For large track lists this results in rapid memory consumption and can end with Android killing the app.

**Describe the ideal solution**

It would be excellent if only the current track was loaded into memory (or maybe the current + previous + next if that results in a faster user experience).

**Describe alternatives you've considered**

I am currently limiting my app to queueing 5 tracks at a time. I load 5 tracks initially and then when the PlaybackQueueEnded event is fired I clear out those 5 and load 5 more. This isn't ideal for a few reasons:
- It's extra logic in my app that feels like it belongs in RNTP
- Dynamically adjusting the queue has some quirks: the lock screen controls temporarily disappear every 5 tracks while I reset the queue, it takes a bit longer to move on to the next track at those points, and I have to manually restart playback (see #1691)
- If my app's task is put on ice by Android then it isn't around to load up the next 5 tracks and playback will stop prematurely

**Additional context**

The particular track list that brought this issue to my attention was ~160 tracks totaling 2.7GB (a year-long Bible reading plan).

I am using expo-file-system to download files and then I'm passing the fileURIs to RNTP.

**How I can Help**

- I can do a deep dive on resource usage to check for additional things like processing constraints on large track lists
- Loading large lists of tracks from the network doesn't have quite the same effect on device memory, but it's still very slow. I can attempt to diagnose that as well.

jspizziri:
@elliotdickison open to a PR for this. For now I'll move it to the feature backlog.

elliotdickison:
Just noting that this occurs whether loading local files or URLs. Tested on version 4 rc9.

elliotdickison:
PEBKAC: I was inadvertently stuffing insane amounts of extra data into the queued track objects. Slimming that data down solved the problem.

PR #2164, opened 2023-10-17
fix(android): Allow setting grace period before stopForeground

dcvz:
This PR adds a configurable grace period - a delay before we attempt to set our service to not be in foreground mode, allowing users a chance to do player actions that might refrain us from deciding to go out of foreground state. Defaults to five seconds.

**Context:**
At the moment on Android as soon as we transition back to idle/stop/error we immediately set our service as not being in the foreground state (setting yourself as not in foreground is good practice). In the use of RNTP where everyone's needs are different, a user might still want to enqueue some kind of playback action once they realize the queue is over, an error has happened, etc. For example, once a user's queue ends, they may want to create a whole new queue of related music.

Immediately setting ourselves as not in foreground state may give the OS a chance to turn off network chips, or kill the service which may result in errors if you try to do any network loading (i.e. playing a remote track) after that as happens on #2159.


elliotdickison:
This will be particularly helpful in the workaround for #1845

PR #2167, opened 2023-10-17
refactor: `usePlaybackState` without initial `undefined`

kirillzyusko:
The idea is to store the data in a singleton and consume it as an initial value for state.

Let me know if you have any concerns with such approach 👀

### DIFF
--- a/src/hooks/usePlaybackState.ts
+++ b/src/hooks/usePlaybackState.ts
@@ -1,42 +1,34 @@
 import { useEffect, useState } from 'react';
 
-import { getPlaybackState, addEventListener } from '../trackPlayer';
-import { Event } from '../constants';
+import { addEventListener } from '../trackPlayer';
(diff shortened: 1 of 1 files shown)


kirillzyusko:
@dcvz may I kindly ask you to review this PR? 👀

puckey:
I am not sure automatically installing a listener like is a good idea, since you are unnecessarily using resources if it goes unused - i.e. when you don't use the hook. If you want to cache this value across unmounts and remounts, another approach would be to lift your usage of the hook up to the root of you app and pass the value down as props / context.. In a any case you would always need to start out with `undefined` instead of `State.NONE`

kirillzyusko:
Ah, okay @dcvz and @puckey 

I understand your point of view and it makes sense to me 👍 Feel free to close this PR then 😊

PR #2181, opened 2023-10-21
fix: android notification removal flag when mediasession focus changed

lovegaoshi:
I started noticing in my app using nightly, the notification when media is not played will get removed when app enters background. This is caused by stopForeground is set to always remove notification when ForegroundGracePeriod expires.
this fix reverts the removeNotificationWhenNotOngoing flag in stopForeground introduced in b13a92208eaf9607a12cfa7596dfa6ae4f4a4c8e and fixes my issue.

### DIFF
--- a/android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
+++ b/android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
@@ -590,7 +590,7 @@ class MusicService : HeadlessJsTaskService() {
                                 delay(stopForegroundGracePeriod.toLong() * 1000)
                                 if (shouldStopForeground()) {
                                     @Suppress("DEPRECATION")
-                                    stopForeground(true)
+                                    stopForeground(removeNotificationWhenNotOngoing)
                                     Timber.d("Notification has been stopped")
                                 }
                             }


puckey:
@dcvz @lovegaoshi nice catch – looking over this code again, should we remove `AudioPlayerState.ERROR` from `REMOVABLE_STATES`? Since if you press play again, it will retry loading.. And currently it will just remove the notification if it reaches this state. The same goes for `AudioPlayerState.STOPPED` – you could press play again, for example.

puckey:
> > @dcvz @lovegaoshi nice catch – looking over this code again, should we remove `AudioPlayerState.ERROR` from `REMOVABLE_STATES`? Since if you press play again, it will retry loading.. And currently it will just remove the notification if it reaches this state. The same goes for `AudioPlayerState.STOPPED` – you could press play again, for example.
> 
> I wondered but didn’t want to try to make such a critical change leading up to v4 release. I also know you were testing extensively there and thought I might be missing something.

Yes, this code is based on https://github.com/Automattic/pocket-casts-android/blob/ee8da0c095560ef64a82d3a31464491b8d713104/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackService.kt#L218 – but looking at it again, I think we can diverge from it.. In any case, I removed it from the fork I am using for RG, because I want users to be able to press play again if a stream stops loading due to a connection error or loss of internet..

PR #2220, opened 2023-12-13
chore(docs) v4 migration requires SwiftUI weak dependency 

cjpete:


dcvz (approved):
Thanks for the addition!

PR #2233, opened 2024-01-03
refactor: the docs now recommend using the remove function on the AppState subscription

Pnlvfx:
This pull request addresses the deprecation of AppState.removeEventListener in favor of the recommended alternative in the React Native documentation. The changes ensure that the useAppIsInBackground hook remains up-to-date with the latest best practices.

React Native has deprecated AppState.removeEventListener in favor of a more robust and future-proof alternative. This change ensures that our codebase aligns with the latest recommendations, reducing the risk of issues in the future.

### DIFF
--- a/src/hooks/useAppIsInBackground.ts
+++ b/src/hooks/useAppIsInBackground.ts
@@ -9,10 +9,10 @@ export function useAppIsInBackground() {
       setState(nextState);
     };
 
-    AppState.addEventListener('change', onStateChange);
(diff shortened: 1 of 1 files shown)


dcvz:
@Pnlvfx do we know when this change was made in RN? We should make sure it'll work with the lowest RN version supported by RNTP as well.

Pnlvfx:
I think it's from version 0.65 on. The peer of RNTP indicates 0.60 as the minimum. I think we have to wait to implement it.. Sorry boss

PR #2234, opened 2024-01-05
Fix not_initialized error code

IjzerenHein:
The error-code that is emitted when the player is not initialized is the same as when the player is already initialized (the error messages themselves are okay).


uzegonemad:
@dcvz  @jspizziri Could we get this one merged?

jspizziri (approved):
LGTM @puckey?

dcvz (approved):
LGTM 👍 thanks for bringing this to par with Android

IjzerenHein:
Thanks for merging and this great library guys!

PR #2235, opened 2024-01-08
chore(android): distinguish url and artwork empty string error message

AronBe:
Current error message is not descriptive enough and empty artwork url string may be considered as an empty track url field. The player throws an error, so it would be good to catch the artwork may be the problem


dcvz (approved):
Great addition @AronBe !

PR #2249, opened 2024-02-06
chore(android): forward/rewind svg

lovegaoshi:
use [google material icon](https://m3.material.io/styles/icons/overview) svg instead of the blurry material pngs for  forward and rewind.


dcvz (approved):
Thanks for this @lovegaoshi ! drawables only really had support on API version 21 but I think as our min is RN 0.60 we should be good here.

PR #2258, opened 2024-02-27
chore: reduce npm package size

kirillzyusko:
Current npm package size is ~13mb:

image

Which is quite a lot for npm package. And it's mainly because of the fact that build files are included.

We ignore these files in `.npmignore`, but we use `files` property in `package.json` (which kind of has higher priority, so `.npmignore` properties are ignored).

In this PR I included files that are really needed (and omitted unnecessary files/directories, such as build/.idea/etc)

|Before|After|
|-------|-----|
|image|image|

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,9 @@
     "src",
     "lib/**/*",
     "ios/**/*",
-    "android/**/*",
+    "android/src/**/*",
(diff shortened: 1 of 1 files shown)


dcvz (approved):
Great improvement! Thanks @kirillzyusko

PR #2259, opened 2024-02-28
feat: adds iosCategory update functionality

watadarkstar:
**Why**

It allows you to switch modes at run time. For example, if you have an app that plays both audio and video you may need to update the `iosCategoryMode` at runtime 

For some reason we found that to get CarPlay to broadcast the audio from track player using audio it needs IOSCategoryMode.SpokenAudio and for video it needs IOSCategoryMode.VideoRecording.

**How**

I moved some of the logic into an `updateCategory` function so it can be reused and the `iosCategoryMode` can be updated as well as other ios properties (see docs I've updated)

### DIFF
--- a/ios/RNTrackPlayer/RNTrackPlayer.swift
+++ b/ios/RNTrackPlayer/RNTrackPlayer.swift
@@ -140,7 +140,7 @@ public class RNTrackPlayer: RCTEventEmitter, AudioSessionControllerDelegate {
     private func rejectWhenNotInitialized(reject: RCTPromiseRejectBlock) -> Bool {
         let rejected = !hasInitialized;
         if (rejected) {
-            reject("player_already_initialized", "The player is not initialized. Call setupPlayer first.", nil)
+            reject("player_not_initialized", "The player is not initialized. Call setupPlayer first.", nil)
         }
         return rejected;
     }
--- a/ios/RNTrackPlayer/RNTrackPlayer.swift
+++ b/ios/RNTrackPlayer/RNTrackPlayer.swift
@@ -140,7 +140,7 @@ public class RNTrackPlayer: RCTEventEmitter, AudioSessionControllerDelegate {
     private func rejectWhenNotInitialized(reject: RCTPromiseRejectBlock) -> Bool {
         let rejected = !hasInitialized;
         if (rejected) {
-            reject("player_not_initialized", "The player is not initialized. Call setupPlayer first.", nil)
+            reject("player_already_initialized", "The player is not initialized. Call setupPlayer first.", nil)
         }
         return rejected;
     }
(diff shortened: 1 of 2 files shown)


ios/RNTrackPlayer/RNTrackPlayer.swift
```diff
@@ -140,7 +140,7 @@ public class RNTrackPlayer: RCTEventEmitter, AudioSessionControllerDelegate {
     private func rejectWhenNotInitialized(reject: RCTPromiseRejectBlock) -> Bool {
         let rejected = !hasInitialized;
         if (rejected) {
-            reject("player_not_initialized", "The player is not initialized. Call setupPlayer first.", nil)
+            reject("player_already_initialized", "The player is not initialized. Call setupPlayer first.", nil)
```
watadarkstar, line 143:
```suggestion
            reject("player_not_initialized", "The player is not initialized. Call setupPlayer first.", nil)
```

ios/RNTrackPlayer/RNTrackPlayer.swift
```diff
@@ -346,6 +349,10 @@ public class RNTrackPlayer: RCTEventEmitter, AudioSessionControllerDelegate {
             interval: ((options["progressUpdateEventInterval"] as? NSNumber) ?? 0).doubleValue
         )
 
+        updateCategory(config: options)
+
+        configureAudioSession()
```
watadarkstar, line 354: @dcvz Can I call this here safely? I realize that reinitialization is something that should be avoided based on this code inside `setupPlayer`:

```
            reject("player_already_initialized", "The player has already been initialized via setupPlayer.", nil)
```

watadarkstar:
@dcvz @jspizziri 

When I was testing this on my app for some reason switching the `iosCategoryMode` causes the track player controls to disappear. I'm hoping you can provide some guidance as why that may be happening?

dcvz:
> @dcvz @jspizziri 
> 
> 
> 
> When I was testing this on my app for some reason switching the `iosCategoryMode` causes the track player controls to disappear. I'm hoping you can provide some guidance as to why that may be happening? 

Are you perhaps calling update with an empty capabilities array? Docs say that we update new capabilities, but I think we've found that not all of them do, and omitting them is like setting them as empty.

jspizziri:
Possibly related to https://github.com/doublesymmetry/react-native-track-player/pull/2285

ios/RNTrackPlayer/RNTrackPlayer.swift
Egbert-Jan: I think you can update this safely since the phone switches nicely between the two speakers when updating the iosCategory and iosCategoryMode properties.

PR #2265, opened 2024-03-11
docs(web, windows): add documentation for web and remove it for windows

jspizziri:


jspizziri:
@dcvz I removed the documentation for windows in favor of web since we don't really support it? I'm not sure if that's something you want to do, in which case just let me know and i can add Windows back in there in addition to web.

dcvz (approved):
LGTM 👍 Windows is very outdated, I'm good with mentioning it as a supported platform as it implies we keep it up to date

PR #2285, opened 2024-04-02
feat: make it possible to play audio through the earpiece speaker

Egbert-Jan:
This new functionality allows you to set the audioUsageType and audioContentType for Android, and the iosCategroy and iosCategoryMode for iOS with the updateOptions function. This makes it possible to force the audio to come through the earpiece speaker. This can for example be useful to play an audio message in a chat app. For this to work on Android I also made a pull request in KotlinAudio (https://github.com/doublesymmetry/KotlinAudio/pull/105). I tried but could not get the RNTP example app to use the updated KotlinAudio repo while executing the publishToMavenLocal steps as shown in https://github.com/doublesymmetry/react-native-track-player/tree/main/example. 

Example: 
TrackPlayer.updateOptions({
            iosCategory: IOSCategory.PlayAndRecord,
            iosCategoryMode: IOSCategoryMode.VoiceChat,
            androidAudioContentType: AndroidAudioContentType.Music,
            androidAudioUsageType: AndroidAudioUsageType.VoiceCommunication
})


Egbert-Jan:
Hello @jspizziri @dcvz, could you maybe take a look at this pull request? I think it would be a useful feature. If something is wrong with it I could fix it.

docs/docs/api/objects/player-options.md
```diff
@@ -15,4 +15,5 @@ All parameters are optional. You also only need to specify the ones you want to
 | `waitForBuffer` | `boolean` | Indicates whether the player should automatically delay playback in order to minimize stalling. Defaults to `true`. @deprecated This option has been nominated for removal in a future version of RNTP. If you have this set to `true`, you can safely remove this from the options. If you are setting this to `false` and have a reason for that, please post a comment in the following discussion: https://github.com/doublesymmetry/react-native-track-player/pull/1695 and describe why you are doing so. | ✅ | ✅ |
 | `autoUpdateMetadata` | `boolean` | Indicates whether the player should automatically update now playing metadata data in control center / notification. Defaults to `true`. | ✅ | ✅ |
 | `autoHandleInterruptions` | `boolean` | Indicates whether the player should automatically handle audio interruptions. Defaults to `false`. | ✅ | ✅ |
+| `androidAudioUsageType` | `boolean` | (Android only) Specifies why the source is playing and controls routing, focus, and volume decisions. With `androidAudioUsageType` set to VoiceCommunication the audio will come from the earpiece speaker | ✅ | ❌ |
```
jspizziri, line 18: @Egbert-Jan the type of this isn't a boolean, it should be the `AndroidAudioUsageType` enum. Then in the description you should use `AndroidAudioUsageType.VoiceCommunication`.

docs/docs/api/objects/player-options.md
```diff
@@ -15,4 +15,5 @@ All parameters are optional. You also only need to specify the ones you want to
...
 | `androidAudioContentType` | `boolean` | The audio content type indicates to the android system how you intend to use audio in your app. With `autoHandleInterruptions: true` and `androidAudioContentType: AndroidAudioContentType.Speech`, the audio will be paused during short interruptions, such as when a message arrives. Otherwise the playback volume is reduced while the notification is playing. Defaults to `AndroidAudioContentType.Music` | ✅ | ❌ |
```
jspizziri, line 19: @Egbert-Jan since you're doing work in this area, it would be great if you fixed the `type` of this one, as it's not a `boolean` either.

docs/docs/api/objects/update-options.md
```diff
@@ -25,5 +25,10 @@ All parameters are optional. You also only need to specify the ones you want to
...
 | `progressUpdateEventInterval` | `number` | The interval (in seconds) that the [`Event.PlaybackProgressUpdated`](../events.md#playbackprogressupdated) will be fired. `undefined` by default. | ✅ | ✅ | ✅ |
+| `iosCategory` | [`IOSCategory`](../constants/ios-category.md) | An [`IOSCategory`](../constants/ios-category.md). | ❌ | ✅ | ❌ |
+| `iosCategoryMode` | [`IOSCategoryMode`](../constants/ios-category-mode.md) | The audio session mode, together with the audio session category, indicates to the system how you intend to use audio in your app. You can use a mode to configure the audio system for specific use cases such as video recording, voice or video chat, or audio analysis. | ❌ | ✅ | ❌ |
+| `iosCategoryOptions` | [`IOSCategoryOptions[]`](../constants/ios-category-options.md) | An array of [`IOSCategoryOptions`](../constants/ios-category-options.md). | ❌ | ✅  | ❌ |
+| `androidAudioUsageType` | `boolean` | (Android only) Specifies why the source is playing and controls routing, focus, and volume decisions. With `androidAudioUsageType` set to VoiceCommunication the audio will come from the earpiece speaker | ✅ | ❌ |
```
jspizziri, line 31: @Egbert-Jan please fix the typings and descriptions here as well for both "usage" and "content" type.

src/interfaces/PlayerOptions.ts
```diff
@@ -4,6 +4,7 @@ import type {
   IOSCategoryMode,
   IOSCategoryOptions,
 } from '../constants';
+import { AndroidAudioUsageType } from '../constants/AndroidAudioUsageType';
```
jspizziri, line 7: Please export `AndroidAudioUsageType` from `constants/index.ts` so it can be imported in a consistent manner to the others.

jspizziri (changes requested):
@Egbert-Jan 

I left some low-level feedback. All in all, it looks like it probably works, but I'd need to test it. However, before we go any further there's a high level design question that needs to be answered (and @dcvz will need to weigh in on it).

Should things like `androidAudioUsageType` and `androidAudioContentType` be pushed into the `AndroidOptions` object that we introduced in recent versions. So rather than doing:

```ts
TrackPlayer.updateOptions({
   androidAudioUsageType: ...,
   androidAudioContentType: ...,
   iosCategory: ...,
   iosCategoryMode: ...,
   iosCategoryOptions: ...,
   ...
})
```

You'd do:

```ts
TrackPlayer.updateOptions({
  android: {
    usageType: ...,
    contentType: ...,
  },
  ios: {
    category: ...,
    categoryMode: ...,
    categoryOptions: ...,
  },
   ...
})
```

Since we're introducing this into `updateOptions` for the first time, I think that is the right approach personally. However, it does cause an api consistency problem, where `setupPlayer` would be different than `updateOptions` (since these options already exist on `setupPlayer` and we don't want to introduce a breaking change).

What I propose is that we add `android` and `ios` config to `setupPlayer` in an identical manner and only allow the new properties to be set in that way. Then we can add a simple shim in the TS side, that maps legacy config of `androidAudioContentType`, etc. into the new `android: AndroidOptions` property. We then mark the top level properties as deprecated.

This would IMO, allow for a more organized, and consistent API between both `updateOptions` and `setupPlayer`.

Let me know your thoughts.

Egbert-Jan:
Thanks I resolved the documentation and import issues. Have you also noticed I made a pull request here: https://github.com/doublesymmetry/KotlinAudio/pull/105. Those changed are needed for these to work.

jspizziri:
@watadarkstar & @Egbert-Jan , it seems that there is some overlap between your two PRs (https://github.com/doublesymmetry/react-native-track-player/pull/2259). Can the two of you evaluate and and if I'm not mistaken, collaborate?

Egbert-Jan:
My pull request introduces the same functionality as #2259 for the iOS side. However it also adds the functionality to the Android side.

tfkennedy777:
Just chiming in to say that I really appreciate the contributions on this issue, and I'm rooting for this to get merged! Would be super helpful to update these options on the fly after initialization.

PR #2290, opened 2024-04-05
chore: improve bridgeless interop layer compatibility

behenate:
### Why

In `react-native 0.74` when new architecture is enabled bridgeless mode will be enabled by default. 
By default modules which use the bridge will use an interop layer but it's not perfect. After creating a project, which uses `react-native 0.74.0-rc.6` both Android and iOS don't work at all under bridgeless.

### How

I managed to get the features of the example app working on iOS by removing some declarations from `RNTrackPlayerBridge.m`: `getSleepTimerProgress`, `setSleepTimer`, `sleepWhenActiveTrackReachesEnd` and `clearSleepTimer` are declared, but never implemented (correct me if I'm wrong), which causes a segmentation fault when using bridgeless. Removing them fixes the issue. All of the functionality seems to be working ok after the fix.

On Android some of the `@ReactMethod` annotated functions are assigned a value of `scope.launch`. This causes the interop layer to unnecessarily try to pass the returned `Job`s to the JS side, which fails. Added a wrapper for `scope.launch`, so that the assignment can be kept for better readability, but the function return type is changed to `Unit`. 

Unfortunately some things still don't work well with bridgeless on Android. Some of the events aren't sent correctly (eq. the `PlaybackActiveTrackChanged`). For some reason all of the notification controls become broken. Those issues are likely easy fixes, but I currently don't have time to learn the codebase and fix them 😕

### DIFF
--- a/android/src/main/java/com/doublesymmetry/trackplayer/module/MusicModule.kt
+++ b/android/src/main/java/com/doublesymmetry/trackplayer/module/MusicModule.kt
@@ -251,8 +251,8 @@ class MusicModule(reactContext: ReactApplicationContext) : ReactContextBaseJavaM
     }
 
     @ReactMethod
-    fun updateOptions(data: ReadableMap?, callback: Promise) = scope.launch {
-        if (verifyServiceBoundOrReject(callback)) return@launch
+    fun updateOptions(data: ReadableMap?, callback: Promise) = launchInScope {
+        if (verifyServiceBoundOrReject(callback)) return@launchInScope
 
         val options = Arguments.toBundle(data)
 
@@ -264,14 +264,14 @@ class MusicModule(reactContext: ReactApplicationContext) : ReactContextBaseJavaM
     }
 
     @ReactMethod
-    fun add(data: ReadableArray?, insertBeforeIndex: Int, callback: Promise) = scope.launch {
-        if (verifyServiceBoundOrReject(callback)) return@launch
+    fun add(data: ReadableArray?, insertBeforeIndex: Int, callback: Promise) = launchInScope {
+        if (verifyServiceBoundOrReject(callback)) return@launchInScope
 
         try {
             val tracks = readableArrayToTrackList(data);
             if (insertBeforeIndex < -1 || insertBeforeIndex > musicService.tracks.size) {
                 callback.reject("index_out_of_bounds", "The track index is out of bounds")
-                return@launch
+                return@launchInScope
             }
             val index = if (insertBeforeIndex == -1) musicService.tracks.size else insertBeforeIndex
             musicService.add(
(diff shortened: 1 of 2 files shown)


behenate:
@dcvz It would be good to have bridgeless working before `0.74` is released. Do you guys have plans for fixing the Android issues I mentioned?

brentvatne:
@dcvz - any chance you could review this? react-native-track-player came up as one of the top 400 most popular packages in the react-native and we're hoping to get as many popular libraries as possible compatible leading up to the launch of react-native@0.74 - https://github.com/reactwg/react-native-new-architecture/discussions/167

lovegaoshi:
thank you for the work @behenate I put together a bare minimum example app 
https://github.com/lovegaoshi/RNTPExampleNewArch

for the event issue, it's actually none of the hooks/useEvents work anymore; the addEventListener inside doesnt do anything I believe. however TP.addEventListener as in SetupService works. tbh i'm completely lost with the new arch stuff, if anyone have any insight, thank you very much in advance
hooks are set up as https://github.com/doublesymmetry/react-native-track-player/blob/main/src/hooks/useTrackPlayerEvents.ts

notification is a bigger issue imo bc all of that should be native.

Medkhat:
Hi there! @lovegaoshi @behenate 

RN version 0.75.1 has already been released. Does anyone have plans to improve this case, or have you found any solution for it?

dcvz:
There’s some stuff in the works, but still needs more time to be refined.

Medkhat:
> There’s some stuff in the works, but still needs more time to be refined.

Thanks a lot, there are many guys here who are working on their startups and are using your invaluable package. I think this needs to be resolved as quickly as possible. If there are any needs from my side, I'm always happy to help in any way I can. :)

migueldaipre:
Hey guys, any news ? I would like to help if needed

lovegaoshi:
sure if u can help on https://github.com/facebook/react-native/issues/46050
tia

Medkhat:
@lovegaoshi Hi! This issue isn't limited to Android, it also breaks the connection between the Metro bundler and the native side on iOS.

Medkhat:
@lovegaoshi Hi! This issue is not only Android related, it also breaks the connection between the Metro bundler and the native side on iOS)

lovegaoshi:
uh sure, u r welcome to report to fb too. im not an ios person.

gh-pro:
Hi, i found new bug, usePlaybackState(), useActiveTrack return not updated state.

afktechnologies:
any updates regarding this issue?

mihaibulic2:
This library might be abandoned (no updates in a while, emailed the company
twice about the future of the lib w/ no answer)

I'd suggest either forking it for yourself or finding another library from
a more reliable source (looks like doublesymmetry is a small dev shop and
this is their only popular OS library, so def a little risky to depend on
that).

Personally, I'm sticking with this library for now and will likely patch it
for my own needs as necessary until I hit a big snag, then I'll switch away.

jspizziri:
Closing this in favor of https://github.com/doublesymmetry/react-native-track-player/pull/2478 which incorporates it's work.

PR #2291, opened 2024-04-05
fix(web): playlist queue logic

Bram-dc:
There were at least 2 bugs that occured in my project using react-native-web:

1. Adding new songs to the end would instead insert them at index 1. This is because in an earlier function you would set an undefined parameter to -1, after evaluating if this value is set you forgot to also check for -1

2. The next song would not play if the repeat mode is set to Queue because it would never get to the logic if it should play the next song.

This PR should fix the issues. :)

### DIFF
--- a/web/TrackPlayer/PlaylistPlayer.ts
+++ b/web/TrackPlayer/PlaylistPlayer.ts
@@ -29,6 +29,8 @@ export class PlaylistPlayer extends Player {
       case RepeatMode.Playlist:
         if (this.currentIndex === this.playlist.length - 1) {
           await this.goToIndex(0);
+        } else {
+          await this.skipToNext();
         }
         break;
       default:
@@ -38,7 +40,6 @@ export class PlaylistPlayer extends Player {
           if ((err as Error).message !== 'playlist_exhausted') {
             throw err;
           }
-
           this.onPlaylistEnded();
         }
         break;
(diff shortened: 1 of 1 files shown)


web/TrackPlayer/PlaylistPlayer.ts
```diff
@@ -20,28 +20,30 @@ export class PlaylistPlayer extends Player {
...
-          await this.goToIndex(0);
-        }
-        break;
-      default:
-        try {
-          await this.skipToNext();
-        } catch (err) {
-          if ((err as Error).message !== 'playlist_exhausted') {
-            throw err;
+    try {
```
jspizziri, line 23: I see that this try/catch is to catch the `playlist_exhaused` error since we're calling another `skipToNext`. However, I'm not convinced that, in `RepeatMode.Playlist`, we should be catching that error. If the whole playlist should be repeated, I can't envision a scenario where `skipToNext` should properly throw that error, and if it does, it seems something strange must've happened (unless I'm missing a use-case as I'm thinking through it).

Is there something I'm not seeing there?

Bram-dc: Ow yeah, I think you are right.

Bram-dc: Beside that however, if the playlist repeats, should this.onPlaylistEnded(); be called?

Bram-dc: And what happends if the playlist is emptied?

jspizziri: Can you look at how Swift and KotlinAudio handle that? I don't think `onPlaylistEnded` should be called, but we'll want feature parity with those.

Bram-dc: I don't think it does. I just committed the smaller try-catch!

jspizziri (changes requested):
@Bram-dc , looks good! Can you please squash?

Bram-dc:
Done!

jspizziri (approved):

PR #2292, opened 2024-04-06
fix(web): removed dynamic shaka-player import

Bram-dc:
Let bundlers decide how to import the package instead of forcing a dynamic import.

### DIFF
--- a/web/TrackPlayer/Player.ts
+++ b/web/TrackPlayer/Player.ts
@@ -1,7 +1,6 @@
 import { State } from '../../src/constants/State';
 import type { Track, Progress, PlaybackState } from '../../src/interfaces';
 import { SetupNotCalledError } from './SetupNotCalledError';
-// @ts-ignore
 import shaka from 'shaka-player/dist/shaka-player.ui';
 
 export class Player {
--- a/web/TrackPlayer/Player.ts
+++ b/web/TrackPlayer/Player.ts
@@ -1,6 +1,8 @@
 import { State } from '../../src/constants/State';
 import type { Track, Progress, PlaybackState } from '../../src/interfaces';
 import { SetupNotCalledError } from './SetupNotCalledError';
+// @ts-ignore
+import shaka from 'shaka-player/dist/shaka-player.ui';
 
 export class Player {
   protected hasInitialized: boolean = false;
@@ -41,9 +43,6 @@ export class Player {
       // TODO: double check the structure of this error message
       throw { code: 'player_already_initialized', message: 'The player is not initialized. Call setupPlayer first.' };
     }
-
-    // @ts-ignore
-    const shaka = await import('shaka-player/dist/shaka-player.ui');
     // Install built-in polyfills to patch browser incompatibilities.
     shaka.polyfill.installAll();
     // Check to see if the browser supports the basic APIs Shaka needs.
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -14,6 +14,6 @@
     "module": "esnext",
     "skipLibCheck": true
   },
-  "include": ["src"],
+  "include": ["src", "web"],
   "exclude": ["node_modules", "**/__tests__/*"]
 }
--- a/web/shaka-player.d.ts
+++ b/web/shaka-player.d.ts
@@ -1,0 +1,3 @@
+declare module 'shaka-player/dist/shaka-player.ui' {
+  export = shaka;
+}


web/TrackPlayer/PlaylistPlayer.ts
```diff
@@ -20,28 +20,30 @@ export class PlaylistPlayer extends Player {
...
-          await this.goToIndex(0);
-        }
-        break;
-      default:
-        try {
-          await this.skipToNext();
-        } catch (err) {
-          if ((err as Error).message !== 'playlist_exhausted') {
-            throw err;
+    try {
```
jspizziri, line 23: Can you please walk me through the motivation of this change?

web/TrackPlayer/Player.ts
```diff
@@ -1,6 +1,8 @@
 import { State } from '../../src/constants/State';
 import type { Track, Progress, PlaybackState } from '../../src/interfaces';
 import { SetupNotCalledError } from './SetupNotCalledError';
+// @ts-ignore
+import shaka from 'shaka-player/dist/shaka-player.ui';
```
jspizziri, line 5: @Bram-dc can you please explain how this is intended to work? Specifically, how does it work in SSR (e.g. Next.js). I'd also like to understand the motivation behind the change.

Thanks!

web/TrackPlayer/PlaylistPlayer.ts
Bram-dc: This got mixed up with my other pull request, It's a bug fix for web.

jspizziri: @Bram-dc cool, can you force push this PR and remove that commit?

web/TrackPlayer/Player.ts
Bram-dc: I have not used Next.js with react native. Is that possible? React-native-track-player has to be on the client right? What has SSR to do with it?

jspizziri: @Bram-dc yes, SSR for web is definitely possible, but in SSR contexts you can't import or execute JS packages that rely on the dom or other things dependent upon the browser. In order to facilitate this, you dynamically import the module so that it's only imported on the clientside and not on the server.

I'm open to a better method of doing this, but I'm not open to removing SSR support. If there's a way we can achieve both (which I'm not aware of) I'd love to learn what it is.

Bram-dc: The example app is working after this change.

jspizziri: I also want to be clear, that the `example` web project is just a basic SPA react app from create-react-app. Meaning, it doesn't do SSR or contain the SSR functionality, so if you're trying to see if your change works on SSR using the example isn't going to do it for you.

web/TrackPlayer/PlaylistPlayer.ts
Bram-dc: ✅

web/TrackPlayer/Player.ts
Bram-dc: What part of this project would be running on the server, before serving to the client? Would the RNTP package not be imported as a client package entirely when using it in a SSR project?

jspizziri: @Bram-dc typically, there are a lot of UI elements that are pretty deeply tied to RNTP functionality (event hooks, helpers, etc). In order for people to adopt web into their projects, they'd likely need to invest a lot of effort in refactoring their UI so that it's based on a dynamically imported version of RNTP. Additionally, that makes things bit tricky, in that, they're also writing code that supports iOS and Android and that same code is deployed across all platforms. I'm open to being convinced otherwise, but asking users to dynamically import RNTP only in web so they can use SSR seems like a challenging proposition. It might be a different story for people who are just getting started with their UI. I'm also not clear on the problems the dynamic import is causing for you, so it's hard to see the advantages or the need.

Again, this is an early feature set so I'm sure there's a lot I haven't considered in it and I'm open to changing it for the right reasons. I just need some more help understanding the challenges you're having and balancing them with the tradeoffs.

jspizziri: @Bram-dc , thoughts here?

Bram-dc:
I now finally understand the issue.

jspizziri:
@Bram-dc , can you describe the issue you're having with the dynamic import?

Bram-dc:
Yeah, I will create a sample project!

Bram-dc:
@jspizziri https://github.com/Bram-dc/rntp-web-demo

Bram-dc:
I also think this has become an issue because shaka-player is not made for modern javascript projects and not a "bug" in RNTP. By importing the file directly bundlers will almost in every case do this extra step automatically. The dynamic import will break the code in some cases. In my projects I use vite as bundler and it has an easy fix for this problem. Maybe other bundlers do not have to options. This is also a very small group of people facing this issue -> react-native-web + rntp, and not worth patching in this library. (shaka-player is not made for these environments).

jspizziri:
@Bram-dc this is interesting... it seems like Vite is doing something wrong. [Dynamic imports are part of the JS spec](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import) so I'm not sure why it's not doing it properly. Do you have any insights there? (never used vite and I'm old enough to have bundler fatigue at this point haha).

When I ran your project, I did notice that _something_ from shaka was importing (specifically `shaka.default` existed), but I'm not sure why things are breaking. I'll have to take a bit more time digging into it. But generally speaking, something doesn't smell right to me at the moment.

jspizziri:
@Bram-dc please take a look at this PR

https://github.com/doublesymmetry/react-native-track-player/pull/2299

PR #2294, opened 2024-04-08
Fix docs for queue remove

xiaoxu193:
.remove() takes number or number[], not track objects:

https://github.com/doublesymmetry/react-native-track-player/blob/db86df1a6c6a2c1e5af0da3d94b35c4c4120b162/src/trackPlayer.ts#L162


dcvz (approved):
Nice catch @xiaoxu193 !

PR #2296, opened 2024-04-15
docs(guides): add an Expo guide

dctalbot:


docs/docs/guides/with-expo.md
```diff
@@ -0,0 +1,48 @@
+---
+sidebar_position: 4
```
jspizziri, line 2: I think this position is incorrect (looks like there are several with position 4) but it should probably come after the others but before the Amazon fire support one.

docs/docs/guides/with-expo.md
```diff
@@ -0,0 +1,48 @@
...
+
+# Developing with Expo
+
+Expo is a popular development platform in the react-native ecosystem.
+
+## Custom Entrypoint
+
+By default, Expo points to an entrypoint in the `node_modules` package. The path for this is configured in the `"main"` field of `package.json`. To configure the Playback Service in this entrypoint file:
+
+1. Copy the file
```
jspizziri, line 13: I have little to no experience with Expo, but this seems a little hacky, so my questions are:

1. Is this like an official thing that Expo says to do? And if so, is there documentation that we can link to?
2. Is this something that can be added as an enhancement to the rntp lib so that Expo folks don't have to go through it?

jspizziri (changes requested):
@dctalbot , thanks for taking the time to write this up. I think this could be a great addition! Left some feedback for you.

docs/docs/guides/with-expo.md
dctalbot: 1. Haha yeah, it's legit. They have an example of doing this (but to support a different use case - monorepos) https://docs.expo.dev/guides/monorepos/#change-default-entrypoint
2. I don't think so, unless rntp is willing to support setting up the playback service at a level below the entry point file.

jspizziri: @dctalbot cool, thanks for that context. I'd love to add that hyperlink directly to that Expo doc that you have in your comment.

jspizziri:
@isilher , thanks for the great feedback, I think that's all quite helpful.

@dctalbot , would you be able to confirm item #2 and incorporate @isilher's other suggestions?

If you're not able to get to all of this LMK.

isilher:
@dctalbot you are correct! I was checking our git history for including this library and assumed the pre-existing `index.js` in our project to be conventional (at least for development builds). However, we already included that as part of our Detox setup earlier, exactly like you specified (changing the `main` entry in package.json).

docs/docs/guides/with-expo.md
jamsch: Just a note for this step: a lot of users are going to be using `expo-router` which uses a different entry point and will likely require modification of `app/_layout.js` instead.

jamsch:
Another note: on iOS, you'll need to update `app.json` / `app.config.js` to include the following for background audio playback:

```js
{
    ios: {
      infoPlist: {
        UIBackgroundModes: ["audio"],
      },
    },
}
 ```

andordavoti:
I had an issue on Expo web because shaka-player is a dynamic import. Might be worth mentioning in this guide. My current workaround is to create two files in a shaka-player folder in my project, index.ts (empty file), index.web.ts (has an import of shaka-player with this line "import 'shaka-player/dist/shaka-player.ui';"). This works as it forces the shaka-player import on web, but would be nice to see first party Expo web support.

jspizziri:
@andordavoti , definitely would love for Expo web to work. I'm curious if [this PR](https://github.com/doublesymmetry/react-native-track-player/pull/2299) addresses your issue. Would you be able to take it for a spin and let me know? Apart from that, I'd be happy to work with you separately on Expo web support. But it's definitely something I'll need assistance on due to my ignorance surrounding Expo.

Why don't we move the conversation to the linked PR for now.

jspizziri:
@dctalbot , thanks a ton for your work on this.

@isilher @jamsch thank you for reviewing and providing feedback.

It's great to see the community come together like this!

jspizziri:
@dctalbot sorry, I have it set to auto-merge, but it looks like the docs build is failing to an invalid link. Would you be able to sort that out?

9mzodiac:
@jspizziri I just moved away from Solito / NextJS to Expo router for my app today. I had gotten RNTP working perfect in Next but instantly hit the issue @andordavoti was having when I moved to Expo web. Just checked this PR and saw you talking about expo web support today. Couldn't be better timing haha, thanks @andordavoti for surfacing this and @jspizziri for keeping this project going!

jspizziri (approved):

ajarian:
@isilher thanks for sharing your example project–that was really helpful 🙏

PR #2369, opened 2024-09-10
docs: the current version of the hook usePlaybackState returns an object.

camilokawerin:
So the comparison in the code example on `usePlaybackState` lacks the property name.


jspizziri (approved):

jspizziri:
@dcvz this LGTM

PR #2385, opened 2024-10-18
Update MusicService.kt to FIX [4.0.1] Crash: com.doublesymmetry.track…

mayank-paryani:
…player.service.MusicService.startAndStopEmptyNotificationToAvoidANR

Update MusicService.kt to FIX [4.0.1] Crash: com.doublesymmetry.trackplayer.service.MusicService.startAndStopEmptyNotificationToAvoidANR 

https://github.com/doublesymmetry/react-native-track-player/issues/2244#ref-commit-5dfc0f0

### DIFF
--- a/android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
+++ b/android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
@@ -120,9 +120,17 @@ class MusicService : HeadlessJsTaskService() {
             notificationBuilder.foregroundServiceBehavior = NotificationCompat.FOREGROUND_SERVICE_IMMEDIATE
         }
         val notification = notificationBuilder.build()
-        startForeground(EMPTY_NOTIFICATION_ID, notification)
-        @Suppress("DEPRECATION")
-        stopForeground(true)
+           try {
+            // reference exception catch of startForegroundIfNecessary function
+            startForeground(EMPTY_NOTIFICATION_ID, notification)
+            @Suppress("DEPRECATION")
+            stopForeground(true)
+        } catch (error: Exception) {
+            Timber.e(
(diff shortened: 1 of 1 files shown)


lovegaoshi:
but this doesnt fix any of the underlying problem. if ur app actually crashes with this, this try catch block isnt fixing anything. so for people reporting this "fixes" their problem, this ANR/crash actually does not affect UX to begin with.

While I mostly guessed my answer in the issue thread, it is indeed correct - the reason for startAndStopEmptyNotificationToAvoidANR is to prevent the android limitation that a playback service must start playback within X seconds of it being started. and since MusicModule creates the service when setupPlayer is called, no media playback can be guaranteed and thus this function is used.

And back to the question - why ru seeing this? several possibilities:

1. your service creation is so late and somehow the activity gets pushed in the background. which is quite unlikely for most people adopted the example app's setup.
2. your service crashed and attempted to be restarted while your app is in the background state. this is very relevant especially since the 5 seconds no playback kill service option that was recently added. Although I dont think it is actually the case here, as this will severely affect UX and it doesnt appear to be so.

In addition, at least with a full media3 service migration[, MusicService appeared to be started by the system automatically as UI will bind to the declared mediaSession.](https://github.com/lovegaoshi/react-native-track-player/commit/70f61ca0e28fd285ba6b4fd1780b62587044c3da)  


TLDR: its either an app setup problem that YOU did, or you can try keeping your MusicService alive for as long as possible.

tamdvyounetgroup:
Screenshot 2025-01-21 at 17 55 34
 if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
            startForeground(EMPTY_NOTIFICATION_ID, notification, ServiceInfo.FOREGROUND_SERVICE_TYPE_MEDIA_PLAYBACK);
        } else {
            startForeground(EMPTY_NOTIFICATION_ID, notification);
        }
the solution for it.

PR #2451, opened 2025-04-13
fix(android): correct onBind method signature to match in MusicService.kt

StuCodeGreen:
### Summary

This PR fixes a Kotlin compilation error in `MusicService.kt` related to the `onBind` method signature.

### Problem

The existing code attempted to override:

```kotlin
override fun onBind(intent: Intent?): IBinder
```

However, the Android Service class defines:
```
public abstract IBinder onBind(Intent intent)
```
Due to Kotlin's strict override checks, the mismatch causes:
```
'onBind' overrides nothing.
```

Fix
Updated the signature to match the exact expected method:

```kotlin
@MainThread
override fun onBind(intent: Intent): IBinder {
    return binder
}
```

### DIFF
--- a/android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
+++ b/android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
@@ -761,7 +761,7 @@ class MusicService : HeadlessJsTaskService() {
     }
 
     @MainThread
-    override fun onBind(intent: Intent?): IBinder {
+    override fun onBind(intent: Intent): IBinder {
         return binder
     }
 


AbdullahZareen:
+1 for now i adde dthe patch-package

occultus73:
@lovegaoshi All that question mark means in Kotlin is Intent is a nullable type. It's overwriting a java super method where Intent is not nullable. The only reason why this breaks RN 0.79 (at least with expo 53) is because they are using the latest Android plugins that are now super strict about nullable type mismatches. But obviously pretending a non-nullable is nullable is completely harmless, which is why it has been permitted by Android until now.

tl;dr this will absolutely not break RN < 0.79. It's just fixing a super pedantic typing error.

jspizziri (approved):

PR #2459, opened 2025-05-04
fix(web): add support for custom headers for web

michabrem:
Ensure that custom headers from the track entity are properly applied to HTTP requests in the web implementation. This fix dynamically registers request filters in the Shaka Player's NetworkingEngine.

### DIFF
--- a/web/TrackPlayer/Player.ts
+++ b/web/TrackPlayer/Player.ts
@@ -1,5 +1,5 @@
 import { State } from '../../src/constants/State';
-import type { Track, Progress, PlaybackState } from '../../src/interfaces';
+import type { PlaybackState, Progress, Track } from '../../src/interfaces';
 import { SetupNotCalledError } from './SetupNotCalledError';
 
 export class Player {
@@ -92,7 +92,7 @@ export class Player {
    */
   protected onStateUpdate(state: Exclude<State, State.Error>) {
     this.state = { state };
- }
+  }
 
   protected onError(error: any) {
     // unload the current track to allow for clean playback on other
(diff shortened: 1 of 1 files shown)


web/TrackPlayer/Player.ts
```diff
@@ -198,4 +198,18 @@ export class Player {
     if (!this.element) throw new SetupNotCalledError();
     return this.element.buffered.end;
   }
+
+  private addHeaders(track: Track) {
+    if (track.headers) {
+      const networkingEngine = this.player?.getNetworkingEngine();
+      networkingEngine?.clearAllRequestFilters();
```
jspizziri, line 205: @michabrem , if `track 1` has headers, and `track 2` does not, wouldn't this result in the headers for `track 1` being sent on the `.load` call of `track 2`?

web/TrackPlayer/Player.ts
```diff
@@ -198,4 +198,18 @@ export class Player {
...
+      networkingEngine?.registerRequestFilter((type, request) => {
+        if (type === 0 || type === 1) {
+          Object.entries(track.headers || {}).forEach(([key, value]) => {
```
jspizziri, line 208: @michabrem it might be more concise to express this like:

```ts
request.headers = {
   ...track.headers,
    ...request.headers,
}
```

I realize this is technically an additional "copy", but it's easier to read IMHO. Thoughts?

jspizziri (commented):
@michabrem thanks for the PR! left you a couple of questions.

web/TrackPlayer/Player.ts
michabrem: True, I have adjusted that to clear the request filters regardless of header or not, that should do the trick imo.

web/TrackPlayer/Player.ts
michabrem: i like that better as well.

michabrem:
> @michabrem thanks for the PR! left you a couple of questions.

thanks, i've finally had the time to come back to this. Tested it locally with my web app and it works with the changes.

PR #2482, opened 2025-06-26
Feat/media3 newarch => Fix ios build issues

tsipho:
Problem

The iOS builds on the feat/media3-newarch branch were consistently failing. The primary error indicated an incompatibility with hermes-engine and CocoaPods, specifically:

[!] The version of CocoaPods used to generate the lockfile (1.16.2) is higher than the version of the current executable (1.15.2). Incompatibility issues may arise.

Further investigation revealed issues with Ruby gem dependencies, particularly activesupport, leading to undefined method 'untaint' errors during bundle install and pod install.

⸻

Investigation

I found that the Gemfile in the example/ directory had strict version constraints on cocoapods, activesupport, xcodeproj, and concurrent-ruby. These constraints were preventing the installation of compatible versions of these gems with the current Ruby environment, leading to the build failures.

Additionally, the presence of a vendor/bundle directory was causing Bundler to use an outdated version, even after a global Bundler update.

⸻

Solution

To resolve these issues, I took the following steps:
	1.	Removed vendor/bundle directory
This forced Bundler to use the globally installed, updated version.
	2.	Removed strict version constraints in example/Gemfile
I updated the Gemfile to allow for more flexible and compatible versions of cocoapods, activesupport, xcodeproj, and concurrent-ruby.
	3.	Deleted Gemfile.lock and Pods directory
This ensured a clean slate for dependency resolution.
	4.	Ran bundle install
This regenerated a new Gemfile.lock with compatible gem versions.
	5.	Ran pod install
This successfully installed the CocoaPods dependencies, resolving the hermes-engine and other related issues.

⸻

Verification

I have verified that both iOS and Android builds now complete successfully without errors.
	•	iOS build was confirmed by running bundle exec pod install and inspecting the output for successful installation.
	•	Android build was confirmed by running ./gradlew assembleDebug in the example/android directory.

⸻

This fix ensures that the iOS project can build successfully, allowing for continued development on the feat/media3-newarch branch.

### DIFF
--- a/example/Gemfile
+++ b/example/Gemfile
@@ -4,10 +4,10 @@ source 'https://rubygems.org'
 ruby ">= 2.6.10"
 
 # Exclude problematic versions of cocoapods and activesupport that causes build failures.
-gem 'cocoapods', '>= 1.13', '!= 1.15.0', '!= 1.15.1'
-gem 'activesupport', '>= 6.1.7.5', '!= 7.1.0'
-gem 'xcodeproj', '< 1.26.0'
-gem 'concurrent-ruby', '< 1.3.4'
+gem 'cocoapods', '>= 1.15.2'
+gem 'activesupport', '>= 7.0.0'
+gem 'xcodeproj'
+gem 'concurrent-ruby'
 
 # Ruby 3.4.0 has removed some libraries from the standard library.
 gem 'bigdecimal'
--- a/example/Gemfile.lock
+++ b/example/Gemfile.lock
@@ -5,12 +5,18 @@ GEM
       base64
       nkf
       rexml
-    activesupport (6.1.7.10)
-      concurrent-ruby (~> 1.0, >= 1.0.2)
+    activesupport (7.2.2.1)
+      base64
+      benchmark (>= 0.3)
+      bigdecimal
+      concurrent-ruby (~> 1.0, >= 1.3.1)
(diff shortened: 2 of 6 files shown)


jspizziri:
@tsipho thanks for sending this along. The strange thing here is that, the changes in your diff, are mostly already the PR (at least, the ones that I think are contributing to this error). Take this for example: https://github.com/doublesymmetry/react-native-track-player/blob/9be123a5516bdecdffc37c61a249fa2498547cd4/example/ios/Podfile.lock#L1651


It appears that the CI pipeline is using a cached version of the installed pods based on a hash of the Podfile.lock. I think that cache is bad basically, and it might be due to me force pushing or something.

Anyways, my instinct is that this issue is a caching issue and I don't actually think anything needs to be changed except the cache needs to be invalided. LMK if you disagree.

jspizziri:
Yeah, for some reason the new build is using [the same cache key](https://github.com/doublesymmetry/react-native-track-player/actions/runs/15901227809/job/44868904994#step:8:26) for the pods as the [last successful build](https://github.com/doublesymmetry/react-native-track-player/actions/runs/15886493479/job/44799649757#step:8:4) (which was on an older version of react-native).

So this definitely seems to be a problem with the cache. I'm going to close this PR, thanks for the help!

Issue #2433, opened 2025-02-16
Reopening the app when clicking a notification


George3d6:
Is there any way, at least in cases where the app is in the background or foreground (i.e. not killed) to get the notification click behavior to just => opening the app (foregrounding it) or doing nothing (if the app is foregrounded).

At present it opens a deep-link to `trackplayer://notification.click` and I can't even figure out how to configure said link.

lovegaoshi:
[RTFM](https://reactnative.dev/docs/linking#handling-deep-links) and the example app

https://github.com/doublesymmetry/react-native-track-player/blob/2765d09cb93541d89be94a4809320e4df33c0d47/example/src/App.tsx#L60

George3d6:
@lovegaoshi what I meant to say is "could I change this link to be something else"

Figuring out how to get "trackplayer://notification.click" to "do nothing except foreground app" is a bit difficult.

Though in this specific case, I did manage ot.

Supporting a deep link that's not `trackplayer://notification.click` might help. The docs point to that being the case but mention no way of configuring it, nor can I find one in the example app.

MarianLudwig:
@lovegaoshi This works fine on Android. However, the deepLinkHandler is not being called on iOS when clicking the notification. Is this a known limitation?

lovegaoshi:
so u r a user of expo-router that thinks every single deeplink is a navigation request. This has nothing to do with how RNTP handles things, rather u should post in the expo issue board instead and ask how to ignore certain deeplink URI from their router. 

the example app shows exactly not doing anything about the deeplink but logging. I do [direct to another screen](https://github.com/lovegaoshi/azusa-player-mobile/blob/bb76fd86452a7fe0e9f6a3905d670e0657adc106/src/components/playlists/View.tsx#L76), and I cannot relate since i dont use the opinionated expo-router framework.



if u google trackplayer://notification.click in this repo u'd find this is only in one .kt file, ie. its android only. the [docs](https://rntp.dev/docs/next/basics/background-mode#notification) is poorly made but since its strictly under the android section, u can assume its android exclusive.

PR #2486, opened 2025-07-06
feat(android): add configurable notification click behavior

zeevenn:
# Notification Click Behavior Feature

This feature adds configurable notification click behavior to react-native-track-player for Android.

## Problem

Previously, when users clicked on the notification from react-native-track-player, the app would always launch with a hardcoded URI (`trackplayer://notification.click`) that allows the app to identify that it was opened from a notification click. This behavior was inflexible and didn't provide options for different use cases.

> https://github.com/doublesymmetry/react-native-track-player/issues/2433

And in expo case, `expo-router` will handle the deep link and navigate to an unmatched route. User must add a `notification.click.tsx` to redirect certain page. But sometimes user just want to open app when click notification other than navigate to certain page.

## Solution

Added a new `notificationClickBehavior` configuration option under `android` options that allows developers to:

1. **Disable URI data**: Launch app without any URI information
2. **Use custom URI**: Replace the default URI with a custom one
3. **Use custom action**: Replace the default action with a custom one
4. **Complete customization**: Combine custom URI and action

### DIFF
--- a/android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
+++ b/android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
@@ -98,11 +98,19 @@ class MusicService : HeadlessJsMediaService() {
             }
         })
         fakePlayer = ExoPlayer.Builder(this).build()
+
+        val notificationClickBehavior = androidOptions?.getBundle(NOTIFICATION_CLICK_BEHAVIOR_KEY)
+        val enableUriData = notificationClickBehavior?.getBoolean(NOTIFICATION_CLICK_ENABLED_KEY, true) ?: true
+
         val openAppIntent = packageManager.getLaunchIntentForPackage(packageName)?.apply {
             flags = Intent.FLAG_ACTIVITY_CLEAR_TOP or Intent.FLAG_ACTIVITY_SINGLE_TOP
-            // Add the Uri data so apps can identify that it was a notification click
-            data = Uri.parse("trackplayer://notification.click")
-            action = Intent.ACTION_VIEW
+            if (enableUriData) {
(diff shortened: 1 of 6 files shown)


ghost (approved):

ghost:
From [this Expo page](https://docs.expo.dev/router/advanced/native-intent/#redirectsystempath), they described how to handle incoming native links.

So, I implemented the code below in `app/+native-intent.tsx` file and it worked.

```
import { appScheme } from "@/features-users/auth/utils";

/**
 * Handles deep link redirection for your app, including third-party schemes like TrackPlayer notifications.
 * 
 * @param path - The incoming URL or path to process
 * @param initial - Whether this is the initial URL (cold start) or a hot-start URL
 * @returns A string representing the path to redirect to within your app
 */
export function redirectSystemPath( { path, initial }: { path: string; initial: boolean } ): string {
  try {
    console.log( "redirectSystemPath called with:", { path, initial } );

    // Construct a URL object. The base is your app scheme in case the path is relative.
    const url = new URL( path, `${appScheme}://` );

    // Handle TrackPlayer notification click
    // trackplayer://notification.click → url.host === "notification.click"
    if ( url.host === "notification.click" ) {
      console.log( "Redirecting to /player from TrackPlayer notification" );
      return "/player";
    }

    // No special rule matched → return the original path for normal routing
    console.log( "No redirection rule matched, returning original path:", path );
    return path;
  } catch ( err ) {
    // If parsing fails, log the error and redirect safely to home
    console.warn( "Error parsing URL in redirectSystemPath:", { path, initial, error: err } );
    return "/";
  }
}
```

Perhaps, someone can add this to the examples in TrackPlayer page.

zeevenn:
It seems will redirect to a certain route. I will try it later.

zeevenn:
I tried it again, and `+native-intent.tsx` only redirects the route, but I just want to call the app. I will resolve these conflicts later so that I can merge this feature.

Issue #2495, opened 2025-08-06
[Android] Some Live streams are not working with new media3 branch


tomislav-arambasic:
**Describe the Bug**
On Android, not all live streams are working with new arch & media 3 implementation.
Same live streams **DO** run on main branch.

Trying to play them results in error:
```
{
    "message": "Source error",
    "code": "android-io-bad-http-status"
}
```

Not working:
- https://streaming.live365.com/a39417
- https://streaming.live365.com/a62429

**Steps To Reproduce**
Modify `playlist.json` in example app adding live streams & run example app.
First 2 playlist items are added, live streams that are not working.

**Code To Reproduce**
Edit playlist.json
```
[
  {
    "url": "https://streaming.live365.com/a39417",
    "title": "Radio Stream 1",
    "artist": "Not Working",
    "artwork": "https://rntp.dev/example/smooth-jazz-24-7.jpeg",
    "isLiveStream": true
  },
  {
    "url": "https://streaming.live365.com/a62429",
    "title": "Radio Stream 2",
    "artist": "Not Working",
    "artwork": "https://rntp.dev/example/smooth-jazz-24-7.jpeg",
    "isLiveStream": true
  },
  {
    "url": "https://rntp.dev/example/Longing.mp3",
    "title": "Longing",
    "artist": "David Chavez",
    "artwork": "https://rntp.dev/example/Longing.jpeg",
    "duration": 143
  },
  {
    "url": "https://rntp.dev/example/Soul%20Searching.mp3",
    "title": "Soul Searching (Demo)",
    "artist": "David Chavez",
    "artwork": "https://rntp.dev/example/Soul%20Searching.jpeg",
    "duration": 77
  },
  {
    "url": "https://rntp.dev/example/Lullaby%20(Demo).mp3",
    "title": "Lullaby (Demo)",
    "artist": "David Chavez",
    "artwork": "https://rntp.dev/example/Lullaby%20(Demo).jpeg",
    "duration": 71
  },
  {
    "url": "https://rntp.dev/example/Rhythm%20City%20(Demo).mp3",
    "title": "Rhythm City (Demo)",
    "artist": "David Chavez",
    "artwork": "https://rntp.dev/example/Rhythm%20City%20(Demo).jpeg",
    "duration": 106
  },
  {
    "url": "https://rntp.dev/example/hls/whip/playlist.m3u8",
    "title": "Whip",
    "artist": "prazkhanal",
[...]

PR #2496, opened 2025-08-11
feat(player): add default user-agent for Media3 requests

tomislav-arambasic:
Added a custom user-agent string to prevent 403 errors when playing Live365 streams using Media3.

Resolves https://github.com/doublesymmetry/react-native-track-player/issues/2495.

## Issue
Some streams require user-agent to be defined. Such as
https://streaming.live365.com/a39417 & https://streaming.live365.com/a62429.
Starting those streams with new media3 implementation resulted in playback error 
```
{
    "message": "Source error",
    "code": "android-io-bad-http-status"
}
```
Tested in example app, on various emulators, physical devices and on different Android versions & Android Studio versions... Same issue everywhere.
`main` branch did work as expected, streams played in all scenarios.

After looking into it, server was throwing 403 - Forbidden for media3 implementation.

## Fix
Added default user-agent to comply with those restrictions.

Users are still able to override user-agent, by defining it inside track object.

Verified in example app, running on Motorola g34 5G, Android 14.
Example:
```
{
    "url": "https://das-edge11-live365-dal03.cdnstream.com/a39417",
    "title": "Radio Stream 1",
    "artist": "Not Working",
    "artwork": "https://rntp.dev/example/smooth-jazz-24-7.jpeg",
    "isLiveStream": true,
    "userAgent": "TestUserAgent"
  }
```

### DIFF
--- a/android/src/main/java/com/doublesymmetry/kotlinaudio/players/components/MediaFactory.kt
+++ b/android/src/main/java/com/doublesymmetry/kotlinaudio/players/components/MediaFactory.kt
@@ -34,6 +34,10 @@ class MediaFactory (
     private val cache: SimpleCache?
 ) : MediaSource.Factory {
 
+    companion object {
+        private const val DEFAULT_USER_AGENT = "react-native-track-player/5.0.0"
+    }
+
     private val mediaFactory = DefaultMediaSourceFactory(context)
 
     override fun setDrmSessionManagerProvider(drmSessionManagerProvider: DrmSessionManagerProvider): MediaSource.Factory {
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/doublesymmetry/kotlinaudio/players/components/MediaFactory.kt
```diff
@@ -34,6 +34,10 @@ class MediaFactory (
     private val cache: SimpleCache?
 ) : MediaSource.Factory {
 
+    companion object {
+        private const val DEFAULT_USER_AGENT = "react-native-track-player/5.0.0"
```
jspizziri, line 38: I'm not sure about specifying the version here. I know this is conventional for userAgents, however needing to manually keep this up-to-date with each version release is not something I like the sound of.

Previously this was being [defaulted to `react-native-track-player`](https://github.com/doublesymmetry/KotlinAudio/blob/bf71120704bfe4be2311cf86fc1e2ee1c3c702b7/kotlin-audio/src/main/java/com/doublesymmetry/kotlinaudio/players/BaseAudioPlayer.kt#L600C52-L600C65). I would say we either stick with that or we find a way to pull the version in dynamically.

jspizziri (changes requested):
@tomislav-arambasic this is great! Thanks for running this down and doing the work to resolve it. I've left some feedback.

android/src/main/java/com/doublesymmetry/kotlinaudio/players/components/MediaFactory.kt
tomislav-arambasic: adjusted, thanks for input.

Issue #2158, opened 2023-10-10
[bug] useActiveTrack() doesnt reflect metadata changes


lovegaoshi:
**Describe the Bug**
I use useActiveTrack() to display track metadata on the player screen, but when updating metadata via updateCurrentPlayingItem or updateMetadata with an index, useActiveTrack() does not refresh.

**Steps To Reproduce**
its here in the code:
https://github.com/doublesymmetry/react-native-track-player/blob/e28536c4b3312d6ccb76c43cc7f7fdd86d1fac41/src/hooks/useActiveTrack.ts#L28
useActieTrack only listens Event.ActiveTrackChanged, not when CurrentPlaying metadata is changed i think

**Code To Reproduce**
as above.

**Replicable on Example App?**
fairly sure the update track metadata button in the action panel doesnt change anything in the p.ayer screen

**Environment Info:**
not relevant

**How I can Help**
I will make a fix

lovegaoshi:
I'd prefer a new metadata hook/event emitted, now i think about it since I also listen to Event.ActiveTrackChanged, wouldn't want that listener to trigger on just metadata changes. An event emitted for current metadata changes would work nicely. or maybe I missed this in the doc somewhere?

puckey:
How about adding a param object to specify that you also want to receive track updates: `useActiveTrack({ updates: true })`

neoassyrian:
I was about to request similar thing. I use useActiveTrack to display the track metadata on screens. This works fine with normal tracks, however if we use a live audio such as a radio stream, once we receive the new icy metadata we update the track metadata accordingly, however the details don't update on screen because the active track didn't change.

Having a separate hook to manage that, would be a great idea. We can use that to track changes of the track metadata.

tomislav-arambasic:
@dcvz Any update on this?

My use case:
I too have screens rendering metadata.
First time radio stream screen mounts, metadata UI is rendered because I receive metadata common/timed event and update my UI and notification centre accordingly.
Metadata UI updates whenever new metadata events are incoming, as long as screen is mounted.

However, if I unmount the screen and come back, I don't have the access to metadata until radio stream updates it.
As you said, updating active track metadata is not reflected in `activeTrack` getter / hook.

I think, for now, I'll be able to resolve this using redux - inside playback service, once metadata is received, I'll have to save it to redux state and load it on screen mount, as a temporary metadata, until new timed metadata is received.
But this is hacky, a temporary workaround (hopefully).

[...]

PR #2501, opened 2025-08-14
feat (ios & android): metadata update fixes and improvements

tomislav-arambasic:
## Summary
- `useActiveTrack` fix (https://github.com/doublesymmetry/react-native-track-player/issues/2158)
  - Fixes long existing bug where `useActiveTrack` doesn't update when `updateMetadataForTrack` is called
- Updated `updateMetadataForTrack`
  - Merges track's current metadata with given metadata patch object, instead of overriding it. 
  - This applies to both - stored track and, if target track is active, then changes are reflected in now playing metadata too
  - If given metadata key value is `undefined` or `null`, current metadata will keep existing respective values
  - If given metadata key value is `empty string`, those values will be set to null in metadata after update
  - Emitting `TRACK_METADATA_UPDATED` event when calling `updateMetadataForTrack`. Same event is now observed from `useActiveTrack`, so that it can update track object after metadata update
- Updated `updateNowPlayingMetadata`
  - It'll use the same merge logic when updating now playing metadata (notification Android & Notification Center iOS)
  - As before, currently active track object remains intact, only now playing info is updated, backwards compatible
  - Emitting `NOW_PLAYING_METADATA_UPDATED` event when calling `updateNowPlayingMetadata `. Same event is now observed from new `useNowPlayingMetadata` hook, which retrieves respective now playing metadata
- Updated `ExampleApp`
  - Replaced random image logic. Previously, images would fail to render sometimes, and they'd render in-app and not in Notification Center and vice-versa. Found this while testing ExampleApp... I have no explanation 🤷 Added hardcoded array of valid image urls to be used when randomising images, which will also be preloaded for faster rendering
  - Added new Action Sheet option, showcasing how to set any of metadata values to empty, after recent changes
  - Added annotations explaining these Action Sheet actions
  - Added new events to `PlaybackService`
- Updated docs

## Breaking changes
Updating metadata by `updateMetadataForTrack` and `updateNowPlayingMetadata` now requires values to be explicitly set to empty strings, in case removal of respective metadata values are desired.
Before, metadata was overriding by given patch object, now it'll be merged/patched.

## Discussion
Regarding `merging` metadata objects when updating - I know this has been a discussion point in the past. Subjectively, I believe patching metadata is the correct way to go, but please let me know reasons behind you decision, whatever it is.
I'll update PR after final decision on patching metadata, if necessary.

### DIFF
--- a/android/src/main/java/com/doublesymmetry/trackplayer/model/Track.kt
+++ b/android/src/main/java/com/doublesymmetry/trackplayer/model/Track.kt
@@ -28,7 +28,24 @@ class Track
 
     override fun setMetadata(context: Context, bundle: Bundle?, ratingType: Int) {
         super.setMetadata(context, bundle, ratingType)
-        originalItem.putAll(bundle)
+        // Only update originalItem with fields that were actually provided
+        bundle?.keySet()?.forEach { key ->
+            originalItem.putAll(Bundle().apply {
+                putValue(key, bundle.get(key))
+            })
+        }
+    }
+
+    private fun Bundle.putValue(key: String, value: Any?) {
+        when (value) {
+            null -> remove(key)
+            is String -> putString(key, value)
+            is Int -> putInt(key, value)
+            is Double -> putDouble(key, value)
+            is Boolean -> putBoolean(key, value)
+            is Bundle -> putBundle(key, value)
+            // Add other types as needed
+        }
     }
 
     fun toAudioItem(): TrackAudioItem {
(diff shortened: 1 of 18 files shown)


andy380743909:
hi bro, any progress?

tomislav-arambasic:
> hi bro, any progress?

🤷 No response from maintainers.
It's not that important I guess, just an improvement and fix for a bug that doesn't impact many apps.

jspizziri:
@tomislav-arambasic Sorry for the radio silence. This is important to us and we're great full for the work. Right now I'm locked in on Android Auto. Once that's done this is next in my list.

tomislav-arambasic:
> @tomislav-arambasic Sorry for the radio silence. This is important to us and we're great full for the work. Right now I'm locked in on Android Auto. Once that's done this is next in my list.

@jspizziri That OK! I'm actually more excited about Android Auto anyway 😀

PR #2517, opened 2025-09-16
fix(android): add next prev player commands

sohamnakhare:
**Why / What Problem Does It Solve**

Previously, the command builder didn’t support “skip to next” and “skip to previous” remote/media control commands.

This meant that remote controls (e.g. notification controls, lock screen, media buttons) couldn’t trigger next or previous track actions via the command builder.

The commit adds in those capabilities so that clients using the command builder will have those actions available.

**Implications / Effects**

After this commit, when building PlayerCommandBuilder, users can include “next” and “previous” player commands.

Remote events (or media buttons) that request skip next or skip previous can be wired up using these commands.

Improves usability: users can now navigate tracks via remote controls or UI elements that rely on “next/prev” capabilities.

### DIFF
--- a/android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
+++ b/android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
@@ -834,7 +834,7 @@ class MusicService : HeadlessJsMediaService() {
                 KeyEvent.KEYCODE_MEDIA_PLAY_PAUSE -> {
                     // in oxygenOS, the play button in the expanded view always sends KEYCODE_MEDIA_PLAY_PAUSE 
                     // instead of distinguishing between play and pause actions
-                    emit(if (!player.playWhenReady) MusicEvents.BUTTON_PLAY else MusicEvents.BUTTON_PLAY_PAUSE)
+                    emit(if (!player.playWhenReady) MusicEvents.BUTTON_PLAY else MusicEvents.BUTTON_PAUSE)
                     true
                 }
 
--- a/android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
+++ b/android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
@@ -832,7 +832,9 @@ class MusicService : HeadlessJsMediaService() {
         if (keyEvent?.action == KeyEvent.ACTION_DOWN) {
             return when (keyEvent.keyCode) {
                 KeyEvent.KEYCODE_MEDIA_PLAY_PAUSE -> {
-                    emit(MusicEvents.BUTTON_PLAY_PAUSE)
+                    // in oxygenOS, the play button in the expanded view always sends KEYCODE_MEDIA_PLAY_PAUSE 
+                    // instead of distinguishing between play and pause actions
+                    emit(if (!player.playWhenReady) MusicEvents.BUTTON_PLAY else MusicEvents.BUTTON_PLAY_PAUSE)
                     true
                 }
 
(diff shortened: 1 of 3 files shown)


android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
```diff
@@ -290,12 +290,33 @@ class MusicService : HeadlessJsMediaService() {
...
+                    playerCommandsBuilder.add(Player.COMMAND_SEEK_BACK)
+                }
+
                 else -> {}
             }
         }
+
         customLayout = CustomCommandButton.entries
             .filter { notificationCapabilities.contains(it.capability) }
+            .filter { it != CustomCommandButton.NEXT && it != CustomCommandButton.PREVIOUS }
```
sohamnakhare: Removing this as NEXT, PREVIOUS is removed from `CustomCommandButton.kt`

tomKFM:
next, previous and pause works great, but interestingly i still cannot play the music. And I also had to remove lines to make it work because it wouldnt compile

sohamnakhare:
> next, previous and pause works great, but interestingly i still cannot play the music. 

@tomKFM [Here](https://github.com/doublesymmetry/react-native-track-player/blob/2a9e43ab484a582ba24bb87f8d0b42fc99d0d2ed/android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt#L281) COMMAND_PLAY_PAUSE is added, if pause works, play would also work just that you'd have to detect the player state and call either play / pause.

sohamnakhare:
@tomKFM which lines did you have to remove ?

tomKFM:
react-native-track-player/android/src/main/java/com/doublesymmetry/kotlinaudio/service/MusicService.kt:60:41 Unresolved reference 'NEXT'.
react-native-track-player/android/src/main/java/com/doublesymmetry/kotlinaudio/service/MusicService.kt:61:41 Unresolved reference 'PREVIOUS'.

had to remove those lines 

Concerning the pause/play bug, I know it should work. But it doesnt. I can only pause, but not play
This bug only occurs in the "extended" media notification, which only exist with oxygenOS ? Not sure about that
Im investigating on my own

By the way, my device is OnePlus 13, android 15

tomKFM:
@sohamnakhare Ok, turns out that in oxygenOS, the play button in the expanded view always sends KEYCODE_MEDIA_PLAY_PAUSE instead of distinguishing between play and pause actions. And when called, weirdly it cannot play back the track

I fixed that by simply modifying 
```
emit(if (!player.playWhenReady) MusicEvents.BUTTON_PLAY else MusicEvents.BUTTON_PAUSE)
```
at line 835 at trackplayer/service/MusicService.kt, instead of `emit(MusicEvents.BUTTON_PLAY_PAUSE)`

sohamnakhare:
@tomKFM committed your suggestions.

android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
```diff
@@ -832,7 +832,9 @@ class MusicService : HeadlessJsMediaService() {
         if (keyEvent?.action == KeyEvent.ACTION_DOWN) {
             return when (keyEvent.keyCode) {
                 KeyEvent.KEYCODE_MEDIA_PLAY_PAUSE -> {
-                    emit(MusicEvents.BUTTON_PLAY_PAUSE)
+                    // in oxygenOS, the play button in the expanded view always sends KEYCODE_MEDIA_PLAY_PAUSE 
+                    // instead of distinguishing between play and pause actions
+                    emit(if (!player.playWhenReady) MusicEvents.BUTTON_PLAY else MusicEvents.BUTTON_PLAY_PAUSE)
```
tomKFM: emit(if (!player.playWhenReady) MusicEvents.BUTTON_PLAY else **MusicEvents.BUTTON_PAUSE**)

(my bad, i modifyed my comment since then, but here **BUTTON_PAUSE** instead of **BUTTON_PLAY_PAUSE** seems more natural)

tomKFM (approved):

puckey:
Before approving this, I want to make sure we can't fix things using the custom controls which was removed in this pr. The reason for the custom controls code is to allow customising of icons in older versions of Android.

PR #2519, opened 2025-09-19
WIP: Feat/android auto

jspizziri:


puckey:
When a track in an AA playlist is selected, does that also modify the RNTP queue? I.e. that it changes to reflect the active playlist?

jspizziri:
> When a track in an AA playlist is selected, does that also modify the RNTP queue? I.e. that it changes to reflect the active playlist?

@puckey , yes. The "AA" queue is whatever is in exoPlayer's playlist. So whenever a track in AA is played, it's going to show up on the "queue" screen. However, as we discussed, this playlist is technically different than the RNTP playlist, so ATM we need to carefully manage them to keep them both in sync, otherwise the AA "Queued" screen displays some seemingly confusing behaviour to the end user.

The way I have it currently implemented, when a track in AA is selected by the user, it is inserted at the "next spot in the queue" based on your current position. To me, this seems to make the most sense but I'm open to alternative ideas on how to pull it off.

neoassyrian:
I'm trying to use this branch, but this is not building. Can this be fixed, and I can try it out

PR #2522, opened 2025-09-19
fix(ts): resolve track resources in TrackPlayer.load

puckey:


dcvz (dismissed):
Looks good but TS checks seem to fail

dcvz (approved):

PR #2526, opened 2025-09-23
fix(android): update queue with current track

patlux:
## Summary

Fixes an issue on Android where getActiveTrack() returns the previous track instead of the currently
playing track when loading a new track via the load() method.

## Root Cause

In QueuedAudioPlayer.kt, the load() method updates ExoPlayer's playlist by adding and removing media
items, but fails to update the internal queue list accordingly. The currentItem property relies on this
queue, leading to a mismatch where the queue contains stale data while ExoPlayer plays the correct
track.

## Solution

Modified the load() method to directly replace the current item in both the queue and ExoPlayer's media
items, ensuring they stay synchronized. This prevents getActiveTrack() from returning outdated track
information.


puckey:
Thanks for the pr - this is indeed the fix, but can be done shorter using `replaceItem`: https://github.com/doublesymmetry/react-native-track-player/pull/2527/files

PR #2547, opened 2025-11-07
Fuse/apply initial patch

anettegj:


anettegj:
experimental custom auth scheme implementation

PR #2556, opened 2025-12-03
Fix: Prevent "Session ID must be unique" crash on Android

elieT27:
## Problem
Android MediaLibrarySession crashes with "Session ID must be unique" when MusicService.onCreate() is called multiple times (aggressive lifecycle management on Huawei/Samsung devices, or on phone reboot).

## Solution
1. Generate unique session ID using timestamp
2. Check if mediaSession is already initialized before recreating

## Changes
- Added unique session ID generation with `System.currentTimeMillis()`
- Added `.setId(sessionId)` to MediaLibrarySession.Builder
- Added initialization check with `this::mediaSession.isInitialized`

## Testing
Tested  with phone reboot - crash no longer occurs.

### DIFF
--- a/android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
+++ b/android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
@@ -97,6 +97,14 @@ class MusicService : HeadlessJsMediaService() {
                 return "RNTP-${element.className}:${element.methodName}"
             }
         })
+        
+        // Skip if mediaSession is already initialized to prevent duplicate session crash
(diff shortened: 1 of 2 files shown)


kodraftx:
Hi @dcvz,

This fix is critical for production apps. I'm experiencing the same crash on Oppo/Xiaomi devices with version 5.0.0-alpha0-nightly.

My app is live on Play Store and I cannot downgrade to 4.x due to New Architecture (RN 0.81) requirements. High crash rates may lead to Play Store policy violations.

Could you please prioritize reviewing and merging this PR?

Thank you for maintaining this essential library!

Issue #2507, opened 2025-08-23
Some functionality doesn't work on v5.0.0-alpha0


bahmanworld:
**Describe the Bug**
Some functionality doesn't work on v5.0.0-alpha0

- Remote actions doesn't work on media notification
- capabilities doesn't apply on notification

TrackPlayer`(newArch)` works fine but not as expected like v4.1.1

**Environment Info:**
Expo SDK 54@beta
TrackPlayer 5.0.0-alpha0
Device: Real Device
Platform: Android 12

Any further updates soon?

anultravioletaurora:
Confirming that I'm running into this issue on the `5.0.0` alpha build

[Related Jellify issue](https://github.com/Jellify-Music/App/issues/482)

I'll get this reproduced in the example repo 👍

jspizziri:
@bahmanworld ,

> I couldn't reproduce the example repo

Does that mean you've successfully tested the example app and that it's working there? Or does it mean you couldn't successfully run the example app.

lovegaoshi:
like i said i CANNOT REPRODUCE on S21 + android 14. you have to give more details as the device and android version.

https://github.com/user-attachments/assets/5513358c-b961-4fd3-83d6-d51a06a082b2

sohamnakhare:
@lovegaoshi Interestingly it is working for me in the emulator but not on the device.  I'm running it on One Plus Nord 2T.

lovegaoshi:
this is quickly turning into a device specific problem which nobody here
can help. since u have the buggy device pls follow what i posted above: is
onMediaButtonEvent firing and if so whats the intent.

sohamnakhare:
@lovegaoshi it is not working on many Android devices.

@jspizziri I'm able to reproduce the issue on the example in the repo. 

Attaching the google drive link of the APK -> https://drive.google.com/file/d/1x6D3FrlERXUvboo-d3HGKHauYHbF8BoT/view

I ran this on android 14 device.

lovegaoshi:
did i not post a video i CANNOT REPRODUCE on s21 + android 14? chanting the
same thing will get nowhere unless u somehow make it reproducible on my
device, or u answer my question - is onMediaButtonEvent firing and whats
the intent

jspizziri:
@sohamnakhare ,

[...]

PR #2559, opened 2025-12-09
fix(android): check added to trigger onMediaKeyEvent if OxygenOS < 15

Jojocaster:
## Summary
Fixes an issue on Oxygen OS < 15 where `onMediaButton` doesn't get called due to the OEM's implementation. As a result, the lock screen controls didn't work on Oxygen 14.

## Root Cause
This was also an issue in Android 13 and below. It looks like OnePlus forgot to patch it on their end in Oxygen 14, but managed to fix it in Oxygen 15. I couldn't test it directly on 15, but according to [this](https://github.com/doublesymmetry/react-native-track-player/issues/2507#issuecomment-3586296223) it seems to be working fine on Oxygen 15. 

## Solution
I've extended the original hack for Android 13 so it applies to Oxygen14 as well. Tested on a OnePlus9 (Oxygen 14) and a Pixel 9, to make sure it didn't break anything on non-One Plus devices. Both worked just fine.


Jojocaster:
@lovegaoshi could you please take a look when you have a minute? 😄

lovegaoshi:
i do not maintain this lib. i fork this lib and you should do that too.

Issue #2587, opened 2026-02-23
[v5.0.0-alpha0] - KotlinAudio.QueuedAudioPlayer.load() fn does not clear queue in else branch -> it is causing other funcs to be buggy (for example TrackPlayer. getActiveTrack())


czekFREE:
I fixed it this way (FYI):

    override fun load(item: AudioItem) {
        if (queue.isEmpty()) {
            add(item)
        } else {
            val mediaSource = item.toMediaItem()

            exoPlayer.setMediaItem(mediaSource)

            queue.clear()
            queue.add(mediaSource)

            exoPlayer.seekTo(currentIndex, C.TIME_UNSET)
            exoPlayer.prepare()
        }
    }

lovegaoshi:
its not supposed to clear queue. load replaces the current media item.

jibeleck:
You are probably right, suggested bugfix is biased by my own implementation (I am using TrackPlayer as one item per queue player), replaceItem would be probably better solution for general audience.

dcvz:
We've just released **react-native-track-player v5.0.0** (https://www.npmjs.com/package/@rntp/player) — a near-complete rewrite on the New Architecture — and we're closing open issues and pull requests on the V4 codebase as part of the transition.

V5's internals diverge significantly from V4, so most reports and patches don't translate directly.

Thank you for your contributions to V4.

PR #2589, opened 2026-02-26
fix(android): sync Kotlin queue in QueuedAudioPlayer.load()

bitcrumb:
## Summary

In the `else` branch of `QueuedAudioPlayer.load()`, ExoPlayer's internal playlist was updated (via `addMediaItem` + `removeMediaItem`) but the Kotlin `queue` LinkedList was never updated to match. This caused `queue[currentIndex]` to always return the first-ever loaded item after switching tracks.

The immediate symptom: `updateMetadataForTrack(index, bundle)` reads the current track from `queue[index]`, then calls `replaceItem(index, track.toAudioItem())`. Because `queue[0]` still pointed to the original track, `replaceItem` would call `exoPlayer.replaceMediaItem` with the original stream URL — reverting playback to the first-ever loaded track within ~10 seconds of switching (the metadata refresh interval).

## Root cause

`load()` in the non-empty queue case bypasses `add()` and manipulates ExoPlayer directly, but forgot to mirror the change to the Kotlin `queue`:

```kotlin
// Before — queue[currentIndex] is never updated
} else {
    exoPlayer.addMediaItem(currentIndex + 1, item.toMediaItem())
    exoPlayer.removeMediaItem(currentIndex)
    exoPlayer.seekTo(currentIndex, C.TIME_UNSET)
    exoPlayer.prepare()
}
```

All properties that read from the queue (`currentItem`, `items`, `previousItems`, `nextItems`) and any caller that uses `queue[index]` directly (e.g. `replaceItem`) receive stale data after `load()`.

## Fix

Materialise the `MediaItem` once and assign it to `queue[currentIndex]` before passing it to ExoPlayer:

```kotlin
} else {
    val mediaItem = item.toMediaItem()
    // Keep the Kotlin queue in sync with ExoPlayer's internal queue.
    // Without this, queue[currentIndex] always returns the first-ever loaded item,
    // causing updateMetadataForTrack to call replaceItem() with a stale stream URL
    // every time metadata is refreshed.
    queue[currentIndex] = mediaItem
    exoPlayer.addMediaItem(currentIndex + 1, mediaItem)
    exoPlayer.removeMediaItem(currentIndex)
    exoPlayer.seekTo(currentIndex, C.TIME_UNSET)
    exoPlayer.prepare()
}
```

### DIFF
--- a/android/src/main/java/com/doublesymmetry/kotlinaudio/players/QueuedAudioPlayer.kt
+++ b/android/src/main/java/com/doublesymmetry/kotlinaudio/players/QueuedAudioPlayer.kt
@@ -91,7 +91,13 @@ class QueuedAudioPlayer(
         if (queue.isEmpty()) {
             add(item)
         } else {
-            exoPlayer.addMediaItem(currentIndex + 1, item.toMediaItem())
+            val mediaItem = item.toMediaItem()
+            // Keep the Kotlin queue in sync with ExoPlayer's internal queue.
+            // Without this, queue[0] always returns the first-ever loaded station,
+            // causing updateMetadataForTrack to replace the current stream URL with
+            // the original station's URL every time metadata is refreshed.
+            queue[currentIndex] = mediaItem
+            exoPlayer.addMediaItem(currentIndex + 1, mediaItem)
             exoPlayer.removeMediaItem(currentIndex)
             exoPlayer.seekTo(currentIndex, C.TIME_UNSET)
             exoPlayer.prepare()


bitcrumb:
## Question: could `replaceItem` be used here instead?

The existing `replaceItem` method already handles keeping both the Kotlin queue and ExoPlayer in sync as a single atomic operation:

```kotlin
fun replaceItem(index: Int, item: AudioItem) {
    val mediaItem = item.toMediaItem()
    queue[index] = mediaItem
    exoPlayer.replaceMediaItem(index, mediaItem)
}
```

Which would simplify `load()` to:

```kotlin
override fun load(item: AudioItem) {
    if (queue.isEmpty()) {
        add(item)
    } else {
        replaceItem(currentIndex, item)
        exoPlayer.seekTo(currentIndex, C.TIME_UNSET)
        exoPlayer.prepare()
    }
}
```

The main difference is that the current implementation uses `addMediaItem` + `removeMediaItem` (two separate operations) whereas `replaceItem` uses `exoPlayer.replaceMediaItem` (a single atomic operation). Was the add/remove approach chosen intentionally — for example, to ensure a specific transition behaviour with live streams or to avoid a stream restart that `replaceMediaItem` might trigger? If not, delegating to `replaceItem` would eliminate the risk of this sync bug recurring.

bitcrumb:
Just noticed that this issue is talking about the same thing: https://github.com/doublesymmetry/react-native-track-player/issues/2587

dcvz:

PR #2600, opened 2026-03-13
Start MusicService in foreground for API 26+

haecksenwerk:
When the screen is turned off, Android kills the MusicService after ~2 minutes because it lacks an explicit "Started" state, even if audio is playing.

This PR adds an explicit call to startForegroundService (or startService for older APIs) prior to binding.

Tested successfully on a Pixel 9a with Android 16

### DIFF
--- a/android/src/main/java/com/doublesymmetry/trackplayer/module/MusicModule.kt
+++ b/android/src/main/java/com/doublesymmetry/trackplayer/module/MusicModule.kt
@@ -217,6 +217,11 @@ class MusicModule(reactContext: ReactApplicationContext) : NativeTrackPlayerSpec
         val musicModule = this
         try {
             Intent(context, MusicService::class.java).also { intent ->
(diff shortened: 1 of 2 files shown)


lovegaoshi:
not saying the fix is invalid but if i remember correctly context.startService is now completely delegated to how media3 handles the service cycle (ie no need to manually manage). the startCommand part u commented out was my observation where i saw onStartCommand triggers on every play intent (eg play btn from notifications). do u have a recording showing this issue? i do not experience this bug using a forked lib. 

also fyi a v5 is being released likely from a completely rewritten codebase.

haecksenwerk:
> the startCommand part u commented out was my observation where i saw onStartCommand triggers on every play intent (eg play btn from notifications).

I thought that the onStartCommand triggers were intentionally for Media3 intents to update the foreground notification

> also fyi a v5 is being released likely from a completely rewritten codebase.

Good news, then I can close this one because it seems too ambiguous

Issue #2605, opened 2026-03-21
Android notification remote controls (next/prev/jump) not working on some older devices in react-native-track-player@5.0.0-alpha0 (play/pause works)


solomonmulatu:
Hi! I’m seeing a consistency problem on Android where the notification/lock-screen media controls are shown, but only play/pause works reliably. Next / Previous (and jump forward/backward / seek actions when present) do not respond on older Android devices.

### Environment
react-native-track-player: 5.0.0-alpha0
React Native: 0.81.5
Expo: ~54.0.33 (Dev Client / native build)
New Architecture: enabled (reanimated requires it)
Device-Huawei Android Version-8.1.0
Device-Realme Android Version-11

### What works
Notification media button: Play
Notification media button: Pause

### What does NOT work
On the notification / lock-screen player:
Next does nothing (no track change)
Previous does nothing
Jump forward/backward (e.g. +15s/-15s) does nothing (when buttons exist)
In some cases, the UI button visually presses but playback doesn’t change

https://github.com/user-attachments/assets/90e92e9c-3253-434b-ba10-a8a7cb166b4b

vladvlasov256:
The root cause is in `MusicService.onStartCommand()`:

```kotlin
if (!commandStarted) {
    commandStarted = true
    super.onStartCommand(intent, flags, startId)
}
```

After the first call, super.onStartCommand is never called again. This means Media3 never processes `CUSTOM_NOTIFICATION_ACTION` intents for next/previous/jump, so `onCustomCommand` is never reached.

Play/pause still work because they go through `onMediaButtonEvent` and the `ForwardingPlayer` overrides, not through `onCustomCommand`.

Fix: always forward to super:
```
override fun onStartCommand(intent: Intent?, flags: Int, startId: Int): Int {
    onStartCommandIntentValid = intent != null
    if (Build.VERSION.SDK_INT < Build.VERSION_CODES.TIRAMISU) {
        onMediaKeyEvent(intent)
    }
    super.onStartCommand(intent, flags, startId)
    return START_STICKY
}
```

Tested on Vivo, Android 12.

lovegaoshi:
plz post the relevant android logcat on what intent emits in onStartCommand.

solomonmulatu:
@vladvlasov256  thanks  a lot now it is working.

brianlenz:
Thanks, @vladvlasov256, for the fix here! I have opened a PR with the fix: #2616

[...]

PR #2616, opened 2026-03-29
fix(android): ensure next/previous/jump notification actions work after first command (#2605)

brianlenz:


dcvz:

Issue #2617, opened 2026-03-29
[Android] App restarts headless after being killed when using StopPlaybackAndRemoveNotification, causing side effects


brianlenz:
**Describe the Bug**
When `AppKilledPlaybackBehavior.StopPlaybackAndRemoveNotification` is configured, killing the app from recents causes React Native to restart in headless mode unexpectedly. This produces side effects that can break the app (e.g. re-initializing state, re-running playback service logic) despite the intent being to fully stop playback and remove the notification.

The root cause is that `onStartCommand` returns `START_STICKY` regardless of the configured `AppKilledPlaybackBehavior`. When `StopPlaybackAndRemoveNotification` triggers `exitProcess(0)`, Android sees the service die and restarts it due to `START_STICKY`, which boots React Native headlessly. This has caused many issues in Android in our app that is using the RNTP 5.0 alpha. We currently have a yarn patch with a workaround in place, so I wanted to submit a PR to make it available to the community.

**Steps To Reproduce**
1. Configure `AppKilledPlaybackBehavior.StopPlaybackAndRemoveNotification` in `updateOptions`
2. Start playing audio
3. Kill the app from the recents screen
4. Observe that React Native restarts in headless mode and any side effects in your playback service fire unexpectedly. In the case of our app, there has been a lot of inconsistent state as the result of the headless behavior (Amplify sign in state issues primarily).

The issue is intermittent and has manifested itself in many ways through bug reports from our users.

**Code To Reproduce**
```
TrackPlayer.updateOptions({
  android: {
    appKilledPlaybackBehavior: AppKilledPlaybackBehavior.StopPlaybackAndRemoveNotification,
  },
  // ...
});
```

**Replicable on Example App?**
Likely yes, by adding `AppKilledPlaybackBehavior.StopPlaybackAndRemoveNotification` to the options in the example app and observing the process lifecycle after killing the app.

**Environment Info:**
*(paste `npx react-native info` output here)*
`react-native-track-player` version: `5.0.0-alpha0`

**How I can Help**
[...]

PR #2618, opened 2026-03-29
fix(android): prevent headless restart when stop-playback-and-remove-notification is set (#2617)

brianlenz:
return START_NOT_STICKY when AppKilledPlaybackBehavior is StopPlaybackAndRemoveNotification to prevent Android from restarting the service after exitProcess(0). return START_STICKY for ContinuePlayback and PausePlayback to allow background playback to persist.

### DIFF
--- a/android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
+++ b/android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
@@ -175,7 +175,11 @@ class MusicService : HeadlessJsMediaService() {
             commandStarted = true
             super.onStartCommand(intent, flags, startId)
         }
-        return START_STICKY
+        return if (appKilledPlaybackBehavior == AppKilledPlaybackBehavior.STOP_PLAYBACK_AND_REMOVE_NOTIFICATION) {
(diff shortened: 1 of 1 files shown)


brianlenz:
@dcvz I'm not 100% sure this is a correct solution. In our app, I have a custom patch that forces onStartCommand to always return `START_NOT_STICKY` since we use `STOP_PLAYBACK_AND_REMOVE_NOTIFICATION`. The challenge here is how to make the return value work generically for any consumer of the app (for all `AppKilledPlaybackBehavior` options).

I also question whether the defaults are established correctly. `MusicService` defaults to:
```
private var appKilledPlaybackBehavior =
        AppKilledPlaybackBehavior.STOP_PLAYBACK_AND_REMOVE_NOTIFICATION
```
But if/when you call `updateOptions()`, the default _changes_ to `CONTINUE_PLAYBACK`. Shouldn't those be the same?

Finally, I'm wondering if `onStartCommand` needs to be triggered whenever `updateOptions` is called. This would be the only way to get the return value updated to "match" the new `appKilledPlaybackBehavior`.

So, please consider this a work-in-progress 🙏 I'd love feedback here!

dcvz:

Issue #2619, opened 2026-03-29
[Android] Crash on new architecture when ReactHost is null or React context unavailable during headless task startup or event emission


brianlenz:
**Describe the Bug**
On new architecture (bridgeless mode), `HeadlessJsMediaService.getReactContext()` and `createReactContextAndScheduleTask()` call `getReactHost()` without guarding against a `null` return value. If `ReactHost` is null (e.g. during app startup or teardown), this causes an NPE that crashes the app silently in the background.

Additionally, calls to `emit()` and `emitList()` in `MusicService` are unguarded — if the React context is unavailable at the time of emission (common during backgrounding on new architecture), an uncaught exception crashes the service.

**Steps To Reproduce**
1. Enable new architecture (bridgeless mode)
2. Start playback and background the app
3. Trigger a playback event (track change, state change, etc.) while backgrounded
4. Observe a crash or silent service termination in logcat

**Code To Reproduce**
No special configuration needed beyond enabling new architecture in your app.

**Replicable on Example App?**
Likely yes with new architecture enabled.

**Environment Info:**
```
System:
  OS: macOS 26.4
  CPU: (10) arm64 Apple M1 Max
  Memory: 337.66 MB / 64.00 GB
  Shell:
    version: 5.3.9
    path: /opt/homebrew/bin/bash
Binaries:
  Node:
    version: 25.6.1
    path: /opt/homebrew/bin/node
  Yarn:
    version: 4.13.0
    path: /opt/homebrew/bin/yarn
  npm:
    version: 11.9.0
    path: /opt/homebrew/bin/npm
  Watchman: Not Found
Managers:
  CocoaPods:
    version: 1.15.2
    path: /Users/brian/.rbenv/shims/pod
SDKs:
  iOS SDK:
    Platforms:
      - DriverKit 25.4
      - iOS 26.4
      - macOS 26.4
      - tvOS 26.4
      - visionOS 26.4
      - watchOS 26.4
  Android SDK:
    API Levels:
      - "29"
      - "30"
      - "31"
      - "33"
      - "34"
      - "35"
      - "36"
    Build Tools:
      - 26.0.2
      - 26.0.3
      - 28.0.3
      - 29.0.2
      - 30.0.2
      - 30.0.3
      - 31.0.0
      - 31.0.0
      - 31.0.0
      - 33.0.0
      - 33.0.1
      - 34.0.0
      - 35.0.0
      - 36.0.0
    System Images:
[...]

PR #2620, opened 2026-03-29
fix(android): guard against null ReactHost and emit crashes on new architecture (#2619)

brianlenz:
add null checks for ReactHost in HeadlessJsMediaService.getReactContext() and createReactContextAndScheduleTask() to prevent NPEs on new architecture (bridgeless mode). wrap emit() and emitList() in MusicService with try/catch to prevent uncaught exceptions from crashing the service when the React context is transiently unavailable while backgrounded.


dcvz:

Issue #2621, opened 2026-03-29
[Android] ExoPlayer (fakePlayer) and MediaSession leaked when service is destroyed before player initialization, causing OOM crashes


brianlenz:
**Describe the Bug**
`MusicService.onCreate()` always creates a `fakePlayer` (a full ExoPlayer instance) and a `MediaLibrarySession`. Once `setupPlayer()` is called, `fakePlayer` is released and the session is handed off to the real player. However, if the service is destroyed before `setupPlayer()` is ever called — which can happen on service restart, OS-initiated teardown under memory pressure, or abnormal initialization — the original `onDestroy()` only entered the `if (::player.isInitialized)` branch, leaving `fakePlayer` and `mediaSession` unreleased.

ExoPlayer holds substantial native resources (audio decoders, codec handles, buffers). With `START_STICKY`, Android restarts the service after termination, creating a new `fakePlayer` on each `onCreate()`. Without releasing the previous instance, these accumulate and cause OutOfMemory crashes — particularly noticeable when the app is backgrounded and the service cycles through stop/start sequences.

**Steps To Reproduce**
1. Configure `AppKilledPlaybackBehavior.ContinuePlayback` or `PausePlayback` (so `START_STICKY` is in effect)
2. Kill and restart the app repeatedly, or allow the OS to cycle the service under memory pressure before `setupPlayer()` completes
3. Observe OOM crashes or growing native memory usage in logcat/profiler

**Code To Reproduce**
No special configuration needed beyond standard playback setup with `ContinuePlayback` or `PausePlayback` behavior.

**Replicable on Example App?**
Likely yes under memory pressure or repeated service restart cycles.

**Environment Info:**
```
System:
  OS: macOS 26.4
  CPU: (10) arm64 Apple M1 Max
  Memory: 337.66 MB / 64.00 GB
  Shell:
    version: 5.3.9
    path: /opt/homebrew/bin/bash
Binaries:
  Node:
    version: 25.6.1
    path: /opt/homebrew/bin/node
  Yarn:
    version: 4.13.0
    path: /opt/homebrew/bin/yarn
  npm:
    version: 11.9.0
    path: /opt/homebrew/bin/npm
  Watchman: Not Found
Managers:
  CocoaPods:
    version: 1.15.2
    path: /Users/brian/.rbenv/shims/pod
SDKs:
  iOS SDK:
    Platforms:
      - DriverKit 25.4
      - iOS 26.4
      - macOS 26.4
[...]

PR #2622, opened 2026-03-29
fix(android): release fakePlayer and mediaSession in onDestroy when player is uninitialized (#2621)

brianlenz:
fakePlayer (ExoPlayer) and mediaSession are always created in onCreate but were only released in onDestroy when player was initialized. if the service is destroyed before setupPlayer() is called, these native resources were silently leaked. with START_STICKY causing service restarts, each cycle creates a new fakePlayer without releasing the previous one, leading to accumulated native resource consumption and OOM crashes. add an else branch to onDestroy to release both resources when player has not been initialized.


dcvz:

Issue #2564, opened 2025-12-29
Notification completely broken on MIUI 13 / Android 12 - buttons unresponsive and in wrong order


u2gilles:
**Description**
On MIUI 13 (Android 12), the notification media controls are completely non-functional:

Buttons do not respond to taps - pressing Play, Pause, or any button has no effect
Buttons are in wrong order - not matching the configured notificationCapabilities
"Skip to Next" button is missing entirely
The same app on Android 15 (Xiaomi HyperOS) shows the correct button layout and all controls work perfectly.

**Environment**
RNTP Version: react-native-track-player@^5.0.0-alpha0-nightly-359af5a12d712d3b685530aed9b9625865a25d74
Broken Device: MIUI Global 13.0.16, Android 12 (security patch 2022-01)
Working Device: Xiaomi, Android 15 (HyperOS)

**Configuration**
```
await TrackPlayer.updateOptions({
  notificationCapabilities: [
    Capability.Play,
    Capability.Pause,
    Capability.JumpBackward,
    Capability.JumpForward,
    Capability.SkipToNext,
    Capability.SkipToPrevious,
  ],
  forwardJumpInterval: 30,
  backwardJumpInterval: 30,
});
```

![Image](https://github.com/user-attachments/assets/48c32176-4d49-4222-9f1c-f0b803e4fc56)

LetItRock:
Bump, I can confirm I’m seeing the same behavior on my end as described in this issue. 
Is there a workaround or patch that works? @jspizziri

niklavsBariss:
Same issue on Google Pixel 3 XL with android 12. Pause/play button doesn't work

Thijmen:
Is this project dead?

dcvz:
We've just released **react-native-track-player v5.0.0** (https://www.npmjs.com/package/@rntp/player) — a near-complete rewrite on the New Architecture — and we're closing open issues and pull requests on the V4 codebase as part of the transition.

V5's internals diverge significantly from V4, so most reports and patches don't translate directly.

Thank you for your contributions to V4.

Issue #2254, opened 2024-02-19
React native track player doesn't show "skip to next" button in background menu for some Android phones


nicksoft2020:
React native track player works on Samsung phone correctly, but it does not show "skip to next" button in background menu for other Android phones like Redmi 9. I use 4.0.1 version of react native track player.

![redmi 9 (screen 2)](https://github.com/doublesymmetry/react-native-track-player/assets/50168853/059d94ad-b35d-45ef-b483-6ab224f07983)
![redmi 9 (screen 3)](https://github.com/doublesymmetry/react-native-track-player/assets/50168853/80d9e110-09eb-410d-a8b8-737c958dcadd)
![redmi 9 (screen 4)](https://github.com/doublesymmetry/react-native-track-player/assets/50168853/8141da45-1c19-4e11-8068-86fd4870d880)
![redmi 9 (screen 1)](https://github.com/doublesymmetry/react-native-track-player/assets/50168853/46c473ea-a679-45cd-8f86-b695fbc600dd)

I tried to use all available properties for Capabilities such as notificationCapabilities, compactCapabilities and capabilities, but it did not help me to fix an issue for Redmi 9 phone. There is setup track player function code below:

`try {
    await TrackPlayer.getCurrentTrack();

    isSetup = true;
}
catch {
    await TrackPlayer.setupPlayer();
    await TrackPlayer.updateOptions({
        android: {
            appKilledPlaybackBehavior:
            AppKilledPlaybackBehavior.StopPlaybackAndRemoveNotification,
        },
        capabilities: [
            Capability.Play,
            Capability.Pause,
            Capability.SkipToNext,
            Capability.SkipToPrevious,
            Capability.SeekTo
        ],
        compactCapabilities: [
            Capability.Play,
            Capability.Pause,
            Capability.SkipToNext,
            Capability.SkipToPrevious,
        ],
        progressUpdateEventInterval: 2,
    });

    await TrackPlayer.setRepeatMode(RepeatMode.Off);

    isSetup = true;
}
finally {
    return isSetup;
}`

See my package.json

[...]

Issue #2100, opened 2023-08-17
Lock screen button Capabilities (Next-previous , JUMP_FORWARD-JUMP_BACKWARD)


MaxJadav:
**Describe the Bug**
I have to display 4 buttons on player lock screen, But it showing only 2 of them

**Code To Reproduce**
 capabilities: [
          TrackPlayer.CAPABILITY_PLAY,
          TrackPlayer.CAPABILITY_PAUSE,
          TrackPlayer.CAPABILITY_STOP,
          TrackPlayer.CAPABILITY_SEEK_TO,
          TrackPlayer.CAPABILITY_JUMP_FORWARD,
          TrackPlayer.CAPABILITY_JUMP_BACKWARD, 	  
          TrackPlayer.CAPABILITY_SKIP_TO_NEXT,
          TrackPlayer.CAPABILITY_SKIP_TO_PREVIOUS,
        ],

<img width="246" alt="1" src="https://github.com/doublesymmetry/react-native-track-player/assets/31030098/ea2c1523-4f54-4cee-b9e3-cf0f11efca0d">

It should display this 4 button 
TrackPlayer.CAPABILITY_JUMP_FORWARD,
TrackPlayer.CAPABILITY_JUMP_BACKWARD, 	  
TrackPlayer.CAPABILITY_SKIP_TO_NEXT,
TrackPlayer.CAPABILITY_SKIP_TO_PREVIOUS

Is it possible? Native iOS Support it?

Issue #592, opened 2019-05-15
NotificationCapabilities next and previous action button is not working in background mode.


RohitNNaik:
# Configuration

Run `react-native info` in your project and share the content.
What `react-native-track-player` version are you using?

# Issue

# Code

Show us the code you are using

RohitNNaik:
react-native-track-player: 1.1.4,
react-native: 0.59.8

brenwell:
More info perhaps?

RohitNNaik:
Player.js

import React from 'react'
import {View, Text, TouchableOpacity, Image,ActivityIndicator} from 'react-native'
import TrackPlayer, {ProgressComponent} from 'react-native-track-player';
import Tracker from '../Tracker';

export default class Player extends React.Component{

	constructor()
	{
        super()
        this.state = {
            AudioStatus: true,
            isLoadingPlay:false,
            isLoadingPrev:false,
            isLoadingNext:false,
        };
    }
    
    componentDidMount() {
        TrackPlayer.setupPlayer();
        TrackPlayer.updateOptions({
            stopWithApp: true,
            compactCapabilities: [
              TrackPlayer.CAPABILITY_PLAY,
              TrackPlayer.CAPABILITY_PAUSE,
              //TrackPlayer.CAPABILITY_STOP,
              TrackPlayer.CAPABILITY_SKIP_TO_NEXT,
              TrackPlayer.CAPABILITY_SKIP_TO_PREVIOUS
            ],
            capabilities: [
                TrackPlayer.CAPABILITY_PLAY,
                TrackPlayer.CAPABILITY_PAUSE,
                TrackPlayer.CAPABILITY_SKIP_TO_NEXT,
                TrackPlayer.CAPABILITY_SKIP_TO_PREVIOUS,
            ],
            notificationCapabilities:[
              TrackPlayer.CAPABILITY_PLAY,
              TrackPlayer.CAPABILITY_PAUSE,
              //TrackPlayer.CAPABILITY_STOP,
              TrackPlayer.CAPABILITY_SKIP_TO_NEXT,
              TrackPlayer.CAPABILITY_SKIP_TO_PREVIOUS
            ]
            
        });
    }
    togglePlayback = async () => {
        const currentTrack = await TrackPlayer.getCurrentTrack();
        
        if (currentTrack == null) {
          console.log("track is null")
          this.setState({
            AudioStatus: false,
            isLoadingPlay:true
        });
          TrackPlayer.reset();
[...]

PR #2632, opened 2026-04-29
fix(android): add SKIP_TO_NEXT/PREVIOUS to available player commands

d-mato:
## Problem

When `Capability.SkipToNext` / `Capability.SkipToPrevious` are passed to `updateOptions`, the corresponding Media3 `Player.Commands` (`COMMAND_SEEK_TO_NEXT(_MEDIA_ITEM)` / `COMMAND_SEEK_TO_PREVIOUS(_MEDIA_ITEM)`) are never added to the player's `availableCommands`. As a result, the next/previous buttons on the lock screen, notification, and Android Auto remain disabled even when the app declares those capabilities.

Other capabilities (`PLAY/PAUSE`, `STOP`, `SEEK_TO`) are already wired up in the same `when` branch, so this looks like an oversight.

## Fix

Add two cases next to the existing `Capability.SEEK_TO` branch in `MusicService.updateOptions` — minimal change, same pattern:

```kotlin
Capability.SKIP_TO_NEXT -> {
    playerCommandsBuilder.add(Player.COMMAND_SEEK_TO_NEXT)
    playerCommandsBuilder.add(Player.COMMAND_SEEK_TO_NEXT_MEDIA_ITEM)
}
Capability.SKIP_TO_PREVIOUS -> {
    playerCommandsBuilder.add(Player.COMMAND_SEEK_TO_PREVIOUS)
    playerCommandsBuilder.add(Player.COMMAND_SEEK_TO_PREVIOUS_MEDIA_ITEM)
}
```

## Why this is safe

`BaseAudioPlayer` already intercepts `seekToNext()` / `seekToNextMediaItem()` (and the previous variants) via a `ForwardingPlayer` override (`BaseAudioPlayer.kt:485-499`). These overrides do not advance the native queue — they only emit `MediaSessionCallback.NEXT` / `PREVIOUS`, which `MusicService` translates into the `BUTTON_SKIP_NEXT` / `BUTTON_SKIP_PREVIOUS` events (`remote-next` / `remote-previous`) consumed by JS.

In other words, the JS-side `Event.RemoteNext` / `Event.RemotePrevious` pipeline is already wired up; only the `availableCommands` registration needed to make the lock screen / notification buttons actually pressable was missing. This PR adds that, without changing native queue-advance behavior or the JS contract.

## Relation to #2517

A previous PR (#2517) addressed the same root cause but bundled additional changes (removal of `CustomCommandButton`, modifications to `KEYCODE_MEDIA_PLAY_PAUSE` handling) and was closed as stale without merge. This PR intentionally limits the scope to only the `availableCommands` registration, leaving `CustomCommandButton`, `onMediaKeyEvent`, and `JUMP_FORWARD/BACKWARD` untouched.

## Repro

1. Configure TrackPlayer with `capabilities: [Capability.Play, Capability.Pause, Capability.SkipToNext, Capability.SkipToPrevious]` and a queue of 2+ tracks.
2. Start playback, lock the device (or pull down the notification shade).
3. Tap the next/previous buttons on the lock screen / notification.

**Before**: buttons are visible but disabled / no-op.
**After**: buttons advance/rewind the queue as expected.

## Platform impact

- Android only. iOS already handles `SkipToNext` / `SkipToPrevious` correctly via `MPRemoteCommandCenter` and is not affected.

## Related issues

Related to #2564, #2100, #2254, #592.

### DIFF
--- a/android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
+++ b/android/src/main/java/com/doublesymmetry/trackplayer/service/MusicService.kt
@@ -290,6 +290,16 @@ class MusicService : HeadlessJsMediaService() {
                     playerCommandsBuilder.add(Player.COMMAND_SEEK_IN_CURRENT_MEDIA_ITEM)
                 }
 
+                Capability.SKIP_TO_NEXT -> {
+                    playerCommandsBuilder.add(Player.COMMAND_SEEK_TO_NEXT)
+                    playerCommandsBuilder.add(Player.COMMAND_SEEK_TO_NEXT_MEDIA_ITEM)
+                }
+
+                Capability.SKIP_TO_PREVIOUS -> {
+                    playerCommandsBuilder.add(Player.COMMAND_SEEK_TO_PREVIOUS)
+                    playerCommandsBuilder.add(Player.COMMAND_SEEK_TO_PREVIOUS_MEDIA_ITEM)
+                }
+
                 else -> {}
             }
         }


dcvz:

